PR #10, opened 2026-07-26
Task Manager: list instances, not tasks; bill callbacks per instance

jggonz:
## Why Mines and Hello never showed up

The Task Manager's list was a **scheduler task-table** view — `MAX_TASKS` rows indexed by task slot. Loaded packages are task-less by design: `loader.inc` sets `I_TASK = 0xFF` because SPEC §20.2 forbids packages from spawning. They run entirely inside `W_PAINT` / `W_ONKEY` / `W_ONCLICK` callbacks on the UI task, so a task-indexed list could never see them — and the same was true of every task-less built-in (About, Note Pad, Disk). Their RAM was folded into the aggregate `RAM uuuK/tttK` line but never attributed.

## Per-instance CPU accounting (SPEC §8.1)

- **`I_CYC`** — dword in the instance record at offset 28 (the old reserved bytes), zeroed by `inst_alloc`.
- **`task_cycles` / `task_debit`** (sched.inc, placed outside the `sch_isr → sch_switch → sch_resume` fall-through run). Both readings come from one `sch_cycles` slot, which only advances while that task is on the CPU, so the delta is real CPU time rather than wall-clock across a pre-emption. `task_debit` **moves** the cycles off the task instead of copying them, so instance rows and the System row still partition one total, and nesting (an onclick that repaints every window) needs no special case.
- **`inst_win_owner` / `inst_charge`** — called from the three callback dispatch sites (`wm.inc` W_PAINT, `ui.inc` W_ONKEY / W_ONCLICK). The stamp rides on the stack, never a static, so nested paints bill correctly. The callbacks' own register contracts are unchanged.

## The list is now an instance list (SPEC §28)

`INST_MAX + 1` rows: **System** (task 0's remainder — the kernel's own work, once every callback has been debited away) plus one row per instance slot, a stable slot↔row mapping like the dock. Columns `NAME / ST / CPU / MEM` in exactly 20 chars. State gains `evt` (task-less: runs only in window callbacks) and `die` (closing, task not yet reaped). MEM is the instance's own region in KB, or `-` for a built-in kind, which owns none; System shows the kernel footprint. Window grew 250 → 264px for the extra rows, and the RAM sum now comes off the same snapshot instead of a second `cli` window.

## Bug found while verifying

`sch_account`'s pending-IRQ disambiguation can return a slightly **backwards** timestamp, so an end-minus-start delta goes negative. Unclamped, the borrow wrapped MINES's 32-bit tally by −39594 cycles (42399 → 2805) — visible only by reading `I_CYC` over QMP, not on screen. `task_debit` now forces a negative delta to 0 and subtracts nothing; after the fix the counter is monotone across bursts.

## Verification (QEMU)

| Check | Result |
|---|---|
| MINES / HELLO listed | `evt`, 2K / 1K; RAM 38K → 40K on load, back on close |
| MINES CPU under load | **17%** during repeated full repaints |
| Close a package | row frees, region reclaimed, HELLO keeps its slot |
| `W_ONKEY` path | Note Pad accrued 5,739 cycles typing "hello jop" |
| Soak | 7 instances (2 Clocks, 2 Bounces, 2 packages, TaskMgr) stable |

Both 1.44MB and 360KB images build clean with `-w+error`. Kernel 13,384 bytes; image + bss well under the 0xA000 guard.

Note: task-less apps read 0% at idle, which is arithmetically correct — a Mines click is ~0.3 ms against a 0.5 s sample. The number only becomes visible when they do real work.

SPEC.md §8.1, §11, §13, §28, §29.1/29.4 updated ahead of the code, plus CLAUDE.md.


jggonz:
Also checked in four 86Box screenshots (`vm/xt/screenshots/`), which upgrade the verification story: the change is no longer QEMU-only.

`Monitor_1_20260726-122422-376.png` shows the new instance list running on the emulated **IBM PC/XT** — 256K conventional RAM, 360KB floppies, the `make xt` path that has historically gone unverified:

- `MINES  evt  0%  2K` and `HELLO  evt  0%  1K` — the two packages, listed with their region sizes
- `Disk   evt  0%  -` and `Note Pa evt 0% -` — task-less built-ins, also previously invisible
- `Bounce slp  7%` / `TaskMgr run 49%` / `System rdy 43% 38K` — task-owned rows and the kernel remainder still adding up
- `RAM 40K/256K` — the 360KB/XT memory picture rather than QEMU's 639K

The 2026-07-25 pair is the same machine on the previous build (`#  TASK  ST  CPU`, no packages listed), so the two read as a before/after of the list format.

PR #43, opened 2026-08-03
A package can ask the kernel for memory, and Paint arrives (SPEC.md 2.6/42)

jggonz:
Paint is a bitmap editor **contributed as a fork of os8088 by [Elendilon](https://github.com/Elendilon)** — eight tools, a 4bpp offscreen canvas, one-level undo that doubles as redo, an internal clipboard, and BMP *and* GIF load/save through the Standard File dialog. It was written against the v2 ABI, and porting it is what this PR is really about.

Under v2 a package shared the kernel's segment and took the memory it needed by **picking an address**. Paint claimed a quarter megabyte above `BB_SEG`, mirrored the kernel's own `DB_MIN_KB` to guess whether the back buffer would ever want that block, and wrote `W_W`/`W_H` into the window record by hand. Under v3 neither works: a package's DS cannot reach the record, and the address it used to pick is inside the arena every package segment now comes from. `docs/PAINT-NOTES.md` had already written down what the kernel would have to provide instead — so that is what it provides.

## The arena's second claim map (`kernel/cmem.inc`, SPEC.md §2.6)

The same memory, asked for rather than taken:

```
cm_alloc(paragraphs) -> segment          OSAPI_MEM_ALLOC  0x01B8
cm_free(segment)                         OSAPI_MEM_FREE   0x01C0
cm_caps -> largest free, total free      OSAPI_MEM_AVAIL  0x01C8
```

**One walk, two maps.** `ld_alloc` is now a jump into `cm_fit`, which rejects a candidate overlapping any package region *or* any grant, enumerated together by `cm_claim`. Two allocators over one arena that searched separately would eventually hand out the same paragraph, so there are not two allocators.

Grants are stamped with the calling instance (`snd_req_inst`, the §34.3 rule) and force-freed by `cm_release_rec` at all three §29.4 teardowns — which is what makes the API safe for a **task-less** package, since nothing ever tells one that its window is closing.

### The bug that fell out of it

`ld_alloc` reserves a region at §21 step 5, but the instance record is not published until step 10, *after* the entry proc returns — so for the whole of that call the loading package's own region reads as free to a walk whose only evidence is `I_STATE`. Paint was the first package to call `OSAPI_MEM_ALLOC` from its entry and was handed `ARENA_SEG`: **the segment it was executing in**. It filled its canvas with white and the machine wedged mid-repaint on the first `0xFF` opcode, gfx lock held. `cm_hold`/`cm_unhold` is the reservation for that window, MOV-only so the entry proc's CF still reaches the loader's `jc`.

## `wm_resize` (slot 0x01D0, SPEC.md §11.1)

Retires the other liberty. `ui_grow` and `wm_fullscreen` were the only things that could resize a window and both are kernel-internal, so an app whose content has a size of its own — a picture just opened, a canvas whose shrink must be **refused** because it would crop artwork — wrote the record behind the kernel's back. The slot clamps to `WMIN_W`/`WMIN_H` and the screen, re-clamps the position through `wm_fit`, and **draws nothing**: the caller is inside its own callback and knows what it is about to repaint.

## The Task Manager bills grants

`cm_inst_mem` feeds the RAM line, the ARENA caption, the per-row SIZE column and the memory map's bands. A grant is routinely an order of magnitude larger than the region beside it; omitting it drew a 1px sliver where a quarter of the machine had gone.

## Paint itself

The v3 port is the `retf` rule on all five kernel-called procs, opaque window handles through `OSAPI_WM_GEOM`, `ES:DI` for the file-dialog completion, and one assembly at org 0. Its canvas, undo image, clipboard and scratch area are **one grant**, sized from `OSAPI_MEM_AVAIL`'s largest free run — so the v2 claim record and `pt_dupchk` are gone outright: two Paints get two canvases. It deliberately leaves 64KB behind when doing so still funds the top tier, because a canvas plus an equal undo image otherwise eats a 233KB arena whole and nothing else can load while Paint is open.

## About Paint

The app's **name in the menu bar is now a menu**. A set whose `AM_NAME` is the empty string and whose first menu is titled "Paint" puts a pull-down exactly where the label would have been — which is where a Macintosh keeps About. No kernel change: the dock tile, the Task Manager row and the loader all read the package header's name. The window is a second, instance-less window of the file dialog's species, and its command is dispatched **ahead of the mode test**, so a machine too small to fund a canvas can still be told what the program is and who wrote it.

## Verification

Verified on QEMU: launch, draw, resize through the new slot, Save As to a 320×280 4bpp BMP that a host reader accepts and `os88disk --verify` passes, close (RAM 385K → 152K, the grant released), two instances with the second tiered down to 448×146, Minesweeper loading beside a full-size Paint, and the whole app on the CGA path at 448×110. Fractal's worker task still runs, so the loader and allocator changes are clean.

**Unexercised:** 86Box (`make xt` / `xt-640` / the AT-class targets) and the 256KB floor, where the arena is empty and Paint should put up its notice rather than a canvas.

Guard 2 goes from 20,756 bytes of slack to 20,074.


jggonz:
## Follow-up: "Disk error" on opening a saved BMP — fixed in 2b89eff

Reported after the PR went up: saving works, opening the same file answers **Disk error**. The file was fine — byte-exact on disk, accepted by a host reader and by `os88disk --verify`.

**It is a kernel bug, and an old one.** `dsk_xfer` transfers one sector per int 13h call, and SPEC.md §18.1 concluded from that "no multi-sector calls, so track boundaries and DMA alignment never matter". The first half is true; the second is not. The floppy controller goes through the 8237, whose 16-bit address register is paired with a **page register that does not increment** — so a transfer of *any* length may not cross a 64KB physical boundary, and a BIOS asked to anyway answers AH=09h. `dskw_ioerr` maps that to `FERR_IO`.

Paint stages a load in its undo image, `[pt_smaxp]` paragraphs above its arena grant. On a 640KB machine that lands at linear `0x79500`, so the 54th sector of the read covered `0x7FF00..0x800FF` and straddled the page at `0x80000`.

**Why it hid until now.** Every disk buffer in the tree before this one was 512-aligned *by accident* — package regions are 32-paragraph aligned **and** 32-paragraph sized, so a package's bss always began on a 512 boundary, and the kernel's own buffers live in `LOW_SEG`. The first unaligned buffer was the first one to come from an arena grant (§2.6), where a base can be any paragraph. The matching *save* wrote from `pt_base`, which happened to be aligned — which is why writing worked and reading did not.

**The fix belongs in the kernel.** `ES:BX` is a package's memory and nothing in that contract mentions alignment, nor should it — an application cannot be asked to know about the 8237. `dsk_xfer` now tests whether a sector would straddle (low 16 bits of the linear address exceeding `0x10000-512`, computed as `(ES << 4) + BX` in 16-bit arithmetic, where the wrap *is* the mod-64K the test wants) and stages such a sector through `dsk_dmabuf`, its own 512-byte buffer in `.lowbss`. At most one extra 512-byte copy per 64KB of transfer. The retry loop moved into `dsk_try` so the direct and staged paths cannot drift.

One trap worth recording: `mov cl, 4` + `shl ax, cl` is the natural way to multiply a segment by 16 on an 8086 — and **CL is the sector number** at that point, six instructions after the LBA→CHS conversion. Writing it that way made every mount fail. Four `shl ax, 1` is the version that works.

**Verified:** the round trip that failed now answers "Opened" with the picture intact, and the `filetest` gate passes 21/21 with free space restored on all three volumes — plain FAT12, `--scramble` fragmented, and FAT16 — each followed by a clean host `os88disk --verify`.

PR #46, opened 2026-08-04
Tracker: an FT2-style MOD player as the tenth package — worker-fed ring streams, readbig, wide rates, scroll blits (SPEC.md 45/34.5/20.3/18.4/32/5.5)

jggonz:
## What this is

An homage to FastTracker II as a loadable package: **TRACKER.O88**, a fullscreen 4-channel ProTracker MOD player — the first shipped client of `wm_fullscreen` (§11.2). Splash window → any key → the FT2 layout: nameplate, position editor, BPM/Spd/Ptn readouts, 31-instrument list, per-channel volume-bar scopes with mutes, and the pattern editor with the centered current-row band, hex row numbers both sides, and the classic pale-blue cell text. Loads modules through the Standard File dialog (which works on top of the fullscreen surface), plays the full PT effect set, and degrades honestly (CGA/Hercules compact layouts; viewer-only without an SB; "Out of memory" below the arena it needs). Ships **BEVERLY.MOD** (116KB) in APPS next to it, plus `tools/mkmod.py`'s deterministic 5.6KB test module.

## Kernel amendments (SPEC.md first, per §1)

1. **Worker-fed streams + ring mode (§34.3/§34.5/§20.3/§20.6).** Stream verbs 1/3/5/6/7 become any-task (bounded `cli` windows; caller segment in a register instead of the clobberable `osapi_dseg`; the `[snd_inst]` dispatch stamp task-qualified — fixing a latent cross-task misattribution affecting TONE/FM/PLAY). Verb 0 gains **ring mode**: free-running counters over a power-of-two ring, whole-half fills, never padded — endless playback from a 16KB grant. The refill task's fill path gains the act+generation guard its siblings had (`sbl_fill_guard`), closing a close+reopen race.
2. **`dskw_readbig`, slot 0x01E8 (§18.4).** Segment-advancing whole-file read (ES base, DX:CX 32-bit capacity → DX:AX bytes read) — files ≥ 64KB into an arena grant.
3. **SB IRQ discovery rebuilt (§34.5).** The 2xEh bit-7 "confirm" was a misreading — that bit means "response byte waiting", F2h queues none outside QEMU (whose never-consumed 0xAA masked the bug in every headless test), and the same gate in `sbl_isr` would have wedged streaming on real hardware. Now: DSP ≥ 4.00 reads the configured IRQ from **mixer register 80h**; older DSPs keep the F2h probe accept-on-fire with the spurious-IRQ-7 in-service split; the ISR keeps 2xEh as the ACK only.
4. **The wide-rate regime (§34.5/§2.2).** Output 22,223–44,100 Hz on DSP ≥ 4.00: rate via 41h, start via C6h, and the DMA halves widen to 2×4KB (a 2KB half is less than one scheduler tick at 44.1kHz) by annexing the record ring's first 4KB — provably idle under the router's half-duplex rule. Every half-sized quantity became runtime state (`[sbl_half]`).
5. **`osapi_gfx_dbuf`, slot 0x01F0 (§32).** A package-facing `bb_set` returning the previous state, gated on `bb_avail` in *both* directions so a package can never strip a mono adapter's software renderer.
6. **`gfx_scroll`, slot 0x01F8 (§5.5).** Vertical scroll blit — signed dy, byte-column granularity, whole-shape clipping (a blit cannot be cut per pixel, so a clip that cuts the rect is CF=1 and the caller repaints). VGA uses GC write mode 1 (one `movsb` = 8 pixels across four planes through the latches); with the back buffer armed it flushes first, then scrolls buffer and VRAM in step marking nothing dirty — coherent by construction; mono copies per row through `gfx_rowbase`.

## The player's modes (§45.9–§45.12)

- **XT mode** — playback sized for a 4.77MHz 8088: 5,500 Hz, volume tables pre-scaled so four channels sum as byte adds straight into the output buffer, bounds checks hoisted out of the inner loop via conservative run-cutting (~95 cycles/sample vs ~160+), pattern view repainting per *position*. Worst case ~2.1M of the XT's 4.77M cycles/s. **Auto-enabled on tier-0 machines** via `osapi_cpu_info`; toggled by X / File ▸ "XT Mode".
- **Rate menu** — 11/22/44 kHz (R cycles in fullscreen), radio-marked with the `MENU_DIS` twin idiom; 44 kHz refuses honestly on a pre-4.x DSP.
- **Smooth** — fullscreen borrows the §32 back buffer (one lock hold = one flush), restoring the user's mode on exit. Default Off; S / View ▸ "Smooth".
- **Scroll path + delta VU** — a one-row move costs two `gfx_scroll` calls plus three strips (band, the strip that left it, the row entering at the far edge) instead of 35 strips, in both directions, with a clean fallback to the full repaint when a clip cuts the area. VU bars paint only their growth or shrinkage.

## Tooling & machines

`tools/mkmod.py` (deterministic test MOD), `tools/os88disk.py` data-file mode (§24), Makefile wiring, and three 86Box sound machines: **`make xt-sound`** (640KB XT + SB v2.0, 8-bit ISA), **`make 286-sound`** / **`make 386-sound`** (SB16) — configs validated via the launch/SIGTERM/read-back technique.

## Verification

- Full QEMU flows (`make test-snd SB16=1 TESTAPPS=build/tracker-test.img`): launch → fullscreen → dialog → play; **sndcheck dominant 330.3 Hz vs expected 327 ±8%** across every mode — XT on/off, 11/22/44 kHz (the 44k leg exercises 41h/C6h/4KB halves live), Smooth on/off, scroll path both directions, restart-while-playing.
- BEVERLY.MOD (116,085 B) loads via readbig from the shipped apps.img and plays; channel mutes verified; CGA boot-tested with playback (also the mono bank-copy path, settled state byte-identical across consecutive dumps).
- Smoothing verified by six mid-scroll screendumps all catching complete frames.
- Adversarial review workflow: 27 findings → 25 confirmed → 13 fixed, 6 skipped with recorded reasons.
- Kernel: all five §15.1 guards pass; `os88disk.py --verify` passes on all data images.
- **Review note**: `build/kernel.bin` in commit `99b45db` briefly carried a `VIDEO=cga`-forced kernel from a verification run; `e5db04a` rebuilds it clean (`vid_force = 0`). Worth a glance at the final artifact bytes before merge.
- Still owed to real hardware by design: the XT-mode wall-clock claim (`make xt-sound`), the single-cycle DSP < 2.00 branch (`sndcard = sb1.5`), and ears on the 86Box SB16 machines post-discovery-fix.


jggonz:
Third commit: **XT mode** (SPEC §45.9) — playback sized for a 4.77MHz 8088, off by default, toggled by the X key or File ▸ "XT Mode: Off/On". Rate drops to 5,500 Hz, the volume tables pre-scale so four channels sum as byte adds straight into the output buffer (the 16-bit accumulator + zero + conversion passes disappear), the bounds check leaves the inner loop via conservative run-cutting, and the pattern view repaints per position with a clip-gated band relight per row. Worst-case mixing lands ~2.1M of the XT's 4.77M cycles/s. QEMU-verified in both modes at identical pitch; the wall-clock claim itself is `make xt-sound` territory — that's the machine to hear it on.

jggonz:
Fourth commit: **Rate menu** (SPEC §45.10) — 11/22/44 kHz picks (R key cycles them in fullscreen), radio-marked via the MENU_DIS twin idiom. 22 kHz fits the existing TC contract; 44 kHz adds the kernel's **wide-rate regime** (§34.5): DSP ≥ 4.00 only, rate via 41h, start via C6h, and the DMA halves widen to 2×4KB by annexing the record ring's first 4KB (idle by half-duplex) — a 2KB half is less than one scheduler tick at 44.1kHz and would stutter by architecture. QEMU-verified live at all three rates in one session; on a pre-4.x card the 44 kHz pick refuses honestly with "44 kHz needs a DSP 4.x card".

jggonz:
Fifth commit: **Smooth** (SPEC §45.11) — the fullscreen flicker was the erase-then-text row strips hitting VRAM sequentially; the fix is the §32 back buffer, reached through a new package-facing slot **0x01F0 `osapi_gfx_dbuf`** (returns the previous state; both directions gated on `bb_avail` so a package can never strip a mono adapter's software renderer). The tracker arms it entering fullscreen and hands back the user's mode on exit; View ▸ "Smooth: On/Off" (key S, default On) exists so a slow-bus VGA machine can decline the flush cost. Six mid-scroll screendumps now all catch complete frames.

jggonz:
Sixth commit: **`gfx_scroll`** (SPEC §5.5, slot 0x01F8) — the pattern scroll was re-rendering ~1,600 glyphs per row change to reproduce content the screen already held eight rows away. Now it moves those pixels: VGA uses GC write mode 1 so one `movsb` shifts 8 pixels across all four planes through the latches; with the back buffer armed it flushes first, then scrolls buffer and VRAM in step marking nothing dirty (coherent by construction, so scrolling stays cheap under Smooth); mono copies per row through `gfx_rowbase`. The tracker's one-row moves cost **two scrolls + three strips** instead of 35, in both directions, with a clean fallback to the full repaint when a clip region cuts the area (whole-shape rule — the file-dialog case). VU bars are delta-drawn too: only the growth or shrinkage is painted. Verified VGA (Smooth on/off, both scroll directions) and CGA (mono bank copy, settled state byte-identical across dumps).

PR #47, opened 2026-08-04
Experimental -> Main : loadable drivers, a 1KB granularity claim heap, and much less redrawing

Elendilon:
`experimental` merged `main` at `b401eda` after a capability-by-capability audit, so **everything `main` has is here** — including the Tracker, Arkanoid, Solitaire, the Sound Blaster tier, `xmem` and the foldered apps disk. What follows is what this branch adds on top.

111 commits, 66 files. `BRANCH-DIFFERENCES.md` is the full catalogue; `docs/PORTING.md` covers bringing forward work written against `b401eda`.

## Features

- **Loadable drivers** (`.DRV`, header v4) — OPL2 and Sound Blaster moved out of the kernel into `SOUND.DRV`, so a machine with no card never carries their thousands of lines.
- **FAT12 system disk** — drive A: mounts, browses and writes. `SYSTEM.CFG` persists the whole Control Panel across reboots.
- **A four-rung RTC ladder** — MC146818, RP5C01/TC8521, MM58167, then the BIOS. An XT with a SixPakPlus keeps time; `int 1Ah` alone could not see it.
- **Note Pad has a real caret** — arrow keys, Home/End, Delete and click-to-position. It was append-only before.
- **Paint** — resizable canvas with a negotiated resize (`W_ONSIZE`), and GIF alongside BMP.
- **Folders everywhere** — create in File->Save, delete (recursive), and a current-directory the file API resolves against for Save.
- **New API**: `MEM_CLAIM`/`FREE`/`AVAIL`/`REGROW`/`CLAIM_DMA`, `WM_ONSIZE`, `WM_TITLE`, `FONT_GLYPHS`, `FILE_HERE`/`GOTO`, `GFX_SCROLL`, `GFX_DBUF`, `DRV_TASK`.
- **Task Manager** — memory-map view, per-instance CPU and RAM billing, CPU tier and XMS.

## Optimizations

- **Boots and runs usefully on 128KB.** Measured floor: 80KB boots, browses both floppies and loads a package; 176KB runs every shipped app at full function. `heap = int 12h − 73KB`.
- **Memory is claimed in 1KB increments, on demand** — not reserved at boot. The 60KB package pool and the four pinned segments (`SND_SEG`, `SAVE_SEG`, `VIEW_SEG`, `BB_SEG`) are gone; a package's region is an ordinary claim, freed with the instance.
- **Damage-rectangle repaint** — closing, hiding or dragging a window repaints the rectangle it vacated, not the screen.
- **Raising a window costs one window**, not a full repaint. Retitling costs one caption strip.
- **Clip regions** — a partly covered window draws its visible region instead of skipping the frame.
- **Incremental chrome** — the menu bar and dock redraw per changed tile, so a focus change costs two tiles and a quiet desktop costs none.
- **Redraw gating in the apps** — Note Pad hashes rows and redraws only what moved; the file dialog repaints the name field alone while you type; the Task Manager keys every element on its own hash.
- **The kernel is one contiguous span** at `0x0060`, every rung derived from the measured size of the one below, held by a build guard.

## Bug fixes

- **Sound grants were misattributed across tasks** — a worker pre-empting a foreign callback inherited that callback's instance, so its tone was billed to, and released at the teardown of, the wrong app.
- **Arkanoid** — a caught powerup was left on screen; the worker's relative sleep produced a frame-rate cliff (fixed by pacing against an absolute deadline).
- **The Tracker over-claimed 128KB per module** regardless of size, and now hands back the difference: 128KB → 6KB on a 5.6KB module.
- **Disk writes could straddle a 64KB DMA boundary** — every `int 13h` destination is 512-byte aligned by construction now, asserted at build time.

## Testing

`make && make check-images && python3 tools/checkdocs.py` are clean. Booted and driven over QMP on VGA and CGA; the scroll blit was verified byte-exact against unscrolled references on both the banked mono path and the buffered VGA path. Timing claims are cycle counts, not measurements — QEMU does not model 8086 timing.


jggonz:
Really great work! I went ahead and pulled this into a new branch, resolved a couple of things that seemed reasonable according to Fable 5 and merged it into the main branch. I made sure that you received credit on the PR.

jggonz:
This is my first open-source project, so I still have to figure out how to accept pull requests across repos correctly.

PR #61, opened 2026-08-07
Elendilon: hard disks, fullscreen exclusive, file associations, 30% mono rendering performance

Elendilon:
Hard disks are a driver, and a volume is an index (SPEC.md §52, drivers/hdd/). [disk_drive] becomes a volume index, never an int 13h drive number, so 'A'+drive, FS_DRV, the desktop zones and the file API keep working untouched. The kernel's half is five API slots, a six-row volume table and one branch in dsk_xfer; the driver owns the probe, its own Control Panel page, and one window with one button that partitions and formats. Mount survives a reboot inside SYSTEM.CFG rather than a file of its own. The FAT becomes a window (§18.8), per-volume and heap-claimed (§18.8.1), because a 32MB FAT16 FAT is 254 sectors. Two transports: the controller ROM, and the IDE task file gated on CPU_286 — an 8088's in ax, dx is two 8-bit cycles at one port and loses the drive's high byte.

Fullscreen exclusive (SPEC.md §53, kernel/fsx.inc). A bracket, not a latch: the app far-calls back through its own dispatcher and the kernel never runs while the video mode is foreign. Nine modes behind an adapter caps mask, and on a 286+/VGA machine with XMS the desktop's four planes are stashed and written back for an instant restore. Three consumers, each proving a different half: Missile Command (Mode X), Tracker (FSXM_TEXT80 on an XT — which brings the full scrolling pattern grid back at zero glyph cells), and Paint (same mode, no caps bit, runs on all three adapters — what it borrows the machine for is the lock).

File type associations (SPEC.md §54). Document icons composed at mount, double-clicking a document opens it in its program, packages declare their own extensions, icons in the Standard File dialog, and ASSOC.DAT as the volume's icon cache so the composition is paid once.

The Task Manager leaves the kernel (SPEC.md §28). kernel/taskmgr.inc → apps/taskmgr/taskmgr.asm: ~6KB of kernel that every machine carried forever for a window most sessions never open, off both memory guards. It ships in the root of both floppies because the chip menu loads it by name off whatever is in A:.

The system clipboard (SPEC.md §55). One text buffer for the machine, three slots, sized to what is in it — copy in Note Pad, close Note Pad, paste elsewhere.

Note Pad gains a scrolling view with bounded layout walks, selection with cut/copy/paste, undo grouped by typing bursts, and regex find/replace; its document is a heap claim that grows rather than bss.

The package ABI grows by 25 slots, and renumbers none of the 71 that exist (SPEC.md §20.3). A package can now take the machine (OSAPI_FSX_RUN, 0x02C8), share text with an app that is not running (OSAPI_CLIP_*, 0x0320..0x0330), receive the document it was launched with and claim an extension of its own (OSAPI_ARG_FILE 0x02E8 / OSAPI_ASSOC_SET 0x02F0), draw an arbitrary-angle line and resume the walk later (OSAPI_GFX_LINE 0x02E0, plus LINIT/LSTEP/LSTEPV), set the pen and the disabled flag in one call (OSAPI_GFX_PEN 0x0310 — §47 rule 1, because a bare grey rounds to black on 1bpp and every package here shipped a greyed control that was pixel-identical to a live one), ask whether it is the active application rather than merely frontmost (OSAPI_MENU_OWNER 0x02B8 — the question a paused game needs and wm_top cannot answer, since clicking the bare desktop moves nothing in the z-order), and read the kernel's own tables a snapshot at a time (§20.9, 0x0298/0x02A0 — which is what finally let the Task Manager leave).

The mouse is on COM1 or COM2 (SPEC.md §9.5). Both ports are probed and listened to; the first to deliver a run of clean packets wins. The hard part is that the other port usually has a modem on it and a Hayes OK is a well-formed mouse packet — so each port decodes independently, and a device that cannot win must not be able to starve the one that should.

The cursor and the gfx lock (SPEC.md §7.1). A field log put the lock/unlock pair at 21.8% of a Missile Command session with no pixel of the game in it. The arrow is 8x12 and the tables were 16x16; on 1bpp the three passes fuse into one read-bank-paint-write; on VGA the planar draw is one store; a move writes every byte exactly once (this was the Hercules wash-out); and the hide is deferred, with wm_clip_set spending it. 17.82 → 5.41 counts on mono, 1.54x on VGA, framebuffer byte-identical.

Drawing and repaint economy. gfx_line as a primitive (§5.6), its three-column dilated walk (§5.6.6, 1.91x), a resumable walk so an erase can replay a draw (§5.6.7), and a vector of steps (§5.6.8). Above that: coming to the front costs one window and going away costs a rectangle (§11.90/§11.91), retitling costs a strip (§11.92), a write marks the folder and the focus spends it (§22.8 — the "Bad package" bug a 5150 found), a paste repaints the Disk windows rather than the screen (§22.3), and mounting costs the zones that moved (§26.3).

The measurement apparatus, which is most of why the rest is shaped this way. PERFORMANCE.md (new) carries fourteen sets measured on a real IBM PC 5150; docs/FIELD-MACHINES.md records who owns the iron and what a run costs them; docs/FIELD-NOTES.md records what hardware found and the harness could not. tests/ gains gfxbench, sysbench, benchlib.inc, linetest and stackprobe, and make field builds the two bootable field disks. The headline correction: ~756us of fixed cost per drawing call and ~1ms per glyph cell — a redraw is priced by how many primitive calls it makes, not by how many pixels it covers, which is the opposite of what every estimate in this tree used to assume.

The file path gets faster where the time actually goes. dskw_rdata — the body behind OSAPI_FILE_READ — claimed in its own header comment to mirror the write side's run coalescing and had quietly stopped doing it, so the largest file operation in the system issued one int 13h per sector: a 116KB Tracker module was 295 sectors in 244 calls, and is 295 in 34 (§18.4.2 — the sector count staying identical is what says the splitter is not dropping work). Alongside it, a volume switch stops being a full mount (§18.9), which took the reference copy from 348 icon-harvest reads to 8, and a copy now costs two volume switches per file where it cost five (§22.5).

Missile Command, now playable on mono (§48.8–§48.18). On the two 1bpp adapters the game was unplayable — a busy frame was 190ms against a 55ms tick — and it now holds its 18fps on the machine it was written for. The useful part is the record of what did not work, kept in the spec rather than quietly reverted: §48.10 is an earlier optimisation of ours trading 39 one-row fills for 6 nested rects, which overlap, so covering a 39-row disc wrote 184 scan lines; §48.13 is this branch discovering that "fullscreen" had been an ordinary window all along, still in the z-order and still taking the gfx lock once a frame, at 21.8% of a session; and §48.18.1 is a vector gfx_fill costed against the measured floor and not built, because a batch removes ~170 of ~4,381 clocks — a walk step's fixed cost is per call, a fill's is per rect.

Also: 720KB images throughout, drivers may own a Control Panel page (§31.9), per-class driver publication (§51.2.1 — a sound card and a hard disk on one machine used to disconnect each other), and no driver row is wanted by default, which took a fresh image's boot from 24 transfers / 40 sectors to 5 / 13.

Verification

Testing per docs/TESTING.md — QEMU on all three adapters (VIDEO=cga, tools/hercshot.py for Hercules), 86Box for the detection probe and period timing, and the field disks on the 5150 for the three defects neither emulator can show.


jggonz:
## Pre-merge safety review — 15 issues found, all fixed in a776fed

I reviewed this branch specifically hunting for things that could **destroy someone's data, corrupt memory, hang the machine, or create a legal problem** — not style. The review ran as ten parallel passes over different failure domains, and every finding was then handed to a second reviewer whose job was to *disprove* it. Six candidate findings died at that stage. Fifteen survived, and I've fixed all fifteen and pushed them to this branch.

I've rebased onto your latest (`3c1b2e5`), rebuilt, and re-tested. **No code that already worked has changed behaviour** — every fix either adds a check that refuses something dangerous, or corrects arithmetic that was silently wrong.

Below, each item is written as: *what could have happened* → *why* → *what I changed*.

---

# The nine serious ones

### 1. Saving files to a hard disk would have quietly corrupted them — and the FAT, and kernel memory

**What could have happened.** You format a 32MB hard disk partition, use it happily for a while, and then somewhere past about 4MB of accumulated files things start going wrong: files that overwrite each other, a disk a real PC's `CHKDSK` reports as cross-linked, and — worse — random kernel data being overwritten while it happens.

**Why.** This branch made a genuinely good change: the FAT (the disk's table of contents) used to be loaded into memory whole, but a 32MB partition's FAT is 254 sectors, which is too big. So it became a **sliding window** — only nine sectors of the FAT are held in memory at a time, and it slides as needed.

The problem is that only the **reading** half got converted. The **writing** half — `dskw_setfat`, which records "this cluster belongs to that file" — kept calculating positions as if the *whole* FAT were still in memory. Think of it as a book: the code was told "you've got pages 90–98 open in front of you," but it kept counting from page 1. Ask it for page 300 and it writes 200 pages past the end of what it's holding.

Two things went wrong at once, and this is what makes it the worst item on the list:

- The entry never reached the real FAT, so the allocator **handed the same disk cluster out twice** — two files claiming the same space.
- The write landed *outside* the buffer, into the memory region immediately after it — which holds **task stacks and the directory listing**. Then the flush wrote those foreign bytes to **both copies** of the FAT, destroying the redundancy the system relies on to recover from a bad read.

The code contradicted its own spec, which describes the correct behaviour word for word. And it passes every test on a floppy, because a floppy's entire FAT fits in the nine-sector window — below cluster 2,048 the wrong arithmetic and the right arithmetic give the same answer. It only bites on the hard disks this branch just added.

**What I changed.** Both the write and the flush now go through `dsk_fat_ofs` — the exact same routine the *reading* side uses, which is exercised constantly and known good. I also made the flush **refuse** a write whose range has drifted outside the window rather than write bytes it can't vouch for, and made a failed FAT write propagate as a failure instead of being silently dropped.

---

### 2. The hard disk driver could write to the wrong physical drive — at boot, with nobody watching

**What could have happened.** You have two hard disks. The machine boots, and before the desktop even appears, os8088 writes to the *second* one when it meant the first.

**Why.** IDE drives are addressed with a byte that packs several things together: which drive on the cable (bit 4), and which head (the bottom 4 bits). The code built that byte by OR-ing the head number in **with no mask**. Its own comment said "CH = head (0..15)" — but nothing enforced it. A head number of 16 sets bit 4, which is *the drive-select bit*. Head 64 sets a different bit that switches the drive into an entirely different addressing mode.

The dangerous part is how a bad head count gets in. Geometry is restored from `SYSTEM.CFG` on every boot, and the automount runs during driver startup — **before the first screen is drawn, with no user present**. The sibling routine right next to this one *does* mask correctly, so the two disagreed.

**What I changed.** A geometry an IDE drive cannot express is now **refused** rather than silently truncated — checked when the geometry is validated, when you type it into the Control Panel (the field simply stops at 16), and again at transfer time. The mask at the port is still there, but as a last line of defence rather than the only one.

---

### 3. Changing a drive's geometry made every subsequent read and write land on the wrong sector

**What could have happened.** You correct a drive's geometry in the Control Panel because the machine's BIOS got it wrong — exactly the scenario this feature exists for. From then on, everything the driver reads or writes goes to a different physical spot on the platter than intended. Formatting scatters the boot sector, both FATs and the root directory across the disk.

**Why.** An IDE drive has to be *told* its geometry with a command called INITIALIZE DEVICE PARAMETERS. This branch sent that command exactly once, during hardware detection. But two later paths overwrite the geometry — the Control Panel editor, and the restore from `SYSTEM.CFG` — and **neither re-sent the command**. So os8088 was converting addresses using one geometry and the drive was interpreting them with another. The spec states the rule this code breaks, in as many words.

Boot order hits it with no user involved at all: detect → load config → overwrite geometry → automount.

**What I changed.** Added `hd_geom_push`, which re-teaches the drive whenever the geometry changes, called from both paths. If the drive *refuses* the new geometry, the row is marked unusable rather than being addressed with a translation the drive rejected — better no disk than the wrong sectors on a real one.

---

### 4. A disk that was momentarily hard to read would be reported as blank — and one click would then wipe it

**This is the one I'd have least liked to explain to a user.**

**What could have happened.** You have a hard disk with three real DOS partitions on it. The very first sector is slightly marginal, or the drive hasn't finished spinning up. You open the disk tool. It shows **four empty slots — "Not Formatted"**. Reasonably, you click Format. That single click overwrites the real partition table and formats over the data behind it. No confirmation appeared at any point.

**Why.** Two separate mistakes lined up:

1. `hd_part_load` treated *"the read failed"* and *"this disk has never been partitioned"* as the **same outcome** — both produced a fabricated blank table. It did set a flag, `hd_mbrok`, to record the difference... and **that flag was never read anywhere in the entire codebase.** The author built the guard and then didn't wire it up.
2. The confirmation step was skipped for a slot that looked empty — "nothing here to lose." But the confirmation is *per slot*, while the write replaces the **whole 512-byte sector** containing all four partition entries. So "nothing to lose" was only ever true about the slot, never about the table.

**What I changed.** Those two outcomes are now genuinely distinct (`HMB_BAD` / `HMB_OK` / `HMB_NONE`). If the table couldn't be read, the tool says **"Cannot read the partition table"** and refuses Format, Delete and Install — a blank table is a fine thing to *edit* and never a fine thing to *commit* when we don't know what it replaces. And free slots now confirm too. I confirmed this live on a completely blank disk: it still asks.

---

### 5. The tool could write one disk's partition table onto a different disk

**What could have happened.** Two hard disks. You open the disk tool with drive 0 selected, and click Format — it asks you to confirm. Before clicking again, you click drive 1's row in the Control Panel behind (which stays clickable — there's no modal blocking it). You click Format again to confirm. **Drive 0's partition table is now written to drive 1's first sector**, destroying drive 1's partitions, and the format runs over an area calculated from the wrong disk's table.

**Why.** The tool loaded its table when the window *opened*, but re-read "which drive is selected" at the moment of each *click*. Nothing closed the tool, rescanned, or cancelled the pending confirmation when the selection changed underneath it. A second route needed no confirmation at all.

**What I changed.** The tool and the installer now bind to a specific device when their window opens (`hd_tdev` / `hd_idev`) and every destructive operation uses **that**, not whatever is highlighted now.

---

### 6. A failed disk write was reported as a success

**What could have happened.** A sector goes bad mid-write. The driver says "done." os8088 believes the file's data is safely on the disk, and proceeds to commit the directory entry pointing at it — so you get a file that exists, has a name and a size, and contains garbage.

**Why.** The driver checked for an error at the **start** of each sector. For a *read* that's fine — a failed read never delivers data, so it's caught. For a **write**, the data has already been shifted out to the drive by then, and its completion status was never looked at. Every write in this driver is a single sector, so the only sector's outcome was the one never checked.

This matters more than it first appears: os8088's whole crash-safety design ("write the data, then the FAT, then the directory entry last") is **only safe if a failed data write actually reports as failed**. Otherwise the ordering protects nothing.

**What I changed.** One status check after the command completes, testing both the error and drive-fault bits. One extra poll per command, not per sector.

---

### 7. Roughly one memory layout in 64 would have failed every hard disk write

**What could have happened.** On some boots — and consistently, for that whole session — every format or install fails with a disk error. On other boots it works fine. Nothing you did differs.

**Why.** A rule this project already documents: PC hardware moves disk data in a way that **cannot cross a 64KB boundary in physical memory**. The only thing that reliably prevents it is starting the buffer on a 512-byte boundary. Two of this driver's buffers weren't aligned, and the driver stores the buffer address verbatim without the normalisation the kernel's own disk code does. Because the driver is loaded into whatever memory is free at the time, whether you got a bad address was effectively a coin toss weighted about 1-in-64 — and it stuck for the session.

There was also a fallback that made it worse: when the calculation said "zero sectors fit before the boundary," the code forced it to 1 and issued **exactly the straddling transfer the check exists to prevent**. Its comment said that case couldn't happen — true only for an aligned buffer.

**What I changed.** Both buffers are now `align 512` (verified in the assembled output), and the zero case is a hard refusal, so any future misalignment fails loudly instead of intermittently.

---

### 8. Mounting an IDE drive left the Control Panel showing the wrong button

**What could have happened.** You mount a hard disk. It mounts — the drive icon appears on the desktop — but the panel still shows a **"Mount"** button as though nothing happened.

**Why.** A low-level routine destroyed the `BP` register, which its own documentation promised not to touch (and which its sibling routine correctly preserves). The Control Panel was keeping the panel's position in `BP` across the mount. It came back as a garbage value, so the repaint was silently clipped away to nothing.

**What I changed.** Two instructions — save and restore `BP`, matching the sibling. No memory was ever corrupted by this one; it's a broken contract that other callers were escaping by luck. I confirmed the fix live — the button correctly reads "Unmount" after mounting.

---

### 9. Shipping a copyrighted IBM BIOS ROM — the only item that can't be undone after merge

**What could have happened.** The repository publicly redistributes IBM's copyrighted software. Every clone, fork, mirror and release tarball redistributes it further.

**Why.** The branch adds `tools/martypc/roms/BIOS_IBM5150_27OCT82_1501476_U33.BIN` — 8,192 bytes, the genuine 27 OCT 82 IBM 5150 BIOS. It's the **only binary in the entire 143-file changeset**. The root `LICENSE` is a blanket MIT grant over the repository, which is a promise this project has no authority to make about IBM's code, and `CONTRIBUTING.md` states plainly: *"No vendored third-party code... the whole tree is MIT under one license file."* Merging would make that sentence false.

`tools/martypc/README.md` carved it out in prose — which honestly **records** the problem rather than resolving it. A sentence in a subdirectory doesn't qualify a repository-wide licence grant. (Notably, `docs/MARTYPC-DEBUG.md` shows the project rejecting the IBM/Xebec controller *for exactly this reason*.)

**This is the only item where merging closes the door.** A squash merge copies the blob into `main`'s tree and history; removing it afterwards requires rewriting history, by which point it's already been distributed.

**What I changed.** Deleted it, gitignored `tools/martypc/roms/`, and made `build.sh` print the filename and its md5 when it's missing so anyone can drop in their own copy. **The instrument still builds** — three of the eight machine configs use GLaBIOS, which MartyPC bundles freely. The provenance and checksum stay documented.

---

# The six smaller ones (also fixed)

**10. `OSAPI_FILE_FIND` corrupted the caller's memory if used as documented.** The routine handed back a pointer advanced 14 bytes, while the published example loop sets the buffer once and iterates. Follow the documentation and your write position marches through your own program. Every current caller happens to reload the pointer, so nothing breaks today — but **this slot's contract is frozen forever at merge**, so it was worth the one-line fix now. No existing code can depend on receiving the wrong value.

**11. Note Pad's Find & Replace could delete your text and then refuse.** It deleted the matched text *before* checking it had room for the replacement, contradicting its own "nothing changed" guarantee. Reachable without any memory shortage: a note at the 16KB limit refuses *any* longer replacement. Worse, the failure left the cursor past the end of the document, and the next keystroke walked 65,535 bytes backwards through it. Now the room is secured first, so a refusal really does change nothing.

**12. `ASSOC.DAT` was trusted about bytes that were never read.** The file-icon cache on a floppy declares how many rows it contains. Those counts were checked against this build's limits but **never against how much of the file actually existed** — so a 512-byte file could declare 2,160 bytes of rows, and os8088 would parse uninitialised memory, or *the previous disk's cache*. Also, only the low half of the file's size was read, so a 66,048-byte file looked like 512 bytes. Both now checked; a bad cache is ignored cleanly. (This one is a hostile-input issue: every byte off a removable disk is untrusted by this project's own rules.)

**13. A drawing routine could paint over the mouse cursor.** `gfx_lstep` was missing the call that hides the deferred cursor, which its own near-identical twin has. The result on mono displays is cursor pixels baked into window content and a stale cursor image painted back over your drawing.

**14. Going fullscreen in Tracker could kill the music for the whole session.** The worker task that feeds audio checks a flag before drawing — but a worker that had *already passed* that check would be waiting for the graphics lock, which the fullscreen bracket then holds for the entire session. It never reaches the audio feed, the buffer drains, and the music stops until you exit. Using a menu made it near-deterministic. Fixed with a bounded drain before the bracket arms. (I hit a real bug writing this fix — my first version clobbered the flags register — caught and corrected before commit.)

**15. `docs/KERNEL-MEMORY.md` advertised seven times the headroom that exists.** `CLAUDE.md` sends people to that file *before spending memory*, and it claimed 3,584 bytes / seven steps of budget where there were 512 / one. It also stated the heap offset as 86.0KB when the build says 89.0KB. The auto-blessing tool provably can't fix these — they sit outside its markers. Corrected, and I added a line telling readers to run `tools/kernsize.py` rather than trust the prose.

---

# Things I'm flagging rather than fixing

**Your memory budget is now exactly full.** After rebasing onto your latest nine commits, the build reports:

```
kernsize: footprint  KERN_SIZE 90,112 of KERN_BUDGET 90,112 -> 0 spare (0 steps)
```

Zero bytes. That's from your recent commits, not mine — my fixes added 137 bytes and crossed no rung (430 bytes still free on the image rung). But **the next feature of any size at all needs a budget decision**, which per `CLAUDE.md` is a conversation, not a build fix. Flagging it because it's the kind of thing that's much easier to know now than to discover mid-feature.

**One fix isn't fully exercised by any test.** Item 1's *deep* case — where the FAT window has actually slid, i.e. clusters above 2,304 — needs about 4.6MB of writes to reach, which I couldn't drive through the UI in reasonable time. I'm confident in it for two reasons: the write path now calls **the identical routine the read path uses**, which every disk operation exercises constantly; and I added a refusal that fails loudly rather than corrupting if the range ever does leave the window. But I'd rather say so than imply coverage I don't have. A soak test that fills a hard disk volume past ~5MB would close it.

**One error message got less precise.** Making failed FAT writes propagate means an I/O error during allocation now surfaces as "Disk full" rather than "Disk error". Safe — it refuses the write either way — but slightly misleading. Worth a follow-up if it bothers you.

**Two design observations, not defects.** The two-click confirmation ("click Format again") is genuinely good — no modal dialog exists for a driver, and a second click on the same button can't be mistaken for the first. And the commit ordering in the write path (data → FAT → directory entry last) is correct and well-reasoned; items 1 and 6 were both cases of that good design being undermined from underneath, not of the design being wrong.

---

# How this was verified

Not just "it compiles":

- **Builds clean** at `-w+error` (which makes any warning fatal), rebased onto `3c1b2e5`. `checkdocs` reports 0 problems across 491 headings and 103 API slots.
- **I verified the ABI claim mechanically** rather than trusting it: all 71 pre-existing `OSAPI_*` slots compared between `main` and this branch — **none renumbered, none removed**. The PR description says 25 new slots; it's actually 32. Not a problem, just noting the description undercounts.
- **Boots on VGA and CGA** (the mono path matters for item 13).
- **`tests/filetest` gate: 25/25 PASS**, with `os88disk.py --verify` confirming the image afterwards.
- **Full hard disk exercise under QEMU** on a 40MB IDE disk: partitioned it, formatted a 31M FAT16 volume, mounted it, and created a folder. Then inspected the raw image from the host and confirmed: partition entries 2–4 **untouched**, **FAT1 identical to FAT2**, one cluster allocated with a correct end-of-chain marker, no cross-links, and **nothing written outside the live FAT area** — which is precisely what the old arithmetic in item 1 would have violated.

The two behaviour changes visible on screen: on a completely blank disk the tool
now says **"Erase slot 1? Click Format again"** instead of formatting on the
first click (item 4), and after mounting an IDE drive the panel correctly reads
**"Unmount"** instead of still offering "Mount" (item 8).

---

**Bottom line:** items 1–9 should not merge unfixed, and item 9 is the one that can't be corrected afterwards. All fifteen are fixed in `a776fed`, pushed to this branch. Happy to split any of them into separate commits if you'd rather review them independently, and equally happy to be argued out of any individual fix — the reasoning for each is in the commit message and in comments at each site.

jggonz:
### Follow-up: `5f104d9` — the "Disk full" imprecision I flagged

Closing the loose end from my review above.

**The problem.** Making failed FAT writes propagate (item 1) had a side effect I flagged but hadn't fixed: both callers of `dskw_alloc` reported it as **`FERR_FULL`** — *"Disk full"*. So a machine with thousands of free clusters and a marginal drive would tell you the disk was full, which sends you off deleting files that were never the problem. The write was correctly refused either way, so this was never a data risk — just a wrong answer to the user.

**The fix.** `dskw_alloc` now says *which* failure it was: `CF=1` with `AX = 0` means no free cluster, `AX = 0xFFFF` means the mark could not be written.

The part worth a second look: `dskw_mkbody` routes the I/O case to **`.wfail`**, which is the *rollback* path rather than merely the error path. That matters because of SPEC.md §18.4.1 rule 2 — a half-marked cluster must not survive in RAM where a later, unrelated write would flush it. Sending it to a plain error return would have left exactly that.

**Testing note worth recording:** `tests/filetest` does **not** cover `mkdir`, so the gate passing 25/25 says nothing about `dskw_mkbody`. I checked that path by hand instead — created a folder on drive B, confirmed it listed and sorted correctly, and ran `os88disk.py --verify` clean afterwards. Might be worth adding mkdir/rmdir coverage to the gate at some point, since it's the one write operation with no automated check behind it.

No rung crossed; the budget is still at 0 spare as noted above.

PR #64, opened 2026-08-09
[EXPERIMENTAL] FreeDOS: hand the machine to a foreign volume, and a floppy to boot it from

jggonz:
Opening **FREEDOS** on the system disk warns that the session is about to end, and on a second click tears the machine down and chain-boots the FreeDOS floppy in drive B. Typing `OS8088` at the DOS prompt brings the desktop back. `SPEC.md` §59 is the contract.

Verified end to end in QEMU: desktop → double-click `FREEDOS.O88` → warning → Start → `B:\>` → `OS8088` → desktop. `Special > Restart` regression-tested separately and still reboots through the ROM.

## Why a handover and not a window

DOS is a *boot* of this machine, not a program running on it, and on an 8088 it could not be anything else: one privilege level, no V86 mode, no MMU, no I/O trap. The only interception the hardware offers is an IVT entry, which catches `int n` and not `mov [0xB800:di], ax`. A windowed DOS would work for `DIR` and `TYPE` and fail *silently*, with a corrupt screen, for every graphical game — which is the category that matters. Giving DOS the whole machine is what makes real DOS software behave the way it was written to.

The address collision settles it independently: FreeDOS loads `KERNEL.SYS` at `0x0060`, which is `KERNEL_SEG`. os8088 cannot stay resident underneath it even in principle.

## Two repairs the handover needed

Both present as bad media rather than as themselves:

- **The diskette parameter table.** `dsk_dpt_init` points `0000:0078` at `KERNEL_SEG:dsk_dpt` and saves nothing. Harmless while *we* are the next thing to boot — our boot sector rebuilds it. Fatal when a 70KB `KERNEL.SYS` lands on `0060:0000` while the BIOS is still reading through that table. It goes back to `DPT_AT` first, the address `boot/boot.asm` had already reserved for exactly this.
- **The stub cannot run in place.** A boot sector must be entered at `0000:7C00` = `KERNEL_SEG:0x7600`, inside our own `.text`. Whether `ui_handoff` sits in that range is a link-order accident, so it is copied to `0x0500` and jumped to.

The teardown itself is `ui_cmd_reboot` with only its last instruction swapped — the ordering is load-bearing and a second copy would be a second copy to get wrong.

## Budget

161 bytes of `.text`, 1 of `.bss`. **268 bytes of image-rung slack left, no rung crossed, `KERN_BUDGET` untouched.**

The posted unit is stored as `unit+1`, so `0` — the `.bss` default, and what `ui_reboot_post` writes — means "restart through the ROM". Without that, drive 0 and "nobody asked" would be the same byte and `Special > Restart` would become a chain-boot of the system disk.

## Building the disk

`make dos`, on demand because the first run fetches a ~148MB toolchain. **No Docker and no Linux:** Open Watcom V2 ships native Apple-Silicon host binaries in `armo64/`, and `XCPU=86` gives the 8086 kernel. Two upstream modifications are needed and both are silent when wrong:

- nasm needs `-Dowlinux`, or `segs.inc` quietly picks the non-Watcom segment layout and links a subtly wrong kernel that still builds.
- FreeCOM's macOS rule passes `-DGCC`, which rewrites a Watcom-linked `command.com`'s stack pointer and gets it rejected as `Bad or missing Command Interpreter`. Upstream bug; worth reporting.

`tools/freedos/README.md` is the account of both. Patches are applied to *copies* under `build/`, never to the sibling checkouts.

FreeDOS cannot be a package — `APP_MAX_SIZE` is 60KB and `KERNEL.SYS` plus `COMMAND.COM` are 158KB — so it gets a volume of its own via `os88disk.py --dos`.

## Boot drive and the way back

No patching needed: `DL` at the boot sector is the whole protocol, so entering the drive-1 sector with `DL=1` lands the user on `B:` with `FDCONFIG.SYS`, `COMMAND.COM`, `%COMSPEC%` and `AUTOEXEC.BAT` all resolving there.

The way out is `OS8088.COM`, **not** `EXIT.COM`: `EXIT` is a FreeCOM internal, an internal shadows a `.COM` of the same name, and for a primary shell `EXIT` is defined to do nothing — so `EXIT` silently returns to the prompt. Measured, not theorised. It warm-boots via `0040:0072` rather than `int 19h`, because `int 19h` resets no hardware and a DOS game leaves the PIT, the 8259s and the CRTC wherever it liked.

## Machines

`make xt-dos` and `make 386-dos` put the DOS floppy in the second drive in place of the apps disk; every other machine keeps the apps disk. `xt-dos` is `ibmxt86` at 640KB, not the 256KB `ibmxt` — FreeDOS plus FreeCOM leave ~130KB free at 256KB, which is a prompt with nothing to run at it.

Under QEMU the same arrangement is `make test TESTAPPS=build/dos360.img`, which needed no new target.

## Not yet verified

**Whether an XT BIOS reports two floppy drives in the equipment word.** If it does not, FreeDOS marks B: single-logical and every access raises *"Insert diskette for drive B:"*, which would make the arrangement unusable on the XT. QEMU always reports two, so it structurally cannot show this — `make xt-dos` answers it in one run. The fix (`fdd_0X_type = 525_2dd`) and the reasoning are in `docs/TESTING.md`.

`tests/chainb` is the capability gate for the part that is not os8088 machinery: a 512-byte sector that chain-boots unit 1 and does nothing else.


lindoran:
I think this is the right way to handle dos. Since this is very much a 16 bit OS, sticking to real mode / XMS is really the right way to go.  Even if we can support a 386 protected mode the scope is blown away if we try to support literally everything from dos.  Doing anything else gets us into having to support EMS, system calling convention and essentially unloading huge parts of the os to support.   A small command interpreter would be nice for things like in system development, but that too is a big ask because the command shell is 100% graphics.  You get into the devil you already know which is windows 3.11.  

The key to that is to do a bunch of iterative work to figure out what the shell is for, what the command structure looks like. I don't know if that's necessary for it to be dos/posix like, in fact the refreshing part is there is no command interpreter and it can be literally anything if its ever done

pentagram666:
I had basically the same idea yesterday, but I didn't find this PR at the time. My idea is a little different — and perhaps a bit more thorough — especially regarding integration and the user experience.

If you have a chance, could you take a look?

[Discussion #134 — Native DOS Launcher / Transparent DOS Application Compatibility](https://github.com/jggonz/os8088/discussions/134)

PR #66, opened 2026-08-09
Elendilon -> Main

Elendilon:
Hard Drive Installation Fixed
Notepad performance


jggonz:
## Review: memory safety, wasted cycles, and file integrity

I reviewed this branch against the three things you asked about, and pushed fixes for what
survived verification. Five commits on top of `107f4df` — the tree builds clean, `checkdocs`
passes, and the kernel grew **34 bytes of `.text`** with no rung crossed and `KERN_SIZE`
unchanged (`2,560` spare, as before).

Every finding below was found by one pass and then handed to a second, independent pass whose
only job was to *refute* it. Eight claims did not survive that and are listed at the bottom, so
you can see what was considered and dropped rather than missed.

---

## 1. The raise cache never actually worked — it blitted heap onto the glass

**This is the one to look at first.** `gfx_save` takes the buffer in `ES:DI`. `gfx_restore`
takes it in `ES:SI`. That asymmetry is real and documented in `vga12.inc`, and it is the one
place the two primitives disagree — `menu.inc`'s restore does `mov es, ax / xor si, si` and gets
it right.

`wm_su_try` set up the *save*'s register:

```asm
    mov si, di                  ; DI is the buffer offset from here   <-- backwards
    mov es, [wm_su_seg]
    xor di, di
    ...
    call gfx_restore            ; ...which reads ES:SI
```

So `SI` still held the **window pointer**, and every cache hit read from
`cache_seg:<window_record_offset>` — tens of KB past the end of a claim that is ~21KB for a
default Note Pad. It blitted whatever heap follows the claim into the content rect. And because
`wm_su_try` then returns `CF = 0`, `wm_draw_win` jumps straight to `.growbox`: neither the white
fill nor `W_PAINT` runs, so **nothing repairs it**. The window sits there full of garbage until
something unrelated forces a repaint.

I reproduced it A/B in QEMU rather than reasoning about it. Same build, same click sequence
(open `README.TXT`, click the Disk window to cover Note Pad, click Note Pad's title bar to raise
it), the only difference being those two lines:

| | content rect after the raise |
|---|---|
| PR head as-is | **79.5% non-white** — colored noise, no legible text |
| with the fix | **23.9% non-white** — the document, correct |

Fixed in `d77d5b4`. Worth knowing what this implies: because the cache has been broken for its
whole life, **the §11.96 measurements in `docs/HANDOFF.md` were taken against a path that was
producing garbage.** "0 `W_PAINT` calls on a covered raise" is still true — that is why the
garbage persisted — but the wall figure the HANDOFF says is owed should be re-taken now that the
restore puts real pixels back.

### Four more in the same feature, all found once the restore actually ran

**`wm_su_kb` could ask for less memory than `gfx_save` writes.** It is `menu_save_kb`'s
arithmetic — but `menu_save_kb` is safe only because `MENU_MAXW`, `MENU_POPMAX` and
`[vid_popmax]` bound every factor, and its build-time assert is the proof. A *window* rect has no
such bound. A zoomed Note Pad on VGA is 81 byte-columns × ~398 rows × 4 planes = **128,952
bytes**, and the second `mul` by `[vid_planes_w]` discarded `DX`. Both multiplies are checked
now, and 64KB or more is a **refusal**, not a bigger claim: `gfx_save` advances `DI` contiguously
across the whole buffer and `DI` is 16 bits, so that save cannot be asked for at any price. The
refusal path is the one that was already there — the raise repaints, exactly as on a machine with
no room.

> Trade-off you should know about: on VGA this means a window much past half the screen now gets
> no raise cache at all. Lifting that means teaching `gfx_save` to cross a segment, which is a
> change to a core primitive and not something a review fix should do. I noted it in §11.96.

**The bank was taken after something that can draw over it.** `wm_su_take` sat below
`menu_draw_bar` and `wm_dock_under`, on the reading that neither draws a window. `wm_dock_under`
does: a window hanging over the dock strip (which `ui_grow` permits and `wm_dock_clear` reports)
sends it through `wm_dmg_wins`, which redraws every marked window **whole and unclipped** — and
`wm_lift` has already made the incoming window frontmost, so it is drawn *last*, over the
outgoing one. The bank then held the coverer's pixels inside the covered window's rect, and
`wm_su_ck` cannot see it: the four rect words still agree. Hoisted above both chrome calls.
`wm_su_rect` excludes the title bar, so it records the same rect wherever it runs.

**`OSAPI_WM_OBSCURED` now drops the cache.** §11.3 gives a background painter two ways to be
correct — arm a clip region, or *ask* and decline — and only the first came through
`wm_clip_set`. A painter that asks and is told "nothing is over you" then draws **unclipped and
unannounced**, which is exactly the state the cache is a promise about. Note Pad's worker takes
that route for all four of its background draws. The question is now the notice.

**Note Pad's sowed-scroll branch was the one worker draw that didn't ask.** It called
`np_redraw` unconditionally, so a covered window painted over whatever was on top of it — a §11.3
violation on its own, independent of the cache.

---

## 2. `mem_claim`'s shed-and-retry restaged garbage — §50.6

The three entry points stage the DMA head in `SI` and the direction in `DI`, in registers on
purpose (a global written before the `cli` is another task's to overwrite in the gap).
`mem_claim_1` copies them into `[mem_dma]`/`[mem_dir]` inside its window — and then destroys
both: `DI` walks `mem_tab`, `SI` counts the owner's claims and then walks it again.

The split that made shed-and-retry possible left the `pop di` / `pop si` in the **outer** half. So
every retry — which is to say every path this feature exists to rescue — re-entered `.go` with
`SI` and `DI` holding `mem_tab` offsets.

Concretely: bottom-up data claim, heap tight, raise cache live. The shed succeeds, and the retry
places the claim **top-down, in the region arena** that `mem_pg_ceil` was written to keep whole.
Neither existing guard catches it — `mem_tab`'s offsets are non-zero and under `0x1000`, so the
DMA page test waves them through. Fixed in `cc8234c`.

---

## 3. The installer formatted before it knew it could finish — data integrity

`hd_inst_sys` takes the copy buffer before it formats, and the comment says exactly why: *"a
format we cannot follow with a copy is a wiped partition."* But it only tested that it got **a**
buffer. `hd_ibuf_get` asks for `HIW_KMAXKB` (96KB) and, when the heap cannot supply that
contiguously, **silently falls back to `HIW_CHUNKKB`** (32KB) and returns success.

That is right for `hd_inst_apps` — every file it copies is ordinary and chunks. It is wrong for
`KERNEL.SYS`, which is written whole because §18.4.4 refuses a chunked path to a protected file
and §52.10.2 needs it contiguous from cluster 2.

So the install proceeded on the small buffer: `hd_inst_fmt` wrote the partition entry, the FAT
and the root — **the disk the user pointed at is gone at that moment** — and only then did
`hd_icopy_one` reach `.toobig` and refuse. Wiped partition, failed install, and every retry wipes
it again. A fragmented 640KB heap reaches this as easily as a small machine does, because Install
is a Control Panel click with the desktop live.

It now requires the full `HIW_KMAXKB` before the format and refuses with `hd_icopy_one`'s own
message — the same sentence, said before the damage instead of after it. The `.toobig` arm stays
as belt-and-braces. Fixed in `a9782d8`.

---

## 4. Two Note Pad correctness bugs, and one flags bug

**`np_ckword` was off by one, in the unsafe direction.** Its own header says the earliest index a
keystroke can have touched is `[np_cur] - 1` — and then it bounded the scan at `[np_cur]`. The
difference is precisely the case the test exists to catch: the character an insert added may
*itself* be the space that now ends the word, and then the word was **longer** when the break in
front of it was taken. Type a space into the middle of the long word on a wrapped row and the row
gets seeded at a start the insert has just moved. (A caret at index 0 now refuses outright — the
decrement would otherwise wrap to `0xFFFF` and accept every terminator on the row.)

**`np_paint` left `[np_resume]` set.** It cleared the flag on the way *in* but not on the way
out, so after §27.13's `np_xseed` it returned with the flag set and `[np_sdr]`/`[np_sdi]` still
loaded. Every other `np_walk` site spends it. The next walk that seeds nothing of its own —
`np_measure`, which `np_hmove` reaches bare whenever `[np_ckok]` is 0 — resumed at the top of a
view it was never asked about. `np_xseed` also clears it on refusal now, matching `np_seedck` and
`np_seedrow`.

**`dsk_bpb_bank_put` documents "every register AND the flags preserved"** and then opens with a
`cmp`, so its floppy exit handed the caller the compare's flags. `pushf` goes first now, like
`dsk_bpb_bank_zap`'s and for the same reason it gives.

Both in `7d44a85`.

---

## What I found but deliberately did **not** change

**Left/Right arrows still pay `np_redraw`'s unbounded pass 1** — this is your `HANDOFF.md` §3.2,
and it is real: `np_fastcm` parks `[np_mvbot]` at `0x7FFF` for kind 4, so a Left or Right lays
out from the caret's row to the bottom of the view (~13 rows × ~6 ms ≈ **78 ms**) purely to be
told nothing moved. `Up`/`Down` get the bound from `np_move` and are spared it.

The obvious fix is to derive the bound from `[np_ckpr] + 1`, since one index crosses at most one
row. I did not apply it, because `np_fastcm`'s gate only proves the caret is *at or after* the
checkpoint row's start — not that it is *on* that row. If it can be several rows below,
`[np_ckpr] + 1` is too shallow and pass 1 stops before the caret's row, which leaves stale
signatures and un-repainted rows on screen. That is a worse failure than the 78 ms, and proving
it either way needs the A/B you ran for `np_move` (six identical ArrowLeft states, `pass1_ms` and
`draw_ms` on both builds, `draw_ms` unchanged). **Worth doing — it is the largest remaining win
on the arrow path — but it wants a measurement, not a review.**

---

## Claims that did not survive verification

Listing these so you can see they were considered:

- *Use-after-free: the purgeable block can be shed during `wm_su_try`'s restore.* The mechanism is
  real in the abstract, but `mem_shed_one` has exactly one caller (`mem_claim`'s retry) and every
  `mem_claim` in the tree is a UI-task action or made under the gfx lock — and a raise is both. I
  wrote the invariant into §50.6 instead of adding code, because **a package worker that claims
  memory directly would break it**, and that is the rule worth recording. (It is also the deeper
  reason a worker grows through `OSAPI_MEM_REGROW`, which never sheds.)
- *`gfx_save`'s `DI` wraps at 64KB with no guard.* Same trigger, bit for bit, as the `wm_su_kb`
  overflow — fixing that fixes this.
- *`np_hmark` wipes the whole §27.13 row index on every character.*
- *`wm_su_take` banks on the AL = 0 title-only raise.* `wm_front` asks `wm_obscured` first, so
  `wm_draw_win` never runs on that path.
- *`wm_grow_paint` redraws over pixels `gfx_restore` just put back.* The premise that the bank
  always carries a drawn grow box is false on a reachable path, and skipping it would leave the
  box invisible-but-clickable — the exact failure it exists to prevent.
- *The retry sheds on refusals shedding cannot fix* (size 0, > 1MB, `MEM_OWNER_MAX`) — that is
  §50.6.2 as written.
- Two more on raise-cache cost that priced a path the code does not reach.

---

## Testing

- `make` clean from a fresh `make clean`: builds, `checkdocs: 516 headings, 105 slots, 0 problems`.
- Booted `make test` and drove it over QMP: desktop, Disk A, `README.TXT` into Note Pad, cover +
  raise (the A/B above), 25 × `Down` scroll, and typing into a wrapped paragraph — the reflow
  re-decides correctly and the chrome is right after the reordered `wm_raise` (menu bar owner,
  pinstripes, both dock tiles).
- **Not tested:** Hercules and CGA. Several of these touch drawing and §39.4 says to look at a
  1bpp adapter before calling it done — `wm_su_kb`'s `[vid_planes_w]` path in particular behaves
  differently there (one plane, so the 64KB refusal essentially never fires). Worth a
  `make test VIDEO=herc` pass before merge.
- The installer and allocator fixes are reasoned and build-verified but **not** exercised on a
  disk image; `HDD=<MB>` with a deliberately fragmented heap would be the test for the first.

PR #73, opened 2026-08-10
Elendilon -> Main (Disk I/O 2.5x less, Toasts, Extended Desktop, Format Floppies, Optimizations)

Elendilon:
* Disk I/O cut another 2.5x ish, 10x total since starting optimization
* Toasts
* Quality of life menu updates
* Extended desktop dual screen support
* Format floppies
* Debugging - martypc harness updates, vga/various machine presets on martypc, sysbench +4 tests, installer logging
* HD driver running memory footprint shrink
* Window draw from shadow under buffer (partially working)
* MAKE option to bake in a font

Extended Desktop


jggonz:
Reviewed the branch, merged `jggonz:main` into it, and pushed six commits on top. Summary:

## Merge (`ddd952b`)
Both branches claimed SPEC.md §59 — main for Frotz, this branch for toast.inc. Frotz keeps §59 (CLAUDE.md, the Makefile and the story tooling already cite §59.x); **the toast section is now §60**, with all ~50 references renumbered across the kernel, apps and docs. The only other conflict was the `.PHONY` list.

## Fixes pushed (each commit message carries the full reasoning)

**`b8b2178` disk** — the §18.95 cache could serve the mount's own LBA-0 read, so after a quiet remount the swap detector re-signed the *old* disk's boot sector out of RAM and a write laid the old FAT onto the swapped-in floppy. The mount and format-probe reads now bypass the cache. Also in the probe: the AH=08h parse had three faults that combined into "every drive is 9-sector" (`inc ax` guard that can't fire, `xor ah,ah` clobbering the `jz`'s flags so SI never took the drive's spt — an 18-sector drive formatted 1.44MB media as 720KB — and DI kept the DPT pointer the BIOS returned). A failed speculative fill no longer fails a read of a perfectly good sector.

**`4d0bf16` toast** — `toast_show`'s copy loop only bounded characters *kept*; a package string with no NUL and no printable spun forever with interrupts off (hard hang). Bounded at `TOAST_SCAN` bytes examined. `toast_now`'s "UI task running + flag set ⇒ UI task holds the lock" inference was unsound under pre-emption; `gfx_lock` now stamps an owner byte and `toast_now` compares. It also banks the caller's clip region and pen across its draw, `toast_kill`/`toast_owner_gone` close their check-then-act windows, and TTLs ≥ 0x8000 no longer expire instantly.

**`f510947` drivers** — `DSV_SIZE` grew 24→26 but `drivers/debug/debug.asm`'s service table was never padded, so `drv_publish` captured `dbg_up`/`dbg_busy` as a phantom `DSV_CPCLOSE=1`: closing the Control Panel with DEBUG.DRV attached far-called offset 1 of the image (two reviewers found this independently). The installer's mount thunk read the *live* panel highlight `[hd_sel]` instead of the banked `[hd_idev]` — with two disks and the Format window opened mid-install, Install formatted device A and copied the system onto device B. The pre-format `HIW_KMAXKB` buffer gate is restored (its removal made a low-memory Install wipe the disk on every retry), `hd_svc_mount` no longer leaves the panel selection moved, and `hd_dev_row_ck` makes the tool-seam's "no such device" check real.

**`dd8cbcb` notepad** — three guards added to main in bb30cfd (#66) were missing here (the branch predates that merge in places): the `np_ckword` `[np_cur]-1` bound, the sowed-debt redraw's `OSAPI_WM_OBSCURED` gate, and `np_paint` spending `[np_resume]` after its walk. All restored verbatim, plus `np_xseedi`'s refusal clearing `[np_resume]`.

**`6a86a58` wm/vga/mouse (extended desktop)** — `wm_raise` banked the outgoing window's save-under *after* `wm_dock_under` could draw the incoming window over it (order restored to what main ships); `wm_zoom` was display-aware in x only (snap path overwrote the display origin with absolute 7; y/h were the primary's band, so a BELOW-layout secondary window zoomed onto the primary); `gfx_line`'s clip-fragment loop never translated virtual fragments into display-local space (clipped diagonals vanished on display 1); `wm_paint_all`'s fullscreen shortcut left the whole secondary stale; the save-under bank now refuses on an extended desktop (`gfx_save`/`gfx_restore` have no display hook — it banked/restored the wrong band of the wrong monitor); two pointer fixes (dead-zone clamp aimed at frozen `[cur_disp]`; `kbm_move` borrowed the ISR's `[mou_nx]` staging word).

## Known issues left open (didn't want to decide these unilaterally)

- **`vid_cga_alias`**: the new probe reads different bytes at B000:7F00 vs B800:7F00 as "two cards". A Hercules that decodes its full 64KB regardless of the page bit reads as a phantom CGA, and the comment the PR replaced said MartyPC's Hercules is that card. Worth re-testing `marty` before trusting Extend's offer there.
- **`DSV_SIZE` grew without a driver-header version bump** — any out-of-tree driver with an exactly-sized table gets a phantom `DSV_CPCLOSE`. In-tree drivers are all padded now; consider making `drv_publish` zero-fill instead of trusting the driver to over-provide.
- **`drv_attach`** now banks *any* `DRVV_READY` refusal into `DRVR_ERR` (not just `DRVE_TWICE`) — flagged in case the error caption on a still-working driver is unintended.
- Resident `hd_mount_one` doesn't define CF on every failure path (`hd_vol_of` answering 0xFF self-corrects the visible result); `vid_disp_init` sets `[vid_cur]=0` without republishing `[vid_ox]/[vid_oy]` (correct today only because `gfx_unlock` forces display 0); `TRK_RING_SM`'s feed margin at 44 kHz wide rates looks thin but small-ring machines are unlikely to run 44 kHz.

Build is clean in both kernel variants (all size guards hold, `checkdocs` passes), and the merged branch was boot-tested in QEMU: B: mounts, folders navigate, Note Pad launches, File → Save writes the file and shows the `Saved NOTES.TXT` toast with the new bounded copy and owner-checked lock.

jggonz:
Second review pass over the merged branch. **One blocker, one small real bug, three stale comments — all three pushed** (`4ba548a`, `7a99e26`, `ad1f65e`). Nothing else I could find rises to blocking, and I did a fair amount of work to try; the detail of what I checked and cleared is at the bottom, because "I looked and it was fine" is only worth reading if it says where.

## Pushed

**`4ba548a` pixcheck — the blocker.** `cache_for()` is new in this branch and reads `wm_su_seg` and `wm_su_win`. The same branch deleted both: §11.96.3 gives every window its own cache, so the pair became `wm_su_segs` (a word per record) and the owner word went away — `wm.inc` says so itself, *"The slot IS the window, so `wm_su_win` is gone"*. `os88sym.linear` raises `KeyError` for a name that is not in the map, so this does not answer wrongly, it takes the whole check down on the first call, `--self-test` included:

```
wm_su_segs  0x646b
wm_su_seg   KeyError: no kernel symbol 'wm_su_seg'
wm_su_win   KeyError: no kernel symbol 'wm_su_win'
```

That matters more than a broken script usually would, because `pixcheck.py` is the instrument §11.96.2 names and that `wm_su_ck`'s own comment points at — *"which is exactly what tools/notepad/pixcheck.py does to force a full repaint"*. The round of work that changed the raise cache is the round that left its verification unable to run, and the file's own header is the argument for why that is not a small thing: *"a check reporting zero means nothing until something has shown it can report something else."*

The test is one non-zero word now; the slot is inverted out of the pointer rather than read from a third symbol, so `win_geom` stays the only place that knows `WIN_SIZE`. I audited every kernel symbol the tools and tests name (`.sym("…")` across `tools/` and `tests/`, resolved against the built map) — those two were the only casualties.

**`7a99e26` disk §18.95 — a sector served from the cache does not step the bar.** Both progress bars are scaled in *sectors of the job* — the boot splash off `dsk_read_chain`'s count (§15.3), the file-activity widget off `fpg_begin(AX = the file's sectors)` (§12.8) — and both are stepped from `dsk_xfer`'s `.count`, once per sector an `int 13h` moved. `.served` advances DI/SI/BX and jumps straight back to `.sector`, so every sector the cache answers is one the caller received and the bar never counted, and the bar stops short of the end of an operation that finished. Bounded by the pollution gate (a long coalesced run never reaches the cache), so what is lost is a file's tail and any chunk a directory walk left over its head — small, but "stalled short of the end" is the one shape a progress bar must not have. Fixed with `.count`'s own loop; both steppers preserve everything and are a compare-and-return when their bar is down, against the ~1.6 ms/sector `dsk_copy_seg` has just spent. Ten bytes, no rung crossed in either kernel (big 319 → 309 left on the image rung, small 179 → 169).

**`ad1f65e` three stale comments**, two of which are the *reason* for a decision:

- `apps/hello/hello.asm` declines `WF_SAVEU` because *"there is ONE cache for the machine … taking it for this would be taking it from something that letters."* §11.96.3 retired that bound in this branch — the cache is per window and ranks `MEM_PG_TRIV`, so it is taken from nobody. The honest half of the sentence (the repaint is one fixed string) is what it says now.
- `apps/os88api.inc` and `drivers/hdd/cfg.inc` both tell a driver author that SYSTEM.CFG is written when *"the Control Panel closes or Special > Restart is picked"*. §12.3.1 renamed Special to Builtins and moved Restart to the System menu. The `os88api.inc` one is the SDK — the copy least likely to be checked against the source.

## Left alone, flagged

- **`MEM_P_DIRW` is claimed through `mem_claim_dma`, which places it bottom-up.** `mem_claim` says a purgeable tag *"places ITSELF … there is no third entry point to call by mistake"*; `mem_claim_dma` sets `DI = 0` and jumps to `mem_claim.go`, so the 63KB read-ahead window is the first purgeable claim to skip that and land at the bottom of the heap. The ceiling still applies, and I convinced myself it is harmless — regions go top-down and are unaffected, ordinary claims shed it and retry, and the freed hole is at the bottom where it coalesces — so it is possibly better than the rule rather than worse. But the comment now describes something that is not true of every purgeable claim.
- **`dsk_totsec` is `TotSec16` only.** A volume declaring 0 there (FAT16 past 32MB) makes `dsk_rah_len` answer 1 and the cache never fills. BIOS-backed volumes only, so it is the boot-partition case and it is performance, not correctness — a fill can only ever be shortened by this.
- **Verdicts are transient now.** §60.5 moves every `dskw_*` failure from the per-window status line to a 3-second toast. That is the design and it is written down; worth naming that a failed delete or copy is now missable in a way it was not, and that `toast_show` refuses outright while `[fpg_on]` is set.
- The **`DSV_SIZE` 24 → 26** note from last round still stands: in-tree drivers are all padded, and there is no way to make an out-of-tree one safe without a header version bump, which would refuse it outright. Fine as it is, worth knowing.

## What I checked and cleared

Built and boot-tested both kernels, all three geometries, on all three adapters.

- **Format (§18.96/22.12) end to end.** 1.44MB in a 1.44MB drive and 360KB in a 360KB drive, on VGA and on Hercules. Both images come back `os88disk.py --verify OK`, with the right SPC/root/TotSec/media/FATSz per row, `F0 FF FF` / `FD FF FF` at the head of both FAT copies, a zeroed root, and the data area untouched. The probe picks the right row from `AH=08h` + the descending read in both drives.
- **The toast (§60) on all three adapters**, including the inverse-video strip on 1bpp — `Formatted B:` on Hercules and `Settings Saved` on CGA both letter and read correctly, and the `menu_bpadc` clamp is what keeps a wide one from spinning.
- **The raise cache against a full repaint, on Hercules.** Covered a `WF_SAVEU` window and raised it, then ran the identical sequence on `main` and compared the framebuffer: the restored pixels are byte-identical to main's genuine repaint, edge bytes included. §11.96.2's overhang merge is doing its job. (The 2px overhang visible right of a *newly created* window is on main too — pre-existing, out of scope here.)
- **The §18.95 cache against data.** Copied a 15,889-byte file A: → B: through the file manager on the patched kernel and diffed it on the host: identical, destination fsck clean. Also audited the coherence rule — `dsk_rah_wr` hangs off the one routine every changing sector passes through, no kernel path writes around `disk_write`, driver volumes are never cached, and a full mount flushes.
- Frotz survived the merge intact (`apps/frotz/` untouched, all ten make targets present, 60 references live).
- Read the diff for `disk.inc`, `diskw.inc`, `wm.inc`, `memory.inc`, `driver.inc`, `menu.inc`, `ui.inc`, `files.inc`, `ctrl.inc`, `vidsel.inc`, `viddet.inc`, `vga12.inc`, `font.inc`, `icons.inc`, the HDD split and the app changes. The `mem_claim_1` parameter fix, the `mem_shed_one` priority ladder, `mem_pg_own`'s range rows, `wm_fit`'s stack discipline, `menu_bemit`'s two-run split, the `FMC_*`/`CMD_*` renumber against `fm_menu_base`/`ui_loc_base`, and `np_append`'s signature inversion all check out line by line.

PR #80, opened 2026-08-13
Elendilon -> Main (Performance, os88net + file redirectors, save under 90% done, extended desktop stabilization, QOL)

Elendilon:
** Decisions **
* Kernel split into kern_big and kern_small, different builds via define tags
-- Kern_small is currently more aspirational than realistically small
-- Will do another de-sizing pass on both of them after getting NE2000 and heap compaction
* Removed prohibition on invalidating ABI and WM slots, as it was growing complexity unnecessarily. These should come back once the kernel state is more stable.

** Performance** - Kernel Cost ~4k
* Massive avg 40% mono render performance increase, more in places like paint's full draw on a textured bitmap (20x)
* All windowed apps WF_SNAP to the horizontal 8 pix grid by default - rendering performance gain for the few that were not already snapping on mono. This now applies to VGA also, because it turns out it gets the same performance benefit
* 2.5x mono glyph render speed improvement, and VGA gets 1.5x from this one too. This isn't a duplicate from the last PR, but rather, 2.5x OVER the last PR.
-- Not all apps are converted. Note Pad, Task Manager are still the primary users.
* First VGA scrollx/blit pass, ~10-20% general VGA render performance that isn't just incidental from mono passes
* Shadow Under V2 (More complete than incomplete)
-- Minimize, restore, drag, call to front, dmg rects in progress but not finished

** Drivers / Core Features** - Kernel Cost ~6k
* On adapter mode switch apps dynamically resize for CGA mode, and back
* Kernel allows apps to command cursor replacement. Missile command windowed doesn't draw two cursors.
* Generic kernel driven UI element library (this plus the mouseup interactions below allow button press states next pass)
* Extended desktop draw fixes (various gfx routines)
* File management API for kernel and apps becomes a redirector
-- File manager, file save/load dialogs and all app accessible apis now view files in a generic way, with no view of the underlying hardware
* RamDisk driver for testing (simple, no config yet, usecase of the redirector)
-- This will get a control panel page and XMS support next release
* Remote filesystem via a null modem parallel cable (laplink standard)
-- Run a dos program, os88net, on the dos computer. Link up and then copy/paste/run/etc files on the dos system's drive(s) inside os8088's file manager
* API and backend set up for mTCP forwarding over parallel
-- Generic redirector, will be able to accept input from a NE2000 driver immediately
-- No app surfaces this yet
* Live detection of available floppy drives, not static to 2
* Support for 3rd/4th floppy drive
* Task manager live heap view page added

** Quality of Life ** - Kernel Cost ~2k
* Double click title bar to maximize
* Extended desktop's second desktop trades bar space for window space now
* Click the dock icon of a focused window to minimize it
* Screensaver. Static 5 minutes, no config, no animation for this first pass.
-- This is actually to save my screen, because I accidentally walked away and left my 5151 on for an hour ><
* About box carries commit version
* Toasts reworked to center and not draw over the menu items/file progress bar (uses the clock area instead)
* Mouseup button interactions, instead of mousedown
-- Select actions, drag actions, double click actions and game actions not affected

** Kernel Size Management ** - Kernel Shrunk ~8k
* On-demand kernel modules
-- Load control panel, format from disk on demand, instead of paying for them with RAM all the time ~5kb
-- Limited to actions that either require the system disk anyway, or are infrequent tools
* XMS became a loadable driver. No checkbox. Loads automatically at boot if and only if XMS is available on the system ~1.5KB
* Double buffering removed, ~1.5kb

** Apps **
* Tracker can open files as big as the available heap space (No XMS support yet. XMS has no users at all yet, still, even though the bug with it is fixed)
* Added click to zoom to fractals
* Fractals drawing optimization pass (not calculating it, just putting it on the screen)

Known issues:
* File progress through the redirector became bytes, but bytes are not always available
-- This will become an icon with no bar, or return to being an empty/paused bar like it was before
* Many apps still force full redraws, thus invalidating their save under cache, when they don't need to
-- File manager is the biggest offender


jggonz:
## Review of this PR — and 42 fixes now pushed to the branch

This is a big change (193 files, +73k/−12.6k, 579 commits), so I reviewed it in 14 slices — kernel core, window manager state, window manager painting, graphics, UI chrome, the file redirector, the network driver, the other drivers, the app SDK, the apps, the build system, the docs, a hunt for anything main had fixed that the branch dropped, and a sweep for mechanical rule violations across the whole diff.

Every suspected bug was then handed to a second reviewer whose only job was to **prove it wasn't real**. That killed a good number of false alarms. **44 survived**, and I've fixed all of them and pushed the fixes onto this branch, so you can just merge.

Two encouraging things first: the branch builds clean on both kernels, and **the regression hunt came up almost empty** — the only thing this branch lost from main was one code comment. Given 579 commits and this many merges, that's a good result.

Below is what each fix was, in plain terms. If you only read one section, read the first one and the "Three decisions that are yours" at the end.

---

## 1. The serious one: two Disk windows were sharing the same memory

**What was wrong.** This PR grew the Disk window's saved state from 24 bytes to 57 bytes, because each window now remembers its own folder path. But the table that hands out those blocks of memory was never told — it still spaces them 24 bytes apart.

**What that means.** The system reserves room for four Disk windows at positions 0, 24, 48 and 72. Each window then writes 57 bytes into its slot. So window 1 writes straight over the top of window 2's memory, window 2 over window 3's, and so on. The safety check that's supposed to catch this only checks the pool against itself, so it waved everything through.

**What you'd see.** Open two Disk windows, browse them to different folders, and they corrupt each other's remembered path. Since the path is used to find files on disk, that's the kind of bug that eventually eats a file rather than just looking wrong.

**The fix.** One word: the table now says `FS_SIZE` instead of the hard-coded `24`, so it can never drift again. I confirmed the right number actually reached the compiled kernel (the table entry now reads 57, not 24), and then booted it and drove two Disk windows into different folders side by side — before the fix they clobbered each other, after it they don't. `SPEC.md` §22 and §29.3 were pinning the old `24` too, so those were corrected.

---

## 2. Bugs you would have seen on the screen

**The mouse cursor and a redraw flag were the same bit.** Two new features in this PR were each given "the spare high byte" of the window flags — the per-window cursor shape, and the flag that marks a window's pixels as out of date. They're the same storage, so each silently set the other. In practice: create a new folder while two Disk windows show it, and the second window's mouse pointer turned into Missile Command's aiming crosshair. In the other direction, Missile Command asking for its crosshair permanently marked its own window stale, which meant its window could never use the save-under cache — so every drag or click-to-front repainted the whole game. Fixed by moving the flag to a bit the cursor shape can't reach, and making both sides read and write only their own bits.

**Three drawing routines were handing back damaged registers.** The new second-monitor support added a hook to `gfx_xor_rect`, `gfx_save` and `gfx_restore` — but the hook was placed *outside* the part of the routine that protects the caller's registers. All three are public, documented as "changes nothing", and callers trust that. Fixed by widening the protection to cover the hook, being careful to leave the two pointer registers those routines are *supposed* to advance alone.

**On a two-monitor setup, damage on the second screen was never repaired.** A new routine works out how much of a window still owes a repaint, and it was *replacing* its answer with the primary monitor's dimensions instead of *widening* it. On the "Below" layout every window on the second screen produced a repair rect that didn't overlap the window at all, so those pixels stayed broken until something else happened to redraw them. Fixed with the minimal change — the strip now only ever grows the rect, never shrinks it. Single-monitor behaviour is bit-for-bit unchanged.

**Greyed-out text flickered when clipped.** The checkerboard pattern that stands in for "grey" on the mono adapters restarted from the same phase on every cell, ignoring which row it was actually on. A greyed menu item that was partly clipped would visibly flip its dither pattern. Fixed by seeding the pattern from the row.

**The event queue could be reset while the mouse was writing to it.** The screensaver (new in this PR) clears the input queue when it wakes the screen, using a routine that was only ever meant to run at boot, before interrupts are live. If a mouse click landed in a two-instruction window during that reset, the queue's head and tail were left permanently out of step — and from then on every input event returned the *previous* event, for the rest of the session. Low odds, but it never recovers. Fixed by blocking interrupts around the reset.

**The toast message was never actually centred.** A clamp meant for one case was being applied in both, so the "eye-centring" this PR designed was always clamped away. Now the clamp only applies when the progress widget is actually on screen.

**Task Manager drew a row outside its own window.** The memory view and the heap view share one "how many rows fit" limit, but the heap view fits more rows than the memory view does. On a short screen the memory view would place a row in a third column that doesn't exist — drawn a full column-width past the right edge of the window. Fixed with a bounds test so a row can never be placed in a column that isn't there. (Worth knowing: on the default screen this lands past the edge of the *monitor* and is invisibly clipped, which is presumably why it was never spotted. It's visible on Hercules and on CGA.)

**Paint filled its full-screen background twice.** Entering full screen called the bed-fill routine, then immediately called repaint, which fills the bed again as its first act. The first fill used a stale damage rect. Removed the redundant call — this is exactly the "double-draw flash" that emulators hide and real hardware shows.

---

## 3. Bugs on the parallel cable

The new networking is careful work — every wait is bounded, the handshake is properly interlocked, and I checked all fourteen commands field-by-field against the DOS-side program: **the frames match**. The bugs were all in what happens when something goes *wrong*.

**A read-only refusal killed the link instead of refusing.** When the DOS side is running `/RO` and os8088 tries to write a file, the server sent its "no, read-only" byte back **while os8088 was still sending the file**. Both machines were talking at once on a cable that only works one direction at a time, so the two handshakes cancelled out, no byte crossed, and the link died. You'd get "Disk error" and a dropped connection instead of a read-only message. Fixed by having the refusal be *recorded* and sent after the transfer has finished draining — which meant fixing five places, since three other commands relied on the old early-send behaviour.

**Any keypress on the DOS machine silently killed the link.** Only Escape was meant to. The wait loop treated *any* key as an abort, and it swallowed the keystroke on the way — which also meant a real Escape couldn't stop the program either. Now only Escape aborts, and it actually stops the program.

**A "folder not found" reply left the cable out of step.** For the LIST command the server always sends a status byte *and* a count; on a refusal, os8088 read the status, treated it as a dead link and never read the count. The stray count then sat on the wire and the session ended. Now the refusal reads the count and reports "not found" properly.

**After the cable is pulled, every click froze the UI for ten seconds.** The driver knew the link was gone, but no file command checked before going to the wire — so each one waited out the full reply timeout. And because the volume stays on the desktop, you could keep clicking it. Fixed by checking the link state once in the two shared entry points (so every command, including any added later, inherits it), and by dropping the timeout once a loss is known.

**Two commands treated an ordinary "no such file" as a dead cable** and tore down a perfectly healthy link. Now they report the error and keep the connection.

---

## 4. Quietly broken in the build

**The kernel size report and the "BUILT WITH A KNOB" warning had stopped running.** A new rule was inserted between the kernel's build step and its reporting lines, so the reporting became part of the *new* rule instead — and that rule is normally already up to date, so it never ran. On an ordinary build both printed **zero times**. That warning exists specifically to stop a debug-knob kernel being cut into a release, so it's worth having back. It was also hiding how tight the size budget has become (see below). Moved back where it belongs; it now prints exactly once.

**`make field` was broken.** It asked a sub-build for the field kernel's on-demand modules using paths that sub-build had no idea how to make. It failed with "No rule to make target 'build/cgak/ctrl.drv'". Fixed, and all five field disks build again.

**`make KFZ=1` wouldn't compile at all.** The new freeze-diagnostic heartbeat pushes the kernel 512 bytes over budget — but there's an existing allowance for instrumented builds that just hadn't been extended to cover it. I widened that allowance rather than touching the real budget: the shipped kernel's guard is untouched, and the diagnostic you added for a field freeze can now actually be built.

**A missing `ret`.** The `mod_init` startup stub fell straight through into `loader_init`, which then ran twice. One byte.

**Two build knobs weren't stamp-tracked**, meaning changing them could boot the *previous* configuration — the failure mode your own CLAUDE.md warns reads exactly like the feature being broken.

**Nothing checked that an on-demand module actually fits** its ceiling. It would build, split and ship, and only be refused at boot. Now it's a build error.

**A stale budget symbol** (`KERN_SMALL_BUDGET` said 99,328; the real value was 102,400) plus a new assertion so the two can never disagree again.

**`XMEM.DRV` was shipped on the small-kernel disks**, which have no code to load it. Harmless, but it's disk space on a 360KB floppy.

**A tool hard-coded `/home/user/os8088`** and couldn't run on any other machine.

---

## 5. The file redirector

**Copying a folder off a network volume silently produced an empty folder.** When a driver doesn't support "list this folder" — and `NET.DRV` doesn't — the answer it gives back is indistinguishable from "this folder is empty". So the copy created the top-level folder, put nothing in it, and reported success. Fixed by asking the driver whether it supports the command at all before descending, and refusing honestly if it doesn't. (One note: I put that check on the *source* volume. The original suggestion put it after the code switches to the destination drive, which would have missed the case of copying *from* a network volume *to* a floppy — the exact case that matters.)

**Every folder on a redirected volume drew as a generic application icon.** The icon pass handled files and skipped folders, and the fallback for "no icon" is the application icon. Now folders get the folder icon, same as on a normal disk.

**Progress bars never appeared for writes to a redirected volume** — only for reads. Now armed for both writes and appends.

---

## 6. Documents that no longer matched the code

In this repo SPEC.md is the contract, so where it disagrees with the code, the next person sizes something from the document and ships a bug. In every case below I checked the code first and made the document match it, never the other way round.

- **The SDK promised something that is no longer true.** `os88api.inc` still states that any app built against main's SDK runs unchanged on this kernel. That's no longer so: slot `0x01F0` used to be `OSAPI_GFX_DBUF` and is now `OSAPI_WM_ONMOUSEUP`. Reusing the slot is legitimate and properly recorded — but the blanket promise had to become an explicit exception, because a main-era app disk booted against this kernel would call the wrong thing. **Every `.o88` needs rebuilding against this SDK.**
- **The window record.** SPEC still pinned `WIN_SIZE = 26` and didn't list the new `W_ONMOUSEUP` field. The real record is 28 bytes. SPEC itself warns that a size change here "has broken this tree before".
- **The driver interface table.** SPEC pinned 13 verbs / 26 bytes; the shipped interface is 14 verbs / 28 bytes, and there's no length check anywhere — so a third-party driver written from the spec would have had its 14th slot read from whatever memory followed it, and called if non-zero. Both in-tree drivers are fine.
- **Slot `0x01E8`** was documented as retired in four places while the SDK publishes it as `OSAPI_VOL_KIND`.
- **The Control Panel item table** listed a page that no longer exists and gave one constant three different values in a single section.
- **The back buffer** (removed in this PR) was still in the binding primitives table.
- **`FPG_GAP`** was pinned at 8; the code changed it to 0.
- **`NET-PLAN.md` opened with "Nothing here is implemented"** while `NET.DRV` ships on every disk.
- **The memory baseline was re-blessed** so `make` reports real drift again (see below).
- One code comment in `hello.asm` had reverted to a rule main had corrected — restored.

---

## Three decisions that are yours, not mine

**1. The kernel has 512 bytes of headroom left.** This is the number the restored size report was hiding. `kern_big` is at **104,448 of 104,960 — one 512-byte step**, against the four-step standard your own memory doc grants. This PR raised both budgets (98,304 → 104,960 and 94,208 → 102,400) and moved the footprint +1,536, crossing two rungs — meaning the machine's RAM usage moved twice without the baseline being re-blessed. I've blessed both baselines so drift shows again, and I deliberately **did not raise any budget** to fix anything (the `KFZ` fix widens an existing debug-build allowance instead). But the next handful of bytes added anywhere trips the guard, and CLAUDE.md is explicit that raising it is your call, not a build fix.

**2. The RAM disk and the network volume can no longer be used at the same time.** Both register as the same driver class, and there was no check — so whichever attached second silently took over the first's dispatch. The first driver's volume stayed mounted and browsable on the desktop while every operation on it was answered by the *other* driver. I fixed it by refusing the second attach with a visible error on the Drivers page. That's strictly better than a silent misroute, but it is a real limitation and both drivers ship on every image. The proper fix is to route by volume rather than by class — which the code's own comments already name as future work. **This is a stop-gap, and you should decide whether to do the real thing before release.**

(Related: while verifying that guard, the agent found the refusal path *reached* a second bug rather than avoiding it — unloading one file driver deleted the *other* driver's volumes, because volumes were being dropped by class. It saw this happen on screen and fixed it too.)

**3. The "driver can't do that" refusal reads as a permission error.** There's no error code that means "this driver doesn't support this operation", so the folder-copy refusal uses the protection error, and the file manager will word it as a permissions problem. That's what the kernel already does elsewhere for the same situation, so it's consistent — but it isn't quite the right message.

---

## About the overlap with your latest commits

You pushed four fixes while this review was running, and one of them — `82cf28c`, the window slot handing its band to the next tenant — is **the same bug my review found**. The rebase applied both fixes, which was harmless (they wrote zero over zero) but cost 8 bytes. **I kept yours and removed mine**, because yours is the one verified on a cycle-accurate 5150 and SPEC.md is written around it. Your other three commits don't overlap with anything here.

## How this was checked

Every fix was applied by an agent that then had to verify its own work before the next one started. `make` passes, `make KERN_SMALL=1` passes, `make checkdocs` reports 0 problems, and `make field`, `make small` and `make KFZ=1` all build again. The kernel footprint is unchanged at 104,448. Where a fix had a visible symptom it was checked by booting in QEMU and driving it — the Disk window fix was verified A/B against the unfixed build with a scripted double-click session, and the cursor bug was reproduced and confirmed gone (that one needed a *file operation* to trigger, not a Refresh — a Refresh clears the flag and gives a false pass, worth knowing if you re-run it).

Two things I did **not** do: I didn't touch anything outside the confirmed findings, and where a suggested fix turned out to be wrong on contact with the real code, the agent did the right thing instead and said so rather than applying it blindly. Nothing was declined as a false positive.

PR #92, opened 2026-08-17
Elendilon -> Main (Heap compaction, Calculator, Cyclone 88, Arkanoid controls/flashing, Bounce flashing, Herc detection bugfix)

Elendilon:
87 commits of the integration branch, squashed.

TWO THINGS RENUMBER MAIN'S OWN WORK, and both are the expected shape of a squash rather than a conflict to resolve (docs/UPSTREAM.md): the branch's numbering becomes main's.

  SECTIONS. main's §65 Word and §66 TeXPad arrived on a branch that already
  had §65 Calculator and §66 Heap compaction, so Word is §68 and TeXPad is
  §69 here, and §67 is Cyclone 88. This is §52's precedent exactly - when
  ModPlug landed, main's §52 ModPlug became §52 HDD plus §56 ModPlug.

  SLOTS, and this one invalidates every .o88. main published
  OSAPI_GFX_BLIT1 at 0x03F0 and OSAPI_VOL_SYS at 0x03F8; those cells were
  already OSAPI_KEY_DOWN and OSAPI_FSX_SURF on this branch, so the pair
  moved to 0x0418 and 0x0420 and the table is 131 slots. Note how it
  presents: the two sides added different NAMES, so apps/os88api.inc merges
  CLEAN with two cells pointing at each other's addresses and nothing says
  so until a package calls one and gets the other. The check is mechanical
  and belongs after every merge in either direction - sort the %defines by
  address, look for a duplicate. Verified rather than asserted: word.bin
  built here differs from the same source built against main's SDK in
  exactly three 2-byte runs, all of them these two addresses.

NEW SLOTS: OSAPI_KEY_DOWN (§9.7, key state a real-time app can poll - int 16h has no key-up), OSAPI_FSX_SURF (§53.7.1, the rect a bracket owns, which a same-mode bracket used to hard-code and got wrong on a second display), OSAPI_MEM_MOVABLE / OSAPI_MEM_PARKSAFE / OSAPI_TASK_PARK (§66's compaction - the default stays PINNED, so a claim nobody declares behaves exactly as every claim did before).

KERN_BUDGET, kern_big only: 107,008 -> 108,544, move 24, attributed to the font system; docs/KERNEL-MEMORY.md carries the arithmetic. kern_small is untouched at 103,424 and was re-measured, not assumed.

Also here: the Calculator (§65) and Cyclone 88 (§67) as the sixteenth and seventeenth packages, NET.DRV enumerating and deleting FOLDERS across the cable (§62.10.6/.7), every About box crediting its contributor, two window-manager artifacts fixed with a reproducer each (`wm_dmg_bands` per display, `wm_chrome_clip` subtracting an occluder by its frame), and seven documentation slot references that named the wrong cell - each a VALID slot, just a different routine's, which is the one thing tools/checkdocs.py says it cannot catch.

Tested: make clean through make, zero warnings; checkdocs 0 problems; both kern_big and kern_small assemble; boots to the desktop on a cycle-accurate 4.77MHz 5150 with CGA and with Hercules; and Word opens WELCOME.DOC through the .DOC association and renders its styled proportional text on CGA, which is the renumbered gfx_blit1 doing the work on a 1bpp adapter.

Claude-Session: https://claude.ai/code/session_0112pXRhzdQR7ufkdyGSY9aw

### DIFF
--- a/.claude/skills/release-os8088/SKILL.md
+++ b/.claude/skills/release-os8088/SKILL.md
@@ -152,6 +152,23 @@ a real problem, not something to work around. Report the kernel size; if it has
 grown, say by how much and how much headroom is left (the ceiling is 0xA000 =
 40,960 bytes for image + bss).
 
+**The optional fifth image** is `build/apps-all.img` (SPEC.md 19.9): one
+1.44MB floppy with every application on it, Frotz and both Words included, for
+somebody who wants one disk rather than four. It is not built by `make`,
+because `cword` needs a compiler this tree does not contain:
+
+```bash
+tools/setup-cc.sh && make allapps
+ls -l build/apps-all.img
+```
+
+Offer it, do not assume it. If `tools/setup-cc.sh` cannot run -- no network,
+no host toolchain -- **release the four and say the fifth was skipped**; it is
+a convenience, and a release that waits on it is a release that does not
+happen. When it is included, boot it in step 3 like any other image (it goes
+in B:, `make test TESTAPPS=build/apps-all.img`) and list it last everywhere,
+so the four shipped images stay the obvious download.
+
 ### 3. Smoke-test the build before publishing
 
 Never publish an image that has not been booted.
@@ -457,6 +474,11 @@ gh release create "$VERSION" \
   --notes "<notes>" \
   build/os8088.img build/os8088-360.img build/apps.img build/apps360.img
 ```
+
+Add `build/apps-all.img` to that argument list **only if step 2 built it**, and
+last, after the four. A missing asset makes `gh release create` fail with the
+tag already pushed, which is the one failure in this whole procedure that
+cannot simply be re-run.
 
 `gh` infers the repository from the checkout's remote, so this works for a
 fork or a rename without any edit here.
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ vm/xt-z/nvr/
 vm/386-z/nvr/
 vm/xt-word/nvr/
 vm/386-word/nvr/
+vm/386-c-word/nvr/
 
 # Page snapshots the Playwright MCP server drops in the working directory
 .playwright-mcp/
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -74,7 +74,7 @@ exactly like the feature being broken.
 CGA and a Hercules, a monitor window each — and the only 86Box machine that
 can show §39.12–§39.19's extended desktop; it boots Single, and Control Panel
 → Display → Desktop is what extends it (§39.19.1). `xt-z` and `386-z` are the
-Frotz machines (§61.9), and `xt-word`/`386-word` are the Word machines (§65.5)
+Frotz machines (§61.9), and `xt-word`/`386-word` are the Word machines (§68.5)
 — the four that put a dedicated floppy in B: instead of the apps disk. `make
 zdisk` builds the story disk (`tools/getstories.py` fetches the stories, which
 are never committed) and `make worddisk` builds the Word disk.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -20,6 +20,7 @@ nothing here duplicates it — a second copy is a copy that goes stale.
 | **[docs/TESTING.md](docs/TESTING.md)** | concluding something is untestable — it is the matrix of what each emulator can and cannot do, with a recipe per capability |
 | **[docs/KERNEL-MEMORY.md](docs/KERNEL-MEMORY.md)** | spending any memory |
 | **[docs/HERCULES-TESTING.md](docs/HERCULES-TESTING.md)** | testing on Hercules — it *is* automatable, and all three ways of getting it wrong give you a black image rather than an error |
+| **[docs/C-TOOLCHAIN.md](docs/C-TOOLCHAIN.md)** | writing or building a package in C (§70) — how to install the compiler, the four C rules and what each refusal means, and what the language does not have here |
 
 ## Commands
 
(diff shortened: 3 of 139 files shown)


jggonz:
Merged `main` into this branch and pushed two commits to `elendilon-pr` (maintainer edit): the merge (`0512e73`) and a set of review fixes (`f671cb7`). The PR is now mergeable against `main`.

## Merge

Only `CLAUDE.md` conflicted textually. The real work was the §-numbering: `main`'s #89 landed the C toolchain as **§67**, which met this branch's §65 Calculator / §66 Heap compaction / §67 Cyclone 88 — so per docs/UPSTREAM.md the C toolchain is now **§70** and its Word/TeXPad references follow to §68/§69 (323 lines, scoped to the text `main` added). `checkdocs` cannot see a duplicate `## 67.` heading; UPSTREAM.md now says to `uniq -d` the headings after any merge. The C SDK reaches `OSAPI_GFX_BLIT1`/`OSAPI_VOL_SYS` by name, so the slot moves needed nothing there — verified by building `cword`/`cc-smoke`/`covl`/`chello`/`allapps` against the merged tree.

## Review findings fixed in `f671cb7` (all confirmed by reading the code; details and SPEC sections in the commit message)

| # | sev | where | what |
|---|---|---|---|
| 1 | high | `kernel/instance.inc` `inst_park_wait` | `mov cl,5` ran before `push cx`, so `OSAPI_TASK_ALIVE` returned CL=5 to a worker whenever a park was pending |
| 2 | high | `kernel/clip.inc` `clip_put` | banked the caller's segment in BP across `mem_claim` (which can compact); a movable Note Pad/Artful document moved and the copy read the hole. Source now pinned via `[mem_pinseg]`; `dsk_xfer` nests that word instead of zeroing it |
| 3 | high | `drivers/sound/sb.inc` | staging pool movable while `sbl_v_stage`/`sbl_v_read`/`sbl_fill_half` copy it on the *client's* task (Tracker's worker mid-verb-6), which no park asks. Every pool copy is now `SBL_DCHUNK`-chunked under `cli` re-reading `[sbl_poolseg]` — `sbl_fill_guard`'s idiom. Pool stays movable |
| 4 | high | `drivers/net/os88net.asm` | 13-byte wire name joined unchecked: `T`,0,`..\DOS` rmtree'd `C:\DOS` from a share of `C:\SHARE` (one-file escape pre-existed for the other write verbs). `path_join` is now the fence for every verb (`\ / : * ? < > | "`, leading `.`, empty → `FERR_NOENT`) |
| 5 | med | `inst_alloc` | `inst_parksafe[slot]` never cleared on reuse — a non-declaring package inherited PARKSAFE and could be parked in `gfx_lock` with a live pointer |
| 6 | med | `mem_reloc_call` | preserved only AX/BP/DS/ES while `mem_cp_run` reads SI right after; a proc written to the stated (`wm_pkgcall`) contract would move the fill point onto a live block. Banks everything now |
| 7 | med | `mem_compact` busy guard | refused only a nested *compaction*; a nested `mem_claim` from a reloc proc still first-fit into the fill area. `mem_claim`/`mem_free`/`mem_regrow` now answer CF=1 while `[mem_cp_busy]` |
| 8 | med | `apps/paint/paint.asm` | BMP save is one `OSAPI_FILE_WRITE` out of the movable canvas with the kernel holding ES:BX; `pt_pin`/`pt_unpin` around it (np_dmov's shape) |
| 9 | med | `kernel/driver.inc`, `drivers/os88drv.inc` | `FSV_SIZE` 28→30 with no version bump: a released NET.DRV/RAMDISK.DRV had `FSV_RMTREE` read out of its name string and far-called. Header word +10 is now `DRV_H_FSV` (table length; 0 = 28) and `drv_fs_call`/`drv_fs_has` refuse a verb past it. §62.9.1's table also gained the missing `FSV_RMTREE` row |
| 10 | med | `kernel/ui.inc` `ui_drag_phase` | "legal y's are an interval" is false on CGA-over-Hercules; a drop at y 193..199 could be phased into rows no display has, title bar included. Phased point re-asked with `vid_disp_find` on KERN_BIG |

Also: §66.5.4 and two `instance.inc` headers quoted the wrong slot for PARKSAFE/TASK_PARK; docs/KERNEL-MEMORY.md re-blessed for both variants — `kern_small` grew 100,864 → 102,912 on this branch (one step spare) and that spend is now written down. Footprints unchanged by the fixes (kern_big +126 bytes inside its rung).

Verified: `make` (checkdocs 0), `make small`, C toolchain targets, `allapps`; boots; `tests/heapfrag` 12/12; Tracker streams BEVERLY.MOD through an emulated SB16 with signal in the wav (the chunked path); Paint saves a BMP that verifies on the image (the pinned write).

## Left as follow-ups (low, or design questions rather than one-liners)

- `clip_get`/`clip_put` are documented worker-legal but the clipboard claim moves with no park check for a worker suspended mid-`rep movsb` (`kernel/clip.inc:36-38`, `mem_can_move` `.notslot`). No shipping caller is a worker today.
- `mem_can_move` passes slot-owned (built-in) claims without the worker test (`memory.inc` ~1043); only Disk's view cache is declared and Disk has no task, so latent.
- §66.3 rule 5's rationale says the sector cache is "taken lazily inside `disk_read`", but `dsk_rah_want` is only reached from mount/volume switch — the guard is right, the stated reason isn't.
- `srv_rmtree` tests children's attributes but not the target folder's own (a hidden/system folder gets emptied, then DOS refuses the final rmdir).
- `wm_dmg_gray` header still says "all registers preserved" but clobbers DI on KERN_BIG via `vid_ctx_ptr` (sole caller banks it); it also spends `cur_unlazy` before `wm_dmg_bands` can answer "nothing to dither".
- Cyclone's worker peeks/fetches int 16h non-atomically (`cy_kbdrain`): the UI task can take the head between them. Not memory-related.
- `heapcheck.py` and friends hardcode `/home/user/os8088` — pre-existing pattern in `tests/`, noting only.

PR #97, opened 2026-08-19
Elendilon -> Main (Networking + Browser, Button Down State, Generic Scrollbars, Ram Disk Config + XMS, Bugfixes)

Elendilon:
Browsing reddit, on a 5150, in extended desktop, connected via a parallel cable to a DOS machine (and then through the https proxy on my main box).

And there are 0 hardware concessions on this; the newest thing on the 5150 is the MOUSE from 1985.
-------------------------------------------------------------------------------------
* Buttons now have a down state, and track that state with mouse over
* All buttons are generic and get the mouse up action + down state
* Buttons with keypresses attached animate via a timer (aka calc)
* Scrollbars are now generic, instead of implemented per use
* Cyclone help fix, attract screen lists enemies
* Cyclones web damage bugfix
* File Progress bar tracks "operations" as well as bytes
* Ram drive is configurable
-- And can be mapped to XMS - the first user of XMS in the system
* Don't even try to copy files when there is not room to copy them
* Icon cache warming was out of date again; fixes .tex and newer associations
* A MAKE flag enables sound blaster for a PICOMEM in the sound driver
* BIOS will not beep itself to a stack overflow if you hold a key down in a slow app
-- Unless you try it during a file load or other total lock, nothing we can do about that one
-- Glabios already doesn't have this bug
* Apps that open a file on the system disk are also on the system disk, plus calc and one game (minesweeper)
-- All of them are still on the apps disk too
* Added mTCP forwarding over os88net parallel
* Added ethernet driver support, NE1000/2000 only right now
* Both sockets feed into applications, the first of which are Browser and Terminal
-- Decent amount of work into Browser, Terminal is still basic
-- A proxy app for https->http and formatting is available as python or a windows executable
* Split beverly.mod for 360k off into a "media360.img" disk
-- 720/1.2/1.44 are unchanged
* Tracker optimizations allow 11,000Khz on an XT
-- Barely. For MOST mods, it works. It takes the whole machine. If it falls behind, the extra cost of trying to catch up makes it fall behind again.
-- No fullscreen, there is no CPU room left to animate anything


jggonz:
Merged `main` into your branch and pushed **11 commits** to `elendilon-pr`. The PR is now **MERGEABLE** — it only needs a review approval.

The merge itself was clean (`main` had moved by one commit). Everything below is a defect found afterwards, in a review that split the round into ten subsystems, gave each one a reader, and then handed every finding to a second agent whose job was to *refute* it. 49 survived that and all 49 are fixed. One was refuted (a missing `cld` in the checksum — it turned out not to matter) and I downgraded one myself.

**First, the thing that isn't a bug.** `tests/ethernet.py` was reporting `FAILED` — and the stack it was testing is fine. §71.7 makes the location bar open holding `http://`, focused, caret at the end; the harness still clicked the bar (moving the caret into the middle of that `http://`) and typed the whole URL again, so it asked for `http:/http://10.0.2.2:8090/eth.htm/`. The browser refused it correctly and the gate blamed the driver. With the harness fixed the gate passes end to end — `ethernet: ok`, DHCP at 10.0.2.15, `br_nstate = 6`, 14 lines rendered, and the host's own socket recording `GET /eth.htm HTTP/1.0`. **That's your NE2000 stack fetching a real page, and it did it again after all 49 fixes were in.**

---

## The four that could destroy something

**`RAMDISK.DRV` wrote 4 KB over the interrupt vector table.** Choose "Xms" on the Ram Disk page and press Load: `rd_wbuf_get` lends `[rd_bounce]` whenever the store is *configured* as XMS, but `rd_load` calls it before `rd_store_get`, so the bounce is still 0 and `ES = 0000`. 4,096 bytes of the file went straight over the IVT and the BIOS data area — the machine gone from two clicks on a shipped panel. The fix guards the *lending*, not the preference, because the preference being settable before the store exists is deliberate and the page depends on it. Same edit covers its twin: the bounce is one extent (1–8 KB) and Preserve/Load were pushing 4 KB through it regardless, writing 2–3 KB past the claim on every save with nothing on screen to say so. `a841410`

**`OS88NET.COM` never handed its packet-driver handle back.** `ne_stop` — which calls `release_type` — existed and was called from nowhere; `/N` registered a receiver and then exited through `int 21h/4Ch` with it still pointing into the program's own image. DOS reuses that memory, and the next matching Ethernet frame far-calls into whatever landed there. The DOS box hangs or reboots seconds after you quit, with nothing connecting it to the program that caused it. `e59243a`

**The browser freed a fetch's buffers without stopping the fetch.** File > Open (or double-clicking a `.HTM`) while a page was still downloading: `br_load` calls `br_free_all` and re-claims, and the worker keeps writing into memory that has been handed back and re-claimed *smaller*. `br_abort` already existed ten lines away and `br_go` already used it — the open path just never did. `63bfba5`

**A stale `.DRV` had its name string published as four callable procedures.** `DSV_SIZE` went 28 → 36 with `DRV_VER` still 4, and `drv_publish` copies `DSV_SIZE/2` words unconditionally while `drv_check` validates nothing about the table's length. Copy the new `KERNEL.SYS` onto last release's floppy and one click on Control Panel → Hard Drive far-calls into the middle of that driver's data — and the fourth of those four bytes-turned-pointers is `DSV_PKGCALL`, the package fence, so any program on the apps floppy could reach the raw-sector door `HDD.DRV` was built never to open. The fix follows the precedent already in the tree: the header carries the table's length, the way it already carries `FSV_SIZE`. `f20b529`

## The network, off the wire

**One IP header with a total length under 20 handed the transport layers ~65,500 bytes.** `ip_in` bounds the wire length *above* against the frame and never below, so `sub ax, IP_HLEN` wraps — and any host on the network could get the browser or Telnet to render up to 1 KB of the driver's own transmit buffer (the previous frame: another connection's data, or the DHCP exchange) as if the far end had sent it. Two independent readers found this one. The same edit fixes the upper bound, which was 14 bytes too generous — the IP length doesn't count the Ethernet header. `018b29c`

**DNS.** A hostile `RDLENGTH` let the answer walk wrap below the message and parse the driver's own image as records — a visible stall on a real XT, possibly ending with the browser connected to an address read out of that code. And a reply was accepted from *any* source: guess the 16-bit query ID and you could send the browser to your own server.

**TCP.** A passively-accepted connection never reached ESTAB, so `NETV_LISTEN`/`NETV_ACCEPT` could not actually receive anything — the connection sat half-open forever or was dropped a fraction of a second after the peer connected. And an ACK ahead of `SND.NXT` moved `SND.UNA` past what was sent, so one malformed segment left the connection transmitting from a sequence number the far end throws away.

**`ETHER.DRV` had no re-entrancy guard**, where `NET.DRV` has an explicit `pushf`/`cli` mutex for the identical socket ABI. Browser open, press Renew on the Ethernet page, and the two tasks splice each other's frames, deliver one program's data into another's memory, and — in the wrong two-instruction window — overwrite the card's own hardware address, at which point the machine goes silent until the driver is reloaded. Fixed with *your* mutex: `eth_claim` is `nsk_claim` line for line, the loser answers `NETE_BUSY`, and I checked both shipping clients handle it — `brnet.inc:522` and `telnet.asm:933` already route it to a retry.

## Things you'd see on the glass

**`kern_small` had no release edge at all.** `cp_kinit` — which installs `wm_onmouseup` — was entirely inside `%ifdef KERN_BIG`, so on a 128–256 KB machine every button on every driver's Control Panel page looked pressed and then did nothing: Mount, Format and Install on Hard Drive, Renew and Set Up on Ethernet, Connect on os88net, each left drawn filled black. The guard was wider than it needed to be — `W_ONMOUSEUP` is at offset 26 and exists in *both* window records; only `W_ONDRAG` is `kern_big`'s. Now only the drag half is guarded. Shown A/B: without the fix, pressing Format leaves the button black and no window opens; with it, the Format tool appears. `84897d8`

**The file dialog's New Folder button.** Press it, slide off to cancel, and it's left a featureless black rectangle with no label until something else repaints the dialog. It's in `fdlg`, so it was every application's Save As. `3096192`

**The progress bar was scaling against a screen coordinate.** `fpg_begin` reads its `DX:CX` byte count *after* `fpg_arm`, whose own header says it clobbers CX and DX. Every save or load that raised the widget for the first time crawled about 2% across the trough while the machine sat frozen for the whole operation.

**Telnet repainted the whole terminal per line.** `te_scrollpaint` builds its blit width from DX with a screen coordinate still in DH — harmless on CGA's 200 rows, but with the window in the lower half of a VGA or Hercules desktop it's ~a second of visible redraw per line on a 4.77 MHz XT where it should be one scroll plus one lettered row.

**Note Pad's scroll-bar track was dead** — the call passed the y where `os88ui_sbhit` wants the block pointer, so a long file could only be scrolled a line at a time. **TexPad hit-tested garbage** whenever its window sat right of x=256 (`tp_bhit` indexed with a stale AH). **Piano's five buttons never drew the down state** at all, while still spending two full repaints per crossing to draw pixels identical to what was there.

**And a few that only bite later:** a page too large now says "Page truncated" instead of stopping mid-sentence under a status line reading "Ready"; the HTTP status code was parsed only for a session's *first* page; `make ramseed` no longer assembled; and the three new API slots answered CF only on `kern_small`, so a package asking `kern_big` "do you support timers?" got back its own leftover carry.

---

## What I left alone

**Your design decisions, all of them.** The three moved API slots (`0x0428`/`0x0430`/`0x0438` → `0x0450`/`0x0458`/`0x0460`), the §70/§71 → §73/§74 renumber, `DRVC_NET` going to ETHER, the two budget raises. I checked them rather than changed them: every in-tree caller of the moved slots moved with its symbol — I rebuilt the whole C toolchain, `cword`, `runcpm` and `apps-all.img` to prove it — and I diffed every RunCPM and C-toolchain file against `main` with § references filtered out, and not one functional byte changed. Nothing was reverted.

**No guard was relieved.** `KERN_BUDGET` and `KERN_CODE_MAX` are byte-identical to what you submitted. The fixes cost `kern_big` one 512-byte step of its *existing* headroom (2,048 → 1,536 spare, still 3 steps); `kern_small` is unchanged. Both variants are re-blessed.

**Two files removed, and this one is a judgement call rather than a defect:** `.github/workflows/os88proxy-exe.yml` and `os88proxy-release.yml`. They're well written and they work on your fork — but merged, they'd run in the upstream repo with `contents: write` and an unpinned third-party release action, which is the repository owner's decision to make deliberately. `tools/os88proxy.py` and `tools/build-os88proxy-exe.bat` are untouched, so nothing about the proxy itself changed.

**One thing worth saying plainly:** your `docs/UPSTREAM.md` addition — the script that finds two `%define`s at one address — is the single most useful thing in this round for whoever does the next merge. That collision is the failure that merges cleanly and reports nothing, and you wrote the detector for it unprompted.

## What I checked and how

`make clean && make` from clean, `make small`, both kernels under their guards with no rung crossed, `kernsize` blessed for both, all nine images `--verify` clean, `checkdocs` at 0 problems, no duplicate `§` headings (your branch actually *clears* two that were duplicated on its base), `git ls-files build` = 0. Booted on VGA and on CGA. `make ethertest && python3 tests/ethernet.py` → `ethernet: ok`. `make ramseed`, `make bench`, `make browsertest`, `make socktest`, `make drvcalltest`, `make allapps` (115 files, verifies clean). SPEC.md was updated in the same commit as anything it pins.

**What I could not run:** `tests/brtest.py`, `tests/drvcall.py` and the other MartyPC-driven gates — MartyPC isn't installed here, and I'd rather say so than imply I ran them. Worth a pass on your side before this merges.

Every one of these is in code that took real work, and the round is a lot of good work — a TCP/IP stack and a browser on an 8088 is not a small thing, and the photo at the top of this PR is the proof. Happy to walk through any of the fixes, or to back one out if you read it differently.

PR #103, opened 2026-08-20
Elendilon -> Main (Note Pad Save Changes on Close, Window Animations, Themes, More Extended Desktop Window Sizing)

Elendilon:
* Take action on window close is now implemented
-- First pass is for note pad only, when content changed only, show a "Save Changes?" dialog.
* Window open/minimize/maximize animations
-- Close has no animation as it affected how responsive the OS felt
* Themes in control panel - Bright (default), Dark, and Color (only VGA)
* Most windows now correctly adapt to mode changes
-- Both via control panel and via drag on an extended desktop
-- Window size, and when the package supports it, color/mode/shape/layout/etc as well
-- Tamegram has no concept of CGA and I didn't want to alter the gameplay, so it is the exception
-- Didn't test CPM as it pulled in after this was done
-- Did not expect this to take most of a day, but was quite involved
* A set of unit/functionality tests to run on builds/releases
-- Need to verify these actually do what we want, but the goal is to stop things like the C toolchain breaking with no one noticing
* Minor bugfixing across a few apps
* Don't try to make the windows executable of the python proxy every build


jggonz:
Reviewed, merged and pushed — **8 commits on top of yours**, and the PR is still `MERGEABLE`. Nothing of yours was reverted, and `main` needed no merging from my side: you'd already brought it in, and it's a strict ancestor of your head, so the whole thing reviewed cleanly against `main` as-is. Thank you for the reconstructed-base work in that first commit — I checked it and it holds. Of the 96 lines `main` added since the fork point that this branch removes, every one is a documented replacement, and **zero of 1,787 labels across 19 kernel files changed which `%ifdef` they sit inside**, which settles the guard question mechanically.

The features work. I drove the whole close negotiation by hand on both kernels — Cancel restores the screen exactly, Discard closes clean, Save opens Save As and quits after writing, and `/MEDIA/NOTES.TXT` came back off the floppy with the disk fsck'ing clean. The Dark theme's greyed menu separator is genuinely visible on a CGA, which was the thing most likely to have gone wrong. And `main`-built packages still run on this kernel unmodified, so the five new slots cost the ABI nothing.

Below is what I changed and why, in the terms it would have bitten a user.

---

## Fixed

**1. Every File → Open dialog wore the aiming crosshair.** `ed82880`

Open a file dialog in any application and the pointer turned into a crosshair over its contents and stayed there — on VGA, CGA and Hercules alike. Same for every notice window. I have before-and-after screenshots at the same pixel.

`WF_NOANIM equ 0x0100` is `W_FLAGS` bit 8, and the comment beside it says "bits 0-7 and 15 were taken". Bits 8–14 were taken too — by §7.2.1's per-window cursor **shape**, which is a field rather than flags, and shape value 1 is `CUR_CROSSSH`. The `WF_STALE` comment eleven lines further down describes this exact collision as something already found and fixed once. It ran the other way as well: `wm_cursor`'s read-modify-write kept only bit 15, so any package calling `OSAPI_WM_CURSOR` cleared its own opt-out, and Missile Command and Cyclone — which ask for a crosshair — quietly opted themselves out of your zoom for the life of their window.

The bit is **14** now, the top of the shape field for the same reason `WF_STALE` is 15, and the two places that touch that byte mask both kernel bits off. Bits 9–14 are provably free: `wm_cursor` and `cur_shape_set` both refuse a shape at or past `CUR_NSHAPE`, which is 2. `W_FLAGS` still clears it for free on a reused slot, which a side table would not have.

**2. A window the dock strip was drawn under sometimes kept the dock's pixels on top of it.** `ed82880`

...and sometimes the opposite: a window standing entirely to the right of the painted span took a full repaint it did not need — hundreds of milliseconds on a 4.77 MHz machine, for nothing.

`wm_dmg_bit` answers in `AX`, and `wm_dmg_wins` falls straight through from it into the new dock test, whose first compare reads `AX` as the rect's left edge. `main` tested `DX` alone at that site and `DX` survives the call, so the x-span test brought this with it. Two bytes: `push ax` / `pop ax`.

**3. The zoom played on windows that never left the screen.** `ed82880`

Double-click a drive whose Disk window is already open, or pick File → Task Manager when it's already up, or relaunch an app at its instance cap, or raise a second "Save changes?" alert — and an outline flew out of a dock tile and collapsed onto a window that had not moved. About 160 ms of gfx-lock-held stall per click on an XT. PERFORMANCE.md Part 5 prices a raise of the frontmost window at *"no window at all"*.

`wm_show` is deliberately the raise path too and says so in its header; it already computes "was this on the glass" three lines above the new `wm_an_open` call, which didn't consult it. `inst_unmin` was the other half — `inst_find_kind` and its two siblings match on `I_STATE` and never test `I_FLAGS` bit 0, so it armed a dock tile for an instance that was never minimized. The already-visible path **spends** the arm rather than skipping it, for the reason your own `wm_an_open` `.drop` gives.

**4. `make KERN_SMALL=1` did not pass.** `ed82880`

This is the second build CLAUDE.md asks for after every change, and it exited 1 on this branch. `KERN_SMALL`, `FSNOSTAMP` and `THEMEDARK` are in `$(VIDSTAMP)` but were missing from `$(KNOBS)`, whose own comment says *"EVERY OTHER KNOB IN $(VIDSTAMP) BELOW BELONGS HERE TOO."*

That used to cost only a silent banner. It costs the **build** now, because `all` runs `test-fast` when `$(KNOBS)` is empty and every symbol the suite reads goes through `os88sym`'s byte-identity check — so a knob kernel builds and then fails `api-abi`. (`THEMEDARK`'s own `KNOBS +=` was dead, wiped by the `:=` further down the file. That one's a Make footgun, not an oversight.) All four of `make`, `make KERN_SMALL=1`, `make THEMEDARK=1` and `make FSNOSTAMP=1` are clean now.

**5. The close negotiation's re-entry fence was disarmed from the first dirty close onward.** `35030c6`

`wm_ask_close` banked `[wm_clask]` in **CX** across the package callback, but `wm_pkgcall` restores `DS` and `ES` and nothing else — `ui_ptcall`, the sibling dispatcher forty lines away, pushes CX and DX onto the *stack* around the identical call for exactly this reason. `np_onclose` does clobber CX (it reaches `os88ui_ask`, which creates and paints a window), so after the first dirty close the fence held the package's leftovers and never read 0 again. §75.1 calls it "not decoration" — it's what stops a negotiator that raises an alert from re-entering a close path and recursing on a 256-byte stack. It's on the stack now, the way `[snd_inst]` two lines below it already was. `POP` writes no flags, so your negotiator's CF still reaches the branch untouched.

**6. ...and the deferred close was spent before the lock.** `35030c6`

`wm_close_pass` read `[wm_clreq]` and zeroed it, *then* called `gfx_lock`. §75.2's second guard is `wm_destroy` cancelling a request that names the window it's freeing — a write it makes with the lock held. So the gap between the two was precisely the interval that guard cannot cover. The read and the clear moved inside the lock; the lock-free part is still just the `cmp`, so it's one compare per idle pass as designed.

**7. The Theme page's radios never showed the pressed look.** `ed7b7ab`

Every other Control Panel page shows a control pressed while you hold it — that's how you know which one is about to act, and it's the whole reason the ladder calls `cp_ctl`. `CP_ITHM` simply wasn't in `cp_ctlpaint`'s switch. On top of that the two **Window Animations ids were reversed**: the ladder reports 5 for On and 4 for Off, and the painter drew row 0 as 4 and row 1 as 5 — so pressing one row drew the other pressed. Latent while the first defect kept the painter away, but reachable even then. Costs the kernel nothing: `ctrl.inc` is the driver.

**8. Three of the five new slots were documented at other slots' addresses.** `03817b7`

| section | said | is |
|---|---|---|
| §11.100.1 `OSAPI_WM_PREFER` | `0x0468` | `0x0478` (`0x0468` is `wm_onclose`) |
| §11.100.2 `OSAPI_WM_MINSIZE` | `0x0470` | `0x0480` (`0x0470` is `wm_close_req`) |
| §39.16.4 `OSAPI_WM_DISPLAY` | `0x0478` | `0x0488` (`0x0478` is `wm_prefer`) |

`wm.inc`'s own three routine headers repeated the same numbers, so checking the spec against the implementation gave agreement from both and still the wrong cell. No shipped package is affected — everything goes through the `%define` — but a package written from the spec would have called `wm_onclose` with a table offset in SI, installing whatever `AX` held as its close negotiator. `checkdocs` can't see this and says so at line 10 of its own docstring. I cross-checked every slot citation in the tree **by name**; the only other mismatch is in `docs/HEAP-COMPACTION-PLAN.md` and it's on `main` too, so I left it.

**9. Clicking the alert's message text re-lettered the whole button row.** `c86dffe`

A press landing on no button answered 0, `os88ui_adn` wrote back the same `0xFF` it already held, and all three buttons were redrawn for a state that hadn't changed. It now draws nothing when nothing changed — the rule your kernel twin states in its own header (`cp_setdown`).

**10. Arkanoid's height floor was one pixel into the dock strip on a CGA.** `c86dffe`

`ARK_PREF_SH` is 156 and was passed as the floor as well as the preference. On a CGA `wm_fit` clamps the frame to 155 — that's the `dec di` that keeps a frame off the strip — and `wm_min_floor` runs after that clamp and wins, so the drop shadow landed on row 176, exactly `vid_dock_y0`. From there `wm_dock_clear`'s `jae` is true for that window forever, so every window opened or moved over the game pays a dock repaint. `ark_entry` spends five lines of comment and a `sub bx, MBAR_H + 1` avoiding that very pixel. The preference keeps 156 — a generous height the clamp brings down is the documented idiom — and the floor gets `ARK_MIN_SH` = 155. A screenshot can't show this, because row 176 is the dock's own 1px black rule; I proved it against the constants instead.

**11. Two Disk windows could strand a format question over a file list.** `e91aa35`

`fm_2up` was a single global byte holding per-window state, and the Disk window is multi-instance. Window A puts up "ERASE and format B: as 1440K?"; you click window B, whose paint clears the global; you click back into A to dismiss the question, and `fm_status_only` no longer refuses — so it repaints only the status line and the question stays in A's row area, over the file names, for the rest of the session. That's `docs/FIELD-NOTES.md` 17.1's defect in the one path this hunk was added to close.

It moves to `FS_2UP = 14` — one of the two bytes the `FS_` block already documents as a free hole — with an `fm_l2up` mirror beside `fm_ledit`, which is the module's own idiom. `fm_kinit` clears it on slot reuse. `FS_SIZE` unchanged, and the bytes land in `.cold`, so it costs nothing measurable. I drove the exact scenario to confirm it.

**12. The two CI workflows are removed.** `3618edd`

**This one is the repository owner's decision rather than a defect**, and I want to be straight about what was and wasn't wrong with them. Both are `ubuntu-latest`, both use only first-party pinned `actions/*`, and neither publishes anything — two of the three objections that retired the old proxy workflows don't reach yours. What does reach them is the third: neither declares a `permissions:` block, so both inherit the default `GITHUB_TOKEN` scope, and `tests.yml` builds two pieces of fetched third-party source in that environment. `main` carries no workflows at all, so removing these returns `.github/workflows/` to exactly `main`'s state.

**The suite itself is untouched and still runs** — `make` runs the fast tier, `os88test.py full` is the pre-merge one. Only the automatic run on push goes away.

Two references followed them out. `tools/martypc/build.sh`'s "third state" comment explained the cache-hit case by naming `tests.yml`; the behaviour is a property of any restoring cache, so the comment says that and the code is unchanged. And `docs/PROXY-PLAN.md` §13.2 needed correcting either way — this branch rewrote it to describe `os88proxy-exe.yml` as a live on-demand workflow ("Actions ▸ os88proxy.exe ▸ Run workflow"), and that file exists in neither tree; it was deleted upstream before this branch merged. It now records the decision as a decision and keeps the reasoning you added about the filter that wasn't narrow.

**13. Both kernels re-blessed.** `57d6381` — including the `-4` cold delta the branch was already printing before any of this.

---

## Flagged, not fixed

**Under Dark on a 1bpp adapter, a disabled chrome glyph draws solid** — pixel-identical to a live one. `font_ink`'s own comment says exactly why `[gfx_disink]` exists ("under the Dark theme the chrome's ink IS white, so a greyed caption came out SOLID"), but `font_char_bb` branches `test ch,1 / jnz .srow` *before* it consults `[font_dith]`, on the stated assumption that the dither is "always on the clear-bits side" — which the new white disabled ink makes false. So §76.6.1's fix is in the pen and not in the renderer. Content greying is unaffected (`gfx_pen_dis` still stores `CBLACK`).

The fix wants a dithered set-bits loop mirroring `.dstart`/`.drow`, about 40 bytes — and `kern_big` has **6 bytes of image-rung slack**, so it crosses a rung and spends one of the three remaining budget steps. That's a byte-spend decision, and the owner asked for it to come to you rather than be absorbed quietly. §76.9's whole-screen lit-pixel count structurally cannot see this, so a `tests/dispthm.py` row asserting the greyed "Close Window" differs pixel-wise from the live one on a 1bpp adapter would be worth having with it.

**`thm_set` overwrites a user's Color choice rather than refusing it.** It doesn't just decline Color on a non-VGA machine — it writes the fallback back into `[thm_kind]`, which is the byte serialised to `SYSTEM.CFG`. The two new re-assert sites (`vidsel.inc` on every adapter change, and `drv_cfg_unpack`, which `drv_boot` runs *before* it applies `CFG_VID`) turn a one-way refusal into repeatable destruction: pick Color on a VGA, visit Control Panel ▸ Display, try Hercules and come back, and Color is gone with nothing on screen having said so. I left this alone because the clean repair splits "what the file asks for" from "what the machine can do" — a `[thm_want]` byte beside `[thm_kind]`, serialised in the `TH` key — and that's a second change to an on-disk format this PR is already changing. Your call how to shape it. The naive fix (just don't write back) makes three readers start lying, including `cp_thm_radios`, which would fill a radio §76.12.1 has greyed.

**The alert still repaints all three buttons on a real hit** — only one changed. Fixing it properly wants a one-button painter in `fdlg_btn1`'s shape, which is new code in an SDK header you're already introducing; how much of `fdlg`'s shape `os88ui` should grow is yours.

Smaller things, all real but none urgent:

- **The animation is ~164 ms, not §11.99.4's 137.** `wm_an_deadline` is sampled at the top of the frame so the *draw* is absorbed by the wait, but the frame's **erase** runs after `wm_an_pace` and is added to it.
- **§76.12.3 contradicts §76.13** about what `kern_small` pays for the palette; the build agrees with §76.13.
- **§76.5 still specifies the `TH` key as one byte** holding `[thm_dark]`; the writer emits a ver-2, len-2 record covering `[thm_kind]` and `[wm_animon]`.
- **`ANIMOFF` hides the two Window Animations rows** where §76.11 requires them greyed ("grey a fact" — §47 rule 3, and the section says so explicitly).
- **The SDK note on `OSAPI_WM_PREFER` inverts the `WF_SIZABLE` rule**: it says a drag applies the preference "unless your window is `WF_SIZABLE`", but `wm_apply_size_wh` takes it for every window and only consults `WF_SIZABLE` when there is no preference.
- **`np_onclose`'s header promises a save-and-close fallback** the code doesn't do and §27.15.3 deleted.
- **The alert's minimize box doesn't produce `OS88UI_ACANCEL`**, contrary to the completion contract in the same file's header.

And on the suite, which I like a lot — these are about gates that can't fail rather than gates that do:

- **`tests/dispprefer.py` reports five legs "skipped" if the Disk window regresses.** `find_app` returns "not on this disk" for any reason the row isn't in the listing, but `BROWSER`/`PAINT`/`TEXPAD` are all on `apps360.img` by construction, so that message can only ever be false. Break the re-list-on-focus or the row hit-test and legs A–E all print "skipped" and `main` returns 0. `tests/dispsize.py:401` has the same shape for a drag regression that is itself the subject.
- **`bootsmoke`'s dock assertion can't fail.** `check(dock > 0, ...)` is satisfied by the desktop's 50% dither behind it — delete the dock paint entirely and the row still passes. Asserting the strip's *ground* (solid, either ink) distinguishes it.
- **`t_registry` walks only `tests/`, not `tests/unit/`** — so a new `t_*.py` nobody registers is invisible, which is the exact rot the row exists to stop, one directory down.
- **`t_buildmatrix`'s restore runs the default goal**, which now depends on `test-fast`, so a fast-tier failure is reported as "the matrix left build/ dirty".
- **And a gap I hit myself**: `os88test`'s preflight only fires for rows needing `marty`/`qemu`, and `buildmatrix` needs neither — so a `build/kernel.bin` left over from before a commit reaches it as `FAIL: the shipped kernel was not clobbered`, pointing at the matrix rather than at the commit. It cost me one run before I recognised it as the thing `075e85f` and `44e3468` are about. Widening the preflight to any row that reads `build/kernel.bin` would close it.
- **`dispcorner`'s residue classifier** is now parity-only with an area floor, so a 1px strip is legal where the old rule needed 3×3. I could not construct a reachable defect it excuses and the loosening fixes a real hole (a track *with a thumb in it*), so I left it — but the two new selftest cases are both wide and solid, so the newly-legal thin class has no case behind it.

---

## What I checked and cleared

So nobody redoes it: **API slots** — 140 cells, no duplicate address, all 135 of `main`'s keep `main`'s name, and the SDK diff is purely additive with zero removals, so no existing contract moved. I also built `main`'s packages in a separate worktree and ran them on this kernel: `main`'s Solitaire comes up correctly sized and in colour, and `main`'s Note Pad — which installs no negotiator — opens, takes text and closes silently with no alert and no crash. **Guard nesting** — zero of 1,787 common labels across 19 kernel/boot/driver files changed their `%ifdef`, so `84897d8`'s Control Panel release-path fix is intact. **`SYSTEM.CFG`** — the `CFG_*` renumber is clean; every `CFG_*` is a RAM struct offset, records are matched by name, no ordinal reaches disk. **Preference tables** — every `OS88_PREFER` site is at file scope in the image, none on a stack. **Sizes** — no rung crossed by any of this; `kern_big` 1,536 spare, `kern_small` 512, both unmoved.

**The budget raises are accepted as you argued them** — the union of two long branches crossing a guard neither crossed alone.

**Verification:** `make`, `make KERN_SMALL=1`, `make THEMEDARK=1`, `make FSNOSTAMP=1` all clean; the full tier **11 passed, 0 failed** (`ctoolchain` and `bootsmoke` skip here — no SmallerC, and no Rust toolchain to build MartyPC); `checkdocs` 1150/140/0; all seven images fsck clean; desktops on VGA and CGA.

One thing I could **not** run: MartyPC won't build on this machine, so `dispclose`, `dispprefer`, `dispsize`, `dispthm`, `dockmark` and `bootsmoke` — your own gates for this work — never executed. Everything above was driven by hand over QMP instead. If you can run that tier on your side before merge, it's the half of the verification I'm missing.

Nice piece of work, and the SPEC sections carry their own reasoning better than most of what's already in the tree.

PR #112, opened 2026-08-23
Elendilon -> Main (FTP Server, Screen Saver, Disk Clone, 2.5s Off Boot on XT class, Draggable Scrollbars, Ether Speed, TexPad Optimization)

Elendilon:
* TexPad optimization pass by pagetelegram, from the issues board
* Second TexPad optimization pass, done after importing pagetelegram's changes
* FTP Server Package
* Ether.drv speed optimization. 4KB/s -> 30KB/s
-- When paired with writes to a ST-225 in ftpd, I didn't get a clean "read to memory" delta
* Stack overflow "fix". 12 worker tasks with 256ea -> 8 worker tasks with 384ea
-- This is worker tasks not windows - things that run "in the background simultaneously"
-- Used by windows that animate even when not in focus, drivers that move data in the background (sound only, atm), etc.
-- Max windows is unchanged, but this means a max of 7 'bounce' instead of a max of '10'
-- Cost to increase it back is ram. I can overrun this on purpose if I just spam windows open, but not during "normal" use. Still worth considering a "kern_huge" for 386's or demoing 10 bounces maybe.
* PS/2 mouse support
-- Combined with main's implementation, picking mostly main and adding support for the keyboard-mouse to main's.
* Optimized ability to draw a line. 5-6x speed improvement.
-- Just a line. 700b, just to draw better lines.
* Wireframe demo - not in the shipping disk but there is a MAKE target for it
-- Served as a testbed for screensaver
* Animated screensaver, with configuration
* Clone disk tool
-- disk->disk, file->disk, disk->file
* Boot time disk read optimizations - boot 2.5s faster
-- On an emu that models disk speed at 5150/tandon rates (or a real machine)
-- EDIT: Added a canary/retry fallback and disabled this for 286+ after finding a bios (mr286 in 86Box) that does not read its own disk state memory, thus ignoring the head change and corrupting our kernel read.
-- Every XT bios on minuszerodegrees DOES support this, and the canary fallback will catch it if someone finds one that doesn't (at the cost of ~10s added to each boot, instead of 2s less).
* Draggable scrollbars
* Driver memory usage noted in control panel
-- This is their MAX usage, not including purgable caches
-- Ether will gracefully degrade from a 37kb buffer to an 8kb one, based on free ram
-- Hard drive has a 6kb buffer per mounted volume that drives up this max, but can degrade that on low ram systems to no buffers at an extreme speed penalty
-- Sound can degrade its on demand buffer by 6kb on low ram systems, at a cost of skips when first starting playing
-- Ram disk is whatever you configure it to be, plus 2k for the driver itself
-- os88net is static 6kb, no buffers
* Full fable review pass, finding ~50 issues across the codebase and fixing them.
* Various tests, benchmarks added during development


jggonz:
Reviewed this one end to end and pushed **nine commits** to `elendilon-pr` (`ada83cc..1e7d0ab`). No merge was needed — you had already rebased onto today's `main` and reconciled the PS/2 collision yourself, and `git merge main` says `Already up to date`. The PR now reads **MERGEABLE**, blocked only on approval.

`make test-full` is green again: **16 passed, 0 failed, 1 skipped** (only `bootsmoke`, which wants MartyPC). It was failing when I picked this up — more on that first, because it is the one that mattered most.

Nothing of yours was rewritten, squashed or reverted. Every commit below is additive on top of your two.

---

## The three that stopped the tree building

**`kern_small` was 512 bytes over its own budget, so `make small` did not assemble at all.** Nobody could build the 128KB-machine floppies, and `test-full`'s `buildmatrix` row failed with it. This is worth knowing exactly *why* it slipped past you: at your first commit (`b39ecf5`) `kern_small` measured **106,496 of 106,496 — zero spare, one byte left inside the image rung.** Your `test-full 16/16` was true there. The boot canary and XT gate in the second commit then added six bytes, which crossed a whole 512-byte rung and put it 512 over. The budget move goes to **107,008**, recorded as the thirty-second move in `docs/KERNEL-MEMORY.md` with the attribution — that was the repository owner's call, taken explicitly rather than trimmed out of your features. `72135bf`

The same commit re-ran `kernsize --bless` for **both** variants. The baseline in `docs/KERNEL-MEMORY.md` was describing a kernel the tree no longer builds, so every `make` printed `*** the image rung CROSSED - the machine's RAM moved ***` against a stale reference — which trains a reader to ignore the one alarm that document exists to raise.

**The boot sector was at 510 of 510 bytes, and four documented knob builds no longer assembled.** `make FLOPPY1=1`, `make DISKAL=1`, `make BOOTDIAG=1` and `make BOOTSTOP=1` all died with `boot/boot.asm:639: error: TIMES value -N is negative` (−3, −13, −29 and −4). `main` builds all of them. So nobody could produce **`cqdiag.img`** — the diagnostic floppy for a machine that will not start, which is precisely the machine your whole boot commit is about — or either A/B disk the field notes are taken with. `BOOTSTOP=1` is a knob this PR itself adds, and the Makefile comment beside it says "four bytes — which is exactly what the sector has spare"; it was four bytes over. The headroom is reclaimed without deleting any knob, and `t_buildmatrix` grew rows that ask for `boot360.bin` rather than `kernel.bin`, so this class cannot come back unseen. All seven configurations now produce a 512-byte sector. `8152e0e`

**FTPD's `fd_cdirty` was laid out on top of the "Serve every drive" hit rect.** `fd_rchk equ fd_srects + 8` and `fd_mchk equ fd_srects + 16`, and then `fd_cdirty equ fd_rchk + 8` — which is `fd_srects + 16`, the same address. So `fd_cdirty` and the four scalars chained after it (`fd_lognew`, `fd_scrn`, `fd_scry`, `fd_xbytes`) sat on the rect. Drawing the Setup page wrote screen coordinates over the dirty flag and over the transfer byte total; every hit test on that page strode through them as its third rect. In use that meant **every visit to Setup did a full floppy mount, write and remount on close** — the exact cost the "mark on click, write on close" design exists to avoid. Moved to `fd_srects + 24` with a `%error` guard keyed off `FD_NR`, so a fourth Setup control trips it rather than aliasing. `6bd4992`

## The correctness ones

**A controller that *refuses* a cross-head read halts a boot that `main` completed.** `read_run` now asks for a whole cylinder on exactly the target machine, retries three times at the same width, and gives up with `DSK`. The canary can't help here — it only runs after every call returned CF=0, and this failure mode is CF=1. Your own `dsk_xfer` already has the ladder for it (`kernel/disk.inc:1294`, written for a 360KB clone that answered status 0x80 to a crossing write); the loader now has the same one-shot fallback. `8152e0e`

**`fd_up` walked up with the drive banked at the start of the CWD.** In whole-machine mode, `CWD /B/DOCS/..` banks A:, enters volume B mid-walk, then pops B's cluster and calls `GOTO_QM` with BL still A:. `PWD` answered `/B` while the session stood on an arbitrary cluster of a *different disk* — and every later `STOR`, `DELE` and `MKD` resolved there. `FILE_HERE` does no disk I/O, so asking where we actually are costs one far call. `6bd4992`

**TeXPad's drag-select froze after the first movement.** `tp_sel_hide` had no `push`/`pop` and returned `AX`/`DX` holding the *drawn* span; `tp_ondrag` loads the anchor and cursor into exactly those two registers, calls it, then calls `tp_sel_set`. So from the second mouse report onward the selection was set back to itself and the highlight stopped following the pointer. The same clobber made a scroll-bar click blit the pane by `dsel0 * 8` pixels with a selection live. Fixed inside `tp_sel_hide` — it now matches `tp_caret_hide`, `tp_sel_show` and `tp_xor_span`, all of which already preserve everything. `d7588e1`

**The cloner could be unloaded out from under an armed clone**, and **`CLA_SAVE` and `CERR_NODISK` were both `4`** with the one caller that must tell them apart never testing CF — so a failed reload mid-clone opened a "Save file as…" box, on a half-written floppy, for an operation nobody asked for. The `CLA_*` enumeration moved out of the `CERR_*` range so the next caller cannot make the same mistake. Cancelling with Esc also skipped the re-mount, leaving the machine believing the disk in the drive had the abandoned clone's geometry. `6a91253`

**Deleting or renaming on a write-protected floppy told the client the file did not exist.** `DELE`, `RNTO` and `RMD` threw away the kernel's `FERR` and answered a flat 550. Note the trap here, in case you touch it: `fd_ferr_sel` had no `FERR_NOENT` arm, so routing straight through it would have made the *common* case worse — a genuinely missing file would have come back as "disk full or protected". The arms went in first. `6bd4992`

**The FTP server said "Listening" with no worker behind it** if the task table was full when Start was pressed. Every client was ignored with no error and port 21 stayed claimed. That matters more now that this PR takes `MAX_TASKS` from 12 to 8. `6bd4992`

## The constants this PR moved, in the files that still mirrored the old ones

`MAX_TASKS` 12→8 and `SCH_STACK` 256→384 are good changes, but three places still carried the old numbers. **`apps/cc/crt0.asm`'s `CC_STACK` was still 256**, and it is the window that decides "am I a worker?" — too *short* a window fails the gate **open**, so a C program's background task more than 256 bytes deep was read as the main task and would have claimed memory and read the floppy from the wrong one. SPEC.md §2.1 and §8 still pinned 12 × 256 and still taught the SP-low-bits recipe your `% 128` change retired. And the new `OSAPI_FILE_RMDIR` comment promised `FERR_EXIST` for a non-empty folder where the kernel returns `FERR_PROT`. `122095b`

## Cost, and two contracts that had outrun their code

The Screen Saver settings window repainted its **whole content** on a focus change — about a third of a second of visibly redrawing window on a real XT to move a one-pixel caret. It now redraws the two fields (~40 primitive calls down to ~8). The file dialog repainted its whole list on a thumb release that scrolled nothing — the no-op guard `fm_scroll_by`, Notepad, Browser, Word and Frotz all already have. `2380e07`

SPEC.md §13.10.5.3 described a different thumb-to-position map from the one that ships, and did so precisely in the case your change exists to fix. §5.6.4.3 said the left edge was closed; it is closed until the walk crosses the last bank, so it now says what the code does and where the residue is (392 of 29,520 modelled cases on Hercules, down from 28,800 before your `sar` — the change is a large improvement, the sentence was just ahead of it). `17a4921`, `2380e07`

Also: a ring-less `ETHER.DRV` accepted a send it never copied and would have put the *previous* outgoing frame on the wire — correctly checksummed, with no error anywhere. And WIREFRAME's Composed mode erased a fallback frame with kernel lines that light different pixels than the mask did, so a refused band left a ghost cube accumulating on the glass. `17a4921`

## What I did not fix

- **`make field` is still broken — but it was broken on `main` before this PR.** The 360KB bench disk overflows: `main` needs 356 clusters for a 354-cluster disk, and this PR takes that to 373 because `saver.drv` and `clone.drv` joined those disks. The boot-sector half is fixed; which packages ride that floppy is a call for @jggonz, not a review fix.
- **The cloner's source-prompt identity check.** The check guards the TARGET prompt only, so on a multi-trip one-drive clone a user who forgets to swap at the SOURCE prompt gets a chunk read off the half-written target and written back over itself — and it finishes reporting success. The fix needs a new `CLS_*` warning step and a SPEC amendment, which is a feature decision rather than a repair.
- **Your design decisions, all of them.** `MAX_TASKS` 12→8, replacing `main`'s `wm_wake_sweep` idle backstop with per-drain `wm_wake_eaten` accounting, the withdrawn XOR overlay, deterministic `~1` name mangling, `mou_ptr` never being cleared. On the wake change specifically I checked whether anything `main`'s backstop caught was lost: all seven ring drains call the replacement and `t_wakedrain` enforces it, so I read it as a straight improvement.
- **`CLAUDE.md`'s instrument policy is reverted to `main`'s** — the dependency line, the `make marty` command-block entry, and the MartyPC-first Testing section. That is a maintainer decision, not a defect: I put it to @jggonz that `main`'s own `docs/TESTING.md` already opens MartyPC-first and that 63 of its 76 suite rows are already marty-gated, so your edit was arguably bringing a stale `CLAUDE.md` into line — and he asked for the revert anyway. **Every additive change you made to that file is untouched**: the five new document rows, `make netbench`, `make wiredisk`, all seven new knobs, the three new hard rules, the `pkill` trap and the Layout additions. One sentence now points at `docs/TESTING.md` as the authority and warns that it currently disagrees. `1e7d0ab`

## What I checked and it was fine

- **The path layer, hard.** FTP is the one surface here where every byte comes off the wire, and PR #92 shipped a `..\` traversal in this repository before. I drove a real `ftplib` client at the guest with the root set to `DEEP` and tried `RETR ../`, `SIZE ../`, `LIST ..`, `RETR ../../` and `RETR ..\` — **all refused**, because `dsk_find_x`'s species filter drops on-disk dot links so `..` can never match a component. That was my leading blocker hypothesis and the measurement killed it. Re-ran it after the `fd_up` fix: still confined.
- **The canary offset.** Re-derived `KSIG_OFF = 18432` (file sector 36) from each shipped image's own BPB: it lands in a run's *second* half in all three geometries, and `t_canary.py` genuinely re-derives rather than asserting a constant. That is the trap your own commit message describes falling into once, and the second version does not fall into it.
- **The API slot table.** 140 → 143 slots, zero renumbering, three appended at fresh addresses. Exactly the discipline `docs/UPSTREAM.md` asks for.
- **`ftpd.py`, all twelve assertions**, before and after the fixes — host fsck clean, the uploaded file byte-exact off the image by an independent reader.

## Verification, in full

`make` clean from scratch · `make small` builds both floppies · `kernsize` for both variants inside budget with no unblessed rung crossing · `os88disk --verify` on all nine images · boots to a desktop on VGA **and** CGA, pixel-identical to `main`'s · all seven boot-sector knob configurations at 512 bytes · `make test-full` **16 passed, 0 failed** (I built the pinned C toolchain so `ctoolchain` ran rather than skipped, since the C runtime was one of the things changed) · `make cword` builds · `tests/ftpd.py` all assertions.

**Five of your new gates I could not run**: `saver`, `diskclone`, `sbar`, `linefast` and `evqfull` are all marty-gated and there is no Rust on this machine, so they SKIP. They are not evidence of anything either way — worth a run on a box with `cargo` before this merges.

Genuinely good work in here — the FTP server's stage-and-commit shape, the canary re-derivation, the octant line walk (I had it modelled over ~500k endpoint/clip combinations and it does not put a pixel outside the box), and the honesty of the boot commit's own write-up all made this a much easier review than its size suggests.

PR #115, opened 2026-08-24
Live media: bootable USB image and live CD (SPEC.md 80)

jggonz:
## What

Two new on-demand release artifacts carrying the whole OS and every application, for machines with no floppy drive:

- **`make usb`** -> `build/os8088-usb.img` (32MB raw): a bootable hard-disk image, written raw to a USB stick and booted in legacy BIOS mode.
- **`make iso`** -> `build/os8088.iso`: the **same image** wrapped in an El Torito hard-disk-emulation live CD. `make live` builds both.

The release zip now carries both when they were built, with a "LIVE USB AND LIVE CD" section in its README (the `dd` and QEMU commands), and the release skill builds and boot-tests them.

## How

No new boot code. Both media reuse the field-verified SPEC.md 52.10 hard-disk boot chain: `boot/mbr.asm` -> `boot/boothd.asm` as the VBR -> one FAT16 partition with `KERNEL.SYS` first and contiguous, which the kernel adopts as C: before any driver loads (52.10.3). A machine booted from the stick is an *installed* machine as far as every line of the OS is concerned.

- `tools/os88disk.py --hdd` builds the image, sharing the floppy builder's whole middle (folder tree, attribute rules, warm ASSOC.DAT, `--deep-folders`/`--dir-slots`) rather than duplicating it, and the recipe checks the result with the existing `--verify-hdd`. Geometry is fixed at 16 heads x 63 spt x 65 cylinders - 65,457 partition sectors, just under the kernel's 65,535-sector volume ceiling - with CHS/LBA-consistent partition entries, because a BIOS booting a USB stick or an emulated CD derives its virtual geometry from that table.
- `tools/os88iso.py` (new, ~250 lines) writes a deterministic ISO9660 volume + El Torito boot catalog, hard-disk emulation only; it refuses a boot image without an MBR and an active type-04h entry. The raw image and README ride the ISO as plain files, so a host that mounts the CD can copy the stick image off it.
- Payload is derived, not listed: the system disk's arguments plus `ALLAPPSARGS`, with the RunCPM drive-A selection taken exactly as `allapps` takes it, so the two everything-payloads cannot drift. On demand for allapps' reason (needs the C toolchain and `make runcpm-src`).
- SPEC.md 80 (written before the code) is the contract, including what a CD honestly cannot do: settings and saves last until power-off there (80.3).

## Verification

- Booted **both** wrappers in QEMU (`-boot c` and `-cdrom ... -boot d`): desktop up, C: icon present, Disk window lists all 11 root entries with 31MB free, folder navigation works, Minesweeper launches off C:'s GAMES folder with its cached icon.
- `--verify-hdd` passes on the built image (147 files, FATs agree, no loops/cross-links).
- macOS `hdiutil attach` mounts the ISO; the embedded `OS8088HD.IMG` is byte-identical to the USB image.
- Both images rebuild **byte-for-byte** (delete + rebuild, sha256 match).
- `make` and `make test-fast` green (14/14 rows, doc gate included).
- Test zip packed and round-tripped: SHA256SUMS verify, README renders the live section only when the images are present.

https://claude.ai/code/session_0174maNLRwuJK5QUZhgw6FaT

### DIFF
--- a/.claude/skills/release-os8088/SKILL.md
+++ b/.claude/skills/release-os8088/SKILL.md
@@ -171,6 +171,10 @@ tools/setup-cc.sh                     # SmallerC; needed by cword and allapps
 make allapps                          # apps-all.img -- every program, one disk
 make worddisk cworddisk               # word*.img, cword*.img
 make runcpm-src && make runcpmdisk    # runcpm*.img
+make live                             # os8088-usb.img + os8088.iso -- the live
+                                      # USB image and the live CD (SPEC.md 80).
+                                      # Needs the fetch on the line above and
+                                      # the C toolchain, like allapps
 ```
 
 Offer these, do not assume them. If `tools/setup-cc.sh` cannot run -- no
@@ -204,6 +208,25 @@ menu bar across the top with the chip glyph, then Locator, File and Builtins;
 a Disk A and a Disk B icon down the right-hand side; the mouse pointer.
 If the screen is blank or garbled, stop -- do not publish. Delete the two
 scratch files afterwards; `build/` is gitignored, but leave it tidy.
+
+**If `make live` ran, boot both live images the same way** -- the rule is
+per image, and a zip does not exempt the two biggest files in it:
+
+```bash
+qemu-system-i386 -display none -qmp unix:build/qmp.sock,server,nowait \
+  -drive file=build/os8088-usb.img,format=raw -boot c \
+  -chardev msmouse,id=m0 -serial chardev:m0 &
+sleep 10
+python3 tools/qmp.py build/qmp.sock 'screendump build/smoke-usb.ppm'
+python3 tools/qmp.py build/qmp.sock 'quit'
+# ...then the same five lines with `-cdrom build/os8088.iso -boot d`
+```
+
+**Look at both screenshots.** The desktop must show a Disk A icon AND a
+Disk C icon down the right-hand side -- C: is the live partition the kernel
+adopted, and its absence means the image booted the kernel and lost the
+volume, which is exactly the failure a screenshot catches and a checksum
+cannot.
 
 ### 3a. Pack the release zip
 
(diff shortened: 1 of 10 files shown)


jggonz:
Added a second commit (39aba3f): **`make burn`** -- `tools/os88burn.py`, a macOS interactive text-mode guide for putting the live media on real media (SPEC.md 80.4).

- **USB half**: lists only disks that could be the right answer (USB bus + external + never the disk macOS is running from -- filtered out, not warned about), walks through the choice, requires typing the disk's identifier to confirm the erase, unmounts, writes the raw device with a progress line, then **reads the whole image back and compares SHA-256s** (the check that catches fake-capacity flash). The wizard stays unprivileged; only the write step re-invokes itself under sudo.
- **CD half**: `drutil list` detects a burner (none attached: the menu row says so and does nothing), `hdiutil burn` burns and verifies, with a retry loop for a missing blank disc.
- Refuses a wrong image up front: the USB path demands the MBR + active type-04h partition `os88disk.py --hdd` writes, the CD path demands ISO9660 -- so a floppy image dragged in by mistake is an error message, not a dead stick.
- `--scan` lists drives and burners non-interactively; the zip README now points Mac users at `make burn`.

Tested on a Mac with no stick and no burner attached: scan, both empty-hardware paths, the write+verify halves against a scratch target (byte-identical result), and both validators' refusals. The destructive path itself is exactly what the typed-identifier confirmation gates.

jggonz:
Third commit: **`docs/LIVE-MEDIA.md`**, the reader's guide to the live media, plus a badged section near the top of the README linking it.

- The guide covers the whole journey for someone who has never seen the project: getting the images (release zip or `make live`), writing the stick (`make burn` with a sample session; `dd` by hand on macOS/Linux; Rufus/Etcher in raw mode on Windows), burning the CD, booting on real hardware (legacy BIOS/CSM, USB-HDD emulation) and in QEMU, what the desktop should show, and a troubleshooting ladder that starts at the checksum.
- README additions: a colorful shields.io-badged section with a GitHub `[!TIP]` admonition right after the project intro, `make live` / `make burn` in the target list, and the compiler note now names the live media beside `allapps`.
- CLAUDE.md's document table gets a row for the new doc, with the rule that SPEC.md 80 stays the design record and the guide follows it, never leads.

`make test-fast` still 14/14 (doc gate included).

PR #116, opened 2026-08-26
Elendilon -> Main (VGA Optimization Pass, Text Flicker Pass 2, UI Loop Optimization, Bugfixing)

Elendilon:
* Major VGA optimization pass 1 - 4.5KB kernel growth
* Flickering char drawing pass 2
-- These two turned out to be effectively the same thing, so they ran together
-- VGA get's mono's previous char draw gains, roughly 3x faster than where it sat before, for most scenerios
-- VGA gets the equivalent to mono's blit fast paths. 2-7x where it sat before, depending on content
-- Paint on a VGA 4.77Mhz draws os8088.gif in 170ms, vs 7 seconds before.
---- Note again this is drawing to the screen, not decoding. It still takes a 4.77Mhz CPU 20 seconds to decode os8088.gif from its gif format to something usable
-- Tracker 11Hz on a 286 12Mhz VGA keeps up fullscreen on Elysium.mod now
-- A future step is trying to evict most of this new vga-only ram usage from non-vga systems
* 30% of each work pass was checking for work that hardly ever needs done. Decreased to ~2%.
-- 10% was doing a 16 bit math operation to see if the corner clock needed redrawn
-- 3% checking to see if we needed to beep from holding down a key
-- 3% from checking to see if a file dialog had closed (removed completely)
-- etc, etc in 3%ish chunks
-- This is less bad than it sounds, because when the system is busy it does less work passes, and thus these checks account for less total time. At idle it really was 30% of the cpu's time.
* FTPD, Arkanoid, Cyclone bugfixes/deflashing work
* Minor dead code cleanup

* Kernel size optimization and restructure
-- 5kb smaller. Easy to spend bytes, hard to buy them back
-- A near equal trade with the previous vga/flashing work's increase
-- But the pressure is also off .image
* Ether.drv has 8 sockets instead of 4, and needs 23kb less of buffers
-- Only one socket is a data socket, which is all any current app needs
* Scrollable taskman heap list
* Idle task is accounted, taskman usage percentages are truthful now
* Taskman optimized, doesn't lock the whole system for 100ms chunks
* Boot splash relocated to a droppable region
* Heap compression pass 2 and boot structuring
-- Fully compacted heap at boot time, more re-compactable after use
* Driver init phase added to boot splash
* Bugfixes
-----------------------------------------------
64 commits of work from the integration branch, as one commit. Nothing here is new: this is elendilon's content three-way merged onto main, so main's own Live media work (SPEC.md 80, #115) is preserved rather than overwritten.

CLAUDE.md is main's, wholesale. The integration branch carried a branch-specific "Working in this fork" section that is the fork owner's standing preference rather than a property of the project, and it is not this PR's to bring along; taking main's file drops it, along with the pointer to it.

Graphics and text
- gfx_blit4 stops working per run on VGA: a planar row decoder prices a dithered blit per PIXEL instead of per run (SPEC.md 5.4.1.3). NOPLANE=1 is the A/B and keeps the run-only path assembling.
- The composed title bar (band.inc) becomes kern_big's default and asks the focus question itself; NOBAND=1 keeps the fifteen-call path alive.
- Paint's canvas is four planes and a repaint is a copy (SPEC.md 42.13); gfx_blitp takes a block that is already framebuffer bytes (5.4.3). On a 1bpp adapter Paint banks its whole content, not a band (11.96.11.3).
- The clipboard holds what the canvas holds, and neither clipboard loop tests per byte what it can decide once (42.13.3).
- Text is font_run: docs/TEXT-PLAN.md stage 4 is finished, and the registry's remaining sites each carry a reason rather than a queue position.
- A display swap turns the row table off for the display it was not built for (SPEC.md 39.3.1).

Window manager and UI
- wm_show, wm_hide and wm_front take the gfx lock (SPEC.md 11.101.2).
- The progress widget and the mouse ISR were two painters (12.8.4).
- os88ui_glyph's fallback could not have drawn a glyph, so it goes.

Disks and drivers
- The hard disk mounts when its driver comes up; the install writes where the loader reads, so the BPB is the one geometry, and the boot sector keeps hd_chs's guards.
- A volume this driver cannot reach is not an empty slot.
- osapi_drv_dlg_x returns FAR.
- DEBUG.DRV, the serial monitor, is removed (SPEC.md 58).

Applications
- Minesweeper: the wrong-flag X was black on a black mine (SPEC.md 23).
- Arkanoid: the falling capsule froze because DX and SI came back swapped (44.10.6.1); one band flag was answering two questions.
- Cyclone: holding the mouse button repeats the gun (67.11.3).
- Tracker: the pattern view scrolls n rows, not one (45.12.2).
- ftpd: the Setup page repainted itself for a CARET (77.45).

Tests, gates and docs
- asmrules saw a `clc` before `ret` and skipped 719 sites; textrules could not see UI_STR, the SDK's own macro; the gate can now see a label defined twice.
- dispcheck could not see a secondary drawn with the wrong scan-line bases; kernsize could not see a budget move.
- NOPLANE and GFXAUDIT join the build matrix; ftpdflick can now answer NO.
- A knob kernel is bound by the segment, not by KERN_BUDGET, so two diagnostics stop spending the one guard nobody can raise.
- kern_big's budget takes a step back: 121,344 -> 122,368 (move 35).

Boot — the loader leaves the boot sector (SPEC.md 2.9)
- 512 bytes stopped being enough, so stage 2 is a blob of its own that
  stage 1 reads and enters (2.9). It comes down to the heap's floor rather
  than sitting in the FAT window (2.9.5, 50.3.3), and the boot overlay
  joins it there (2.5.1, 2.9.6).
- Stage 1 checks the blob it read, because nothing else does (2.9.7);
  BOOTDIAG=1 makes that checksum say WHAT it got.
- The diskette parameter table is stage 1's now (2.9.8), and the hard disk
  has its OWN boot sector, which never heard about the blob (2.9.9).
- .text's eleven guarded calls share one gate (2.9.5.2).
- The loading screen leaves .text (2.9.4) and says WHAT it is doing, not
  only how far (15.6). The run bound comes out of the STASH, not out of CX
  (18.93.3).

Scheduler — an idle desktop was 97.7% poll loop (SPEC.md 8.1)
- The accounting charged a slice on every switch and nobody read it. It
  charges only when the task CHANGES (8.1.1): 10.94% -> 0.18%, +10 bytes,
  exact. Measuring less often measures 21 points WRONG, because os8088's
  tasks are metronomes and the tick aliases against them.
- ui_task BLOCKS (8.1.2): an idle desktop is 96.9% halted at 17.7 passes a
  second instead of 1,134. The pointer was the veto on it and it did not
  fire — the mouse ISR draws the arrow itself. NOUIBLOCK=1 is the A/B and
  is what keeps the spinning path assembling.
- The Task Manager's load meter reads the IDLE TASK rather than a spin
  counter (28.7), so "CPU 1%" beside "TaskMgr 97%" cannot happen again.
- docs/SCHED-IDLE-PLAN.md is the design record and the handoff.

Memory
- mem_avail answers what mem_claim would GIVE, not what is free (50.6.3,
  66.10.3), and a cache is not a barrier (50.6.1, 66.10).
- mem_claim_dma skipped the direction test (50.6.1.1).
- A driver's ring belongs at the top of the heap (50.3.2, 66.5.6.1).
- apps.inc goes cold; 43 unpinned far entries and eleven nothing called
  come out (2.6, 2.6.1). Set 95 is withdrawn — price the segment machinery
  instead.
- .vgabuf: the planar decoder's buffers leave .lowbss, and a mono machine
  never claims them.
- tools/heapmap.py reads the claim heap instead of reasoning about it, and
  reports what a claimant can HAVE rather than what is free (50.6.2).

Settings and the clock
- The settings parser is not resident any more: it moves into the boot
  overlay with a second copy in CTRL.DRV, so a setting costs 2 resident
  bytes instead of 14 (51.5.3). docs/SETTINGS-COST.md carries both designs
  and why the free one loses.
- The Date/Time field editor ships in CTRL.DRV (37.93), and so does the
  RTC write half, spent by the panel's CLOSE like every other setting
  (37.94).
- The ladder traces itself, because only the field can run it (37.95).

Drawing
- The title-bar composer is a knob again, not kern_big's default (5.9.6):
  price it in bytes, not in rungs.
- Opening a document redrew the whole Disk window (22.13.2).
- Tracker's fullscreen text screen is a PICK, not XT mode's (45.13.7).
- The Task Manager's quiet pages ask the cheap question first, with no
  lock at all; the heap page and the memory view paint every fourth
  interval, and the heap page scrolls — its scroll range had been a
  function of the scroll position.

Network and installer
- ether/ftpd: a slot and a ring stop being the same thing, plus MSS, ACKs,
  RST, LIST -a and a socket view.
- The installer's action button becomes Restart (52.10.6.1).

Rules, tests and docs
- Design for BYTES, never for rungs. A change that crosses no rung has not
  cost nothing — it spent slack belonging to whoever comes next, and the
  guard bills them the whole 512 instead.
- Two dead gates: xmcheck could not be run at all, and blitplane measured
  nothing. Six registered gates build their own fixture, because `all`
  does not.
- 21 new tests, among them uiblock, schacct, heapmap, heapscrl, dtfield,
  dtwrite, blobsum, hdboot, bootstatus, vgadirty and trktxsurf.
- docs/FIELD-NOTES.md 35: six rounds against one 86Box XT, and 35.8 is
  closed — the bisect that was not a bisect.


jggonz:
Reviewed and merged. `main` is in (it had gained #117, Sheet and Chart), and I have pushed the merge plus fixes onto this branch, so the PR is mergeable now.

Everything below is a defect somebody would have hit. None of it is style, and none of it is a design disagreement — the shape of this branch is yours and I have not touched it.

---

## The merge would not have assembled, and git said nothing

You renamed `OSAPI_FONT_CHAR`/`OSAPI_FONT_STR` to `*_XPARENT` at the **same addresses**, which is the right way to do it — a package built against `main`'s SDK still calls the same cells, so nothing on a floppy breaks. But `main` gained Sheet, Chart and the float test while you were writing SPEC.md 6.6, and those files have never heard the new names: 17 call sites in `apps/sheet/sheet.asm`, 4 in `apps/fptest/fptest.asm`, two comments in `apps/os88chart.inc`. Git merged all of it cleanly into a tree that stops at `symbol 'OSAPI_FONT_STR' not defined`.

Renamed, and both files registered in `tests/textsites.txt` at the count they have — which is what your own §6.6.3 says the ratchet is for. Both lines are honest about being **mixed**: a shaded cell's label and the bold overstrike are real §6.6.2 cases, and a `gfx_fill` three instructions above a `font_str_x` is the pair, so those carry `backlog:`. `docs/INDEX.md` (main's generated index) also had never seen your six new slots, and is regenerated.

I checked the slot table by address, mechanically: **nothing moved and nothing was retired.** Two names changed at their own addresses and six new slots start at a fresh block, 0x04A8–0x04D0. That is exactly the rule.

---

## Fixed — the boot instruments first, because those are the ones nobody can debug

**The kernel-load canary was aimed at a sector that always loads correctly.** `KSIG_OFF` became a *memory* offset when stage 2 moved in front of the image (§2.9), and the number stayed at the one that used to name file sector 36. It names sector **49** now, and 49 is in a transfer run's **first half** on 360KB and on 1.44MB. The whole argument for the canary is that a run reads correctly up to the head boundary and only goes wrong after it — so on exactly the machine the canary exists to catch, a scrambled kernel would have booted with the canary saying nothing was wrong. What that would look like to a user is the same as it looked before the canary existed: a machine that boots into garbage with no message.

`tests/unit/t_canary.py` — the fast-tier row written to prevent precisely this, whose docstring says *"which is how the first one shipped wrong"* — was green the whole time, because it was still deriving the sector from the memory offset. `KSIG_OFF` is 11776 now, which puts the probe back on file sector 36, the same sector your own comment argues for and still in the middle of the band 33..38. The gate adds `BOOT2_SECS` before asking, reads the word at the padded offset, and checks `boot/boot2.asm`'s own `KSIG_OFF equ` against the Makefile's, because nothing did.

**`make rdiag` reported every sector bad on a perfect floppy.** The rdiag boot sector is built from `boot/boot.asm` without `-DFLAT_PAYLOAD`, which the four other flat payloads all carry. Without it stage 1 reads 13 sectors instead of 207 and enters them as though they were the blob. Booted in QEMU it prints `bad=00CE first=0001`; `main` prints `bad=0000 first=FFFF`. This is the instrument you reach for when a machine will not load its kernel, answering its one question with a confident lie — and it would have sent someone chasing a drive that was fine.

**`BOOTSTOP=2` had become `BOOTSTOP=1`.** The knob came down into `boot2.asm` with the loader, but `%define BOOT_NOSPLASH` did not come with it, so `%ifndef BOOT_NOSPLASH` was always true: the splash still ran, and the `*` on the text screen the splash never took over was gone too. Both halves are the point of `2` — a fault *inside* the splash cannot be reached, and a machine that loops instead of halting leaves something legible behind. Measured: before the fix the two kernels were byte-identical (md5 `7551bc3c…`); after it they differ.

**`make field` did not build at all, and the error named the wrong file.** `BOOTHD_DEFS` takes `BLOB_SEG` and `SPL_FSEG` out of the kernel, but asked `kernsize.py` about `build/` rather than `$(BUILD)` and asked `os88sym` for a *plain* map — so under `$(FIELDKNOBS)` os88sym refused, correctly, and because the extraction sat inside `$$(…)` on the nasm line its exit status was thrown away and both `-D` options simply vanished. What printed was `boot/boothd.asm:228: symbol BLOB_SEG not defined`, which points at the sector rather than at the two constants. It took all five field disks with it, `cqdiag.img` included — the diagnostic floppy for a machine that will not start, killed by the change whose whole subject is how a machine starts. The extraction is its own step now and follows the build's own knobs.

**`gfx_blitp` asked `[vid_mono]` before `GFXDENTER`,** so on the two-card machine (§39.12) it answered for whichever display was last active rather than the one the block is bound for. Both mistakes at once: a block for the mono card sails past the refusal and the planar body writes `VGA_SEG` at it, and a block for the VGA is refused because the *other* card is 1bpp. `gfx_blit4` and `gfx_blit1`'s `.pen` both already ask it after the hook; this was the one that did not.

**A pressed New/Folder button painted two white bars where its captions are.** `fdlg_text`'s opaque conversion hard-codes `CWHITE` as the paper, and `fdlg_btn2`'s pressed branch fills the interior `CBLACK` and then letters in `CWHITE` — ink and paper both white, so `font_run` lays a solid block. The paper follows the ground now, in `.text` beside the strings because this is `nasm -f bin` and `.bss` is not zeroed, and it is put back unconditionally so the up path cannot leave it black.

**`icon_draw_x`'s four refusals kept the one-shot icon pen.** §25.6 says the draw puts the pair back; a refusal returned without doing it, leaving the pen armed for whatever the caller drew next — on the failing path, which is the one time nobody is looking. Latent today (`os88ui_glyph` is the only setter and its record is decided at assembly time) and one instruction.

**`tests/heapmap.py` — your own new soak row — could only ever pass on a tree where `build/os8088.img` was missing.** It builds its fixture by scraping `make -n`, and an up-to-date tree prints *"is up to date"* and no recipe, so the assert fired. That is every tree that has been built, which is every tree you would test on.

**SPEC.md 5.4.2.2 published `gfx_blit1_pen` at slot 0x04A0**, which is `OSAPI_DRV_CALL_AT`. The kernel and the SDK both say 0x04A8. §25.6 gained the two numbers it had none of.

---

## Fixed — the ones a user meets on a machine, not in a build

**A hard disk the BIOS already knew about got a second row, and every partition on it mounted twice.** The IDE rung gave its own row to a drive rung 0 had already reported, and since §52.6.1 made the automount start from every probed device, that meant C: and D: over the same sectors — two icons per volume and **two FAT caches over one FAT**. Pairing is by geometry, because it is the only thing the two rungs both answer for (heads and sectors/track equal; the cylinder count only no larger, since a drive-type table reserves the last cylinder or two and SeaBIOS does too), and each BIOS row is spent once so two identical drives with only the first in the BIOS still list both. The case no key can separate — an MFM drive and an unknown IDE drive of exactly the same geometry — is named in the code and errs toward skipping the second.

**A geometry you typed on a probed drive was thrown away at every boot, and then overwritten in `SYSTEM.CFG` by the probe's.** The restore guard read "the probe answered" as "nobody typed one", but the ± editor is live on every drive, so those are not the same fact. The blob's flags byte carries `HDC_F_TYPED` now. The record that travelled between the two 5150s in FIELD-NOTES §33 is mounted and not typed, so it still loses to the probe and that note stays fixed — which the section now says, along with correcting a claim it made about `page.inc` that `page.inc` does not perform.

**A machine with no `SYSTEM.CFG` applied display settings out of uninitialised `.bss`.** `OVLCALL` banks the flags, so `ovl_cfg_load`'s `CF` answer was discarded at the `popf` and `drv_boot`'s `jc .load` could not be taken on any machine — the display block ran on a `drv_cfg` nothing had deserialised into. That is every fresh install. `OVLCALLC` is the CF-returning form, with `stc` on the dead arm so a retired blob still answers "the defaults stand", and it is deliberately a second name so it can never reach `dsk_xfer`'s per-sector site.

**`vid_apply` built the 348-word row table through the boot sector's stack.** The splash runs before `kmain`'s `mov ss, ax`, so 696 bytes went into whatever sat under that stack — and `[vid_rowmax]` was left *armed* across the switch, pointing every `gfx_rowbase` at a table `LOW_SEG` never had. Both the build and the arm are behind `cmp ss, LOW_SEG` now, which also takes 23.6 ms off the boot's critical path.

**`tools/os88ovlchk.py` had stopped judging `drv_task_x` at all.** 600 lines of `CFG_*` macro bodies sat inside its extent, so the tool saw `retf` mixed with the macros' eleven near `ret`s, declined to classify, and silently passed a near `call drv_task_x` from `.cold` — which is the defect that gate exists for. One label ends the extent; the same near call is refused again.

**A check box or radio button drew *nothing* when a clip fragment cut it.** `ico_core` clips an icon whole (§11.3), so the masked-sprite rewrite of `os88ui_glyph` turned every partially-covered control into an empty space where the fills it replaced had drawn the visible half — a background painter redrawing a control through a partly-covered window (ftpd's `fd_paint_now` is the shipped one) lost the control entirely. It asks `OSAPI_WM_CLIP_TEST` about its 12×12 box first now and keeps the per-pixel arm, *fixed*, for the answer; the one-call fast path pays one register-preserving test and nothing else, so PERFORMANCE.md's 24–34 ms win stands.

**Word: `wd_mpad` is back.** `wd_mink` took the pad byte that kept the `WDVAR` run even, so every word from `wd_mrx1` down — and `TY_BSS` with them — sat at an **odd offset**, which is a split word access on every 16-bit machine this runs on.

**Three registered test rows were writing their stub over live kernel bytes.** `gfx_pairtab0` moved to `.lowbss` in this branch, so its symbol is an offset in `LOW_SEG`; composed as `(KERNEL_SEG << 4) + offset` by `evqfull.py`, `linefast.py` and `os88linecost.py`, it addressed the middle of the kernel. They park on `snd_xlat` now.

**`sk_claim` could answer a bigger request with less memory.** The card ceiling *refused* an explicit `[sk_want]` instead of clamping it, and the refusal fell through into the auto ladder — which measures the heap against `SK_SHARESH`, the one test an explicit choice exists to skip. On an NE1000, picking 8K could hand you the 1024 floor while picking 4K got 4096. §72.13.3 already promised a clamp; now it is one.

**`make BUILD=<dir> <knob>` was baking the wrong hard-disk VBR.** `os88sym` hardcodes `build/` for both its includes and its byte-identity check, so a knob kernel built into its own directory got the *plain* kernel's `BLOB_SEG`/`SPL_FSEG` — a volume boot record publishing the blob at an offset that build does not use. `$OS88_BUILD` is the same idea `$OS88_DEFINES` already is. Measured: `BOOTPROF=1` now emits `-DSPL_FSEG=2101` where the plain build emits 2097; before, both said 2097.

**The idle task could sleep through the wake that was meant to end its sleep.** `sch_idle_body`'s `sti`/`hlt` was not the guarded form, so a keystroke or mouse packet arriving at the instant the machine went idle was halted straight through and nothing looked again until the next tick — up to 55 ms of input hitch on the target XT, on exactly the path UIBLOCK exists to make fast. `sch_wake_ui` sets a dedicated byte now, and the test, the halt and the clear are one `pushf`/`cli` window with `sti`'s shadow covering the `hlt`. A dedicated byte rather than a `T_STATE` test because under the fullscreen freeze `ui_task` is READY but ineligible, and testing that would spin for the whole exclusive bracket.

**`font_run`'s unaligned run under a clip region went to a cell renderer that owns whole bytes.** Routed to `font_run_scell`, which is blind to both the adapter and the phase, by testing both bytes rather than one.

**`band_emit_x` left `gfx_blit1`'s pen and polarity armed for the rest of the lock hold**, so a band declared upside down turned the *next* blit of the hold — an icon, a package's blit — over. `HEAPCOMPACT=0` also had `mem_avail` modelling a compaction the stubbed-out `mem_compact` cannot perform, and `docs/KERNEL-MEMORY.md`'s size baseline was three moves stale, so every `make` printed two rung-crossing banners that carried no information. Both baselines are blessed; `kernsize` now says `unchanged`.
---

## Flagged, not fixed

- **`make field` still fails one rule further on**, and that one is not yours: `build/herc.img` wants 379 clusters of a 360KB disk's 354. I built `make field` on `main` to check, and it fails at the same rule with 373. So the field disks have been over budget for a while and this branch adds six clusters to it; the decision about what comes off that disk is not a review's to take.
- **`KERN_BUDGET` 121,344 → 122,368.** Your PR body says so plainly and I have left it exactly as you wrote it. CLAUDE.md makes that the repository owner's call rather than a build fix, so it is on the merge decision and not on this branch.
- **A typed geometry still travels on the install floppy.** The fix below records a TYPED bit so the probe stops overwriting what a user typed on the machine they typed it on, which is what §52.6 already claims. It does not close the case where a record typed on machine A is carried to machine B — there is no room left in the record for a probed-geometry witness (`HDC_ESZ 8 × 4 + 2 = 34 = DRV_BLOB_SZ`). The other answer is to grey the ± fields on a probed drive (§47 rule 2), which is a design decision and yours.
- **`tools/os88ovlchk.py` cannot see past a `%macro`.** One label ends `drv_task_x`'s extent and restores the check there, but the general repair — end an extent at `%macro` and skip to `%endmacro` — would stop the next macro block anywhere in the tree from quietly disabling the same gate.
- **Twelve findings did not survive.** The review ran ten reviewers and then an adversary per unit whose job was to refute; twelve were refuted and are not in this list, including a couple I would have believed on a first read (a banked `BX` across `wm_lk`'s yield, `gfx_blit4`'s zero-size early-out reaching `GFXDLEAVE` on a stale hook). Your code answered all of them.

## What I looked at and it was fine

`wm_destroy_seg` and every other guard in `wm.inc` is byte-identical to `main`'s — that sweep has been lost in a merge here before and it is intact. Stage 1's blob read is 512-aligned and page-bounded, and the §2.9.7 checksum genuinely halts rather than entering a short read. The band buffer is a pinned 2KB claim, so there is no missing relocation proc, and a long title truncates in cells rather than overflowing. Every wake producer into the new blocking `ui_task` was traced and none is lost. `clockw.inc` was diffed routine by routine against `main:kernel/clock.inc` and the move is faithful. The driver ABI is intact: `FSV_SIZE` is still 30 verbs and the header format byte is still 4, so a `.DRV` built against `main` and this kernel still agree about the length of the table they both index — that pair has disagreed silently here before. The settings parser really is written once — four macros expanded twice — so there is no mirror to drift. No binaries rode in, `git ls-files build` is 0, and SPEC.md has no duplicate section numbers, which for a merge this size is unusual and good.

---

## How it was checked

`make` and `make small` both green, both kernels under their guards with `kernsize` saying `unchanged` — big has 4,608 spare (9 steps), small 2,048 (4). `python3 tools/os88test.py full` — **21 passed, 1 skipped** (`bootsmoke`, which needs MartyPC and I do not have one here). All seven shipped images fsck clean in three geometries. Every knob that reaches the boot sector or the hard-disk sector rebuilt: `FLOPPY1`, `DISKAL`, `BOOTDIAG`, `BOOTSTOP=1` and `=2`, `TRACKRUN`, `BOOTMARK`, `BOOTPROF`, `NOPS2`, `MOUIDSLOW`, `NOPLANE`, `NOUIBLOCK`, `BAND`, `GFXAUDIT`, `VGADIRTY`, `HEAPCOMPACT=0`, `HDD=32`. The C toolchain builds and `cword` compiles, so the `*_XPARENT` rename is consistent through the C SDK too.

Booted on VGA and on CGA. Paint drew a stroke on the four-plane canvas, saved `PICTURE.BMP`, and the disk verified clean afterwards with a sane 448×280 4bpp header — which is the end-to-end proof for §42.13 that a screenshot cannot give. The corner clock advanced across two samples 90 seconds apart, which is the proof the blocking `ui_task` still wakes.

One note that is not a defect: the PR body's first bullet says the composed title bar *"becomes kern_big's default"* and a later bullet says it is *"a knob again, not kern_big's default"*. The code agrees with the second — `BANDCOMP` is `%ifdef BAND`, so no shipped floppy compiles `band.inc`, and `kernel.asm` argues the reversal well (1,634 bytes against a measured win). Only the body is inconsistent, which is what a 64-commit squash does.

Nothing here is a criticism of the work. This is a large branch that moves the boot path, the scheduler and the drawing layer at once, and most of what I found is the kind of thing only a second pair of eyes catches — a constant that stopped meaning what its name says, a gate that was green about the wrong sector, a knob that quietly became its neighbour.

PR #117, opened 2026-08-26
Sheet and Chart: a worksheet, a charting package, and software IEEE-754 doubles (SPEC.md 81-84)

Koriban:
Two new packages and three new shared includes: a worksheet, a charting package, the
chart rasterizer both halves draw through, a multi-line text field for any package that
wants one, and a software IEEE-754 double.

**SHEET.O88** is a 4-sheet workbook — a sorted sparse cell array of 20-byte records over a
256 x 16384 grid, formulas with cycle detection and per-repaint memoization, 25 worksheet
functions, **error values**, absolute references, cell borders, number/alignment/font
formats, notes, a 5-command macro language, SYLK/DIF/BIFF read and write (including a
BIFF4 workbook that saves every sheet), a **block clipboard**, a **Sort that moves whole
rows**, both scroll bars, range selection, and Excel 2.1d's own menu bar. It
**claims `.SLK`/`.DIF`/`.BIF`**, so a spreadsheet opens on a double-click, and **Save As
asks for the format** rather than deriving it from whatever extension you typed.
**CHART.O88** reads a SYLK/DIF/BIFF file and draws **all seven** of Excel's chart types,
sharing its rasterizer with Sheet's live chart window so the two cannot drift.
**`apps/os88text.inc`** is the multi-line sibling of `os88line.inc`. **`apps/os88fp.inc`**
is IEEE-754 double arithmetic in software, with a **coprocessor path chosen at run time**.

The target throughout is Excel 2.1d, in look as well as function — checked against the
manuals and against screenshots of the real program running in a VM.

### Numbers are real

The value model is **IEEE-754 double**, not the integer this started as. `3.5` stays
`3.5`, `=A1/A2` is `1.75`, `SQRT(2)` is `1.41421`. `apps/os88fp.inc` implements it in
software; the kernel gained an 8087 probe published through the existing
`OSAPI_CPU_INFO` as `CPU_F_X87` (no new slot), and `fp_init` switches a dispatch table
to the coprocessor path when the part is there.

**Both paths are held to the same suite.** `apps/fptest` runs every case twice — once
with `fp_hw` forced to 0, once as detected — and prints the two failure counts side by
side. `os88fp ALL PASS  soft 0  8087 0` is the acceptance test, and it is
screenshot 22. QEMU always has an x87, so the emulator exercises the hardware path; the
86Box machines that specify no FPU exercise the other.

**The charts are real too, and were not** (§82.13). Everything the chart touched was a
signed word, so 43.6 drew a bar of 43 under a scale that said 44, and a column in millions
arrived already wrapped — the chart drawn from it was not so much wrong as unrelated.
A bar's length is `value * span / max`, which is unchanged if the whole series is
multiplied by one power of ten, so the fix is one exponent per series rather than floating
point in the drawing: `ch_scale` picks it, the words the rasterizer already read stay
words, and only the labels are told. A **negative** exponent is what makes the millions
chartable at all. CHART.O88 gained `os88fp.inc` in the process, because its BIFF reader
accepted the one RK subtype that is an integer and **skipped the cell otherwise** — a file
holding 43.6 charted as an empty sheet, in silence. It reads all four subtypes and the
NUMBER record now (screenshots 26–28).

### Errors are values

A formula with nothing to say produced **zero**, and a zero is an answer. `=A1/A2` with an
empty divisor read as 0, `=1+` read as 1, and a typo'd function name read as 0 and summed
into the total below it. `SH_T_ERR` and Excel's own `ERROR.TYPE` codes are the cure
(§81.20): `#DIV/0!`, `#VALUE!`, `#NAME?` and `#NUM!` are raised today, and they
**propagate** — one broken cell is visible at the total, through an operand or through a
range fold — and they clear themselves when whatever broke is fixed, because the cell is
stored by what it *is* rather than flagged separately. All three file formats carry an
error rather than the zero underneath it: BIFF in a FORMULA record's cached result and in
BOOLERR, SYLK as a bare `;K#DIV/0!`, DIF as its own `ERROR` indicator — which comes back
as `#N/A`, because DIF cannot say which and inventing one would be worse (screenshot 23).

### What is deliberately not here

`ISBLANK`, `ISNUMBER`, `ISERR` and `ISNA` are **absent on purpose**, and the reason has
changed twice during this work. It was the integer value model; that expired when doubles
landed. It was then also that no error value existed; that expired with §81.20. What
blocks them now is that **an argument is folded to a value before the function sees it** —
there is no reference left to ask about, so an empty cell and a cell holding 0 both arrive
as `0`. They need reference-typed arguments in the evaluator, not a wider number. The
error codes are Excel's own `ERROR.TYPE` numbers precisely so that `ISERR`, `ISNA` and
`ERROR.TYPE` are a table lookup once that lands. `SH_T_BOOL` is reserved and nothing
produces one. A version of any of them returning a plausible constant would be worse than
its absence.

`#REF!` and `#NULL!` have codes and spellings but nothing raises them: the reference
rewriters clamp rather than invalidate, which is a real gap and is named as one in §81.20.
A formula cell that references a **label** still shows the number underneath rather than
the label's text — `=B4` where B4 holds `hello` reads `0`, as it always has; only
arithmetic on one is an error. Making a formula carry text needs the formula writeback to
own an arena string, which is its own change.

The **11 database functions** and the Data menu's database half wait on a database app.
**Print** is a stub that says so in the status bar, because there is no print backend.
**Exit** is absent from Sheet's File menu — the OS menu owns closing a package.

Two honest limits inside the charts, both commented at the code: the **horizontal Bar
chart gets no value scale** (its value axis runs along X, so a scale would have to come
out of the bars), and the **pie's legend stops at eight rows**, which is what the canvas
holds.

### SPEC.md

Four new sections, appended after §80, with no existing section renumbered:

- **§81 SHEET** (21 subsections) — cell record layout, the claims, the formula grammar,
  reference rewriting, the format byte, the menu bar, dialogs, file formats including the
  BIFF4 workbook, the macro language, text cells, error values, the bss assertion, and
  §81.21's buffer overrun.
- **§82 CHART** (13) — the offscreen 4bpp canvas and why it is forced rather than chosen,
  `ch_bars_draw`'s segment contract, the gallery, the pie, two-series types, §82.9's band
  derivation, §82.10–§82.12 on the register clobber, labels-as-data and the column picker,
  and §82.13's decimal exponent.
- **§83** (3) — `apps/os88line.inc` and `apps/os88text.inc`, neither of which had SPEC
  coverage before this.
- **§84 software floating point** (7) — the representation, the operations, the decimal
  conversions, and the coprocessor path.

`python3 tools/checkdocs.py` passes (1441 headings, 143 slots, 0 problems), and
`grep -oE '^ *#+ [0-9.]+' SPEC.md | sed 's/^ *//' | sort | uniq -d` is empty.

### Bugs found along the way

Fifteen. Fourteen failed silently — nothing crashed and nothing was reported — and the
fifteenth stopped the machine outright. Five share one shape and it is worth naming up
front: **a command written before multi-cell selection existed, never taught to read the
extent**, or **a register that was correct when it was written and destroyed by a loop
added underneath it**.

**A column 11 characters wide killed the machine** (§81.21), and this one is worth reading
even if the rest are skipped, because it is the only one here that is **pre-existing rather
than introduced by this branch**. `sh_blank` is the row of spaces a blank cell paints as,
and it was 11 bytes: `SH_CW_WIDE / 8` plus a NUL — exactly right while Column Width was
three radio buttons whose widest preset was ten characters. The numeric Column Width
dialog that replaced them validates 1..40 and nothing resized the buffer. A width of 12
writes 13 bytes, and the two that fall off the end land on the next word in the bss
chain: **`sh_chartseg`**, the segment of the chart window's canvas. The next chart opens
by clearing that canvas with a 19,200-byte `rep stosb`, which then goes wherever the
corrupted word points — **the whole machine stops**, not the app, and the chart window
shows the uninitialised memory it never wrote.

It was found by **rebuilding the Kodak worked example by hand**, where widening a column
to fit `Photographic` is simply what one does. The symptom named the wrong subsystem
entirely — a chart of noise, in code that had just changed — and charting from a bare
column, a labelled column, a second sheet and after a BIFF save all had to come back
clean before a column width typed several minutes earlier was what was left to try.
`sh_blank` is `SH_CW_MAXCH + 1` now and `sh_chartseg` follows it by the same constant, so
both track the cap rather than a number that was true once. `sh_numbuf` already did; this
was missed in the same pass. **A segment word is the worst possible neighbour** — every
other overrun corrupts a value, this one corrupts an address, and the next large `rep`
turns two bytes into a dead machine.

**Every dialog leaked its window record.** `OSAPI_WM_CLOSE` means "quit the instance
owning this window"; a dialog has no owning instance, so it fell through to a plain
`wm_hide` — the pixels came down and the slot stayed allocated. `MAX_WIN` is 12, so about
ten dialog opens into a session no dialog would open again, in Sheet **or any other app**.
Fixed with `OSAPI_WM_DESTROY`, whose own documentation names this case. Verified by 17
consecutive open/close cycles where 10 had been fatal. It was pre-existing rather than a
regression, established by running the same test against untouched code.

**A category's pitch was a constant, and three things were wrong with that** (§82.9). A
column chart drew bars at a fixed 6px pitch, so nine bars used 54 of the plot's 218 pixels
and bunched against the left edge. `ch_xfor` already spread Line and Area across the whole
plot, so a Line and a Column chart of the *same* data disagreed about where category five
was. And the horizontal Bar chart runs its categories down a **160**-row canvas while
`CH_MAXBARS` = 40 was derived from the *column* axis — so from index 27 on, `ch_fillrect`'s
clamp folded every remaining bar onto the last row and a thirty-bar chart drew
twenty-seven. `ch_band` now divides the axis by the count and `ch_inset` takes a third of
each band as the gap; measured after, nine bars are 16px wide with a uniform 8px gap
(Excel's 50% gap width) spanning the plot end to end, and thirty bars are thirty.

**Fill Right and Fill Down ignored the selection** (§81.13). Both were written when a
selection was one cell; range selection arrived in stage 3.0a and they were never taught
to read `sh_selcol2`/`sh_selrow2`. Selecting a block and choosing Fill Down changed exactly
one cell and left the selection drawn over cells it had not touched. The subtle half is the
delta: both used a fixed `(+1,0)`/`(0,+1)` shift, so the naive repair fills the block and
puts `=A2` in every row of it. `sh_fill_copy` computes `destination - source` — screenshot
9 is a fill four rows deep where the relative column ends at `=A4` and the `$A$1` beside it
stays pinned.

**The chart's scan cap bounded the wrong thing.** `CT_TCAP` capped cells *collected* while
reading, with the lowest-column filter running afterwards — so on a wide sheet the arrays
filled with other columns, the chosen column's tail was lost, and the column choice itself
could be wrong. Both draw a plausible chart. The filter now runs as the file is read;
retiring the oversized arrays took chart's bss from 1752 to 456 bytes at the time (it is
757 today, now that the two-series types and the column picker carry state of their own).

**A whole-VM hang** from reading own bss *after* pointing DS at another segment: garbage
in, poisoned ES out, and a 19,200-byte `rep stosb` into low memory. Fixed by reading first,
then by making "the array's segment arrives in DX and DS is never touched" an interface
rule rather than a fix.

**A window pointer left in a clobbered register** (§82.10). `ct_ondlg` stashes the window
in BX (`mov bx, si ; stashed`) and calls `ct_render`, which banked AX/CX/DX/SI/ES but not
BX — and `ch_draw` clobbers BX by its own documented contract. So `mov si, bx` handed
`ct_paint` a garbage window and a 240x160 image was blitted through coordinates taken from
unrelated memory. **What it destroyed depended on the data**, since the value left in BX is
wherever `ch_draw`'s arithmetic ended: one document wiped the kernel menu bar and the chart
window's own frame, another blitted somewhere inert so the canvas merely stayed black and
read as "Open failed", and picking a Gallery type afterwards fixed it because that path
restores SI through its own push/pop. Three symptoms, one missing `push`.

**A label charted as the number zero** (§82.11). `ct_parse_c` recorded any `K` field as a
number, and `ct_pint` on `K"Rent"` parses the opening quote as `0`. A column of row
headings charted as a row of zero bars — and being the lowest column, that was the whole
chart — while every header cell became a spurious leading zero in its own column. The other
two readers already had this right (BIFF records only RK and never LABEL; DIF skips its
type 1), which is what makes it a defect rather than a design.

**A `ret` to nowhere.** `ch_legend` pushed SI at entry and never popped it, so its `ret`
took the saved register as a return address — a black canvas and a wedged app (§82.7.3).
That one is now a **gate**: `tools/stkbalance.py` walks each routine's control-flow graph
and requires every `ret` to be reached at the depth it started at. A naive push-versus-pop
count flags 409 of 4130 routines, so the walk is path-aware; deliberate stack surgery
declares itself with `; STKBALANCE-NET: ±n`. It is registered in the fast tier, scoped to
these packages and their includes — the kernel's ISR tails push in one label and pop in
another, which a per-routine walk cannot follow.

**A formula wore the identity of the last cell it referenced** (§81.16). `sh_getcell2`
reads a cell's format, type tag and text offset into three bss bytes and *then* calls the
evaluator, which recurses back through the same routine for every cell the formula names.
So an `=AVERAGE(B2:C2)` that was never formatted drew as `$90` because C2 is Currency, and
`=A2+0` where A2 holds the label `Ann` drew **Ann** — the painter took the label path and
printed another cell's words where a number belonged. The stored value was right the whole
time, which is what makes it bad: the grid is what anyone reads.

**Number, Alignment, Font and Clear applied to the anchor only** (§81.17) — the same
never-taught-about-the-extent shape as Fill. Worth recording that the first repair went
into `sh_docmd_clear`, which **has no callers**: the live path is `sh_fdlg_apply`'s
`.doclear`. Dead code assembles and tests exactly like fixed code.

**Copy formatted the cell with a routine that had stopped being true** (§81.18). It ran an
integer `itoa` over the word at `SH_C_VAL`, which since stage 4.0 is the low quarter of a
double — meaningless for anything but a small whole number — and had no case for a label
at all. `sh_cell_totext` covers values, labels and formulas with one representation, which
is also what Sort needed.

**Sort reordered one column and left its neighbours where they were** (§81.19), which does
not sort a table so much as break the correspondence between its columns — silently, and
the sheet looks perfectly ordinary afterwards. It moves whole rows now (screenshot 25).
Two faults building it, both the recurring shape: `sh_cell_totext` did not bank **DX**,
which was the carry loop's index, so the loop never advanced and the machine stopped dead;
and `sh_sort_permcol` moves `sh_selcol`, so the later "skip the key column" test read
whichever column had been carried last and the key was permuted a second time on top of
its own sort.

**The error-value tag was banked across the evaluation that decides it** (§81.20). Adding
`sh_curaux` to §81.16's bank looked like the obvious completion of that fix and is wrong
twice over: the banked copy predates the evaluation, so a cell whose divisor had since
been fixed still painted `#DIV/0!`, and one that was still broken painted the `#ERR`
fallback, its code having been overwritten by the last cell its formula referenced. Bank
what the callee cannot change; take what it can *from* the callee.

**A pop order that balances and is still wrong** (§82.13.1). `ch_scale` pushes
`ax bx cx dx si di es` and unwound them `es di dx si cx bx ax` — SI and DX exchanged. The
stack balances, so `stkbalance` passes and the build is clean; every caller simply gets
two registers back swapped. Both callers happen to bank those registers themselves, so
nothing failed — which is the worst version of it: a latent fault the next caller inherits.

**A stride helper that folded in the base address** (§82.13.1). Widening four chart arrays
from words to doubles means a stride-8 multiply at every site, so it went into one routine
— which returned `&ct_tval[i]` rather than the offset. Two of its four callers wanted a
different array: the second series wrote over the first, and the sort swapped the wrong
one. Neither showed in any chart tested up to that point, because both need data the
earlier fixtures did not have; the one that catches the sort is a SYLK file whose rows are
written 4, 1, 3, 2, and it charts 40, 30, 15, 5.

### Conventions this got wrong the first time

Both packages shipped without an **`OSAPI_ABOUT_SET` handler** (§12.2) — no
`About Sheet` above `Close` in the name menu — while seventeen other packages had one,
because a `Help > About` was invented instead. That is Excel's convention, not this OS's,
and copying the application is not the same as reading the platform. Both declare one now
(screenshot 2), and `docs/INDEX.md` — generated by `tools/os88index.py`, gated by every
`make` — exists so the next package can ask "does this already exist" before designing.

Sheet's **dialog buttons hung half outside their windows**, because heights were literals
rather than derived from `wm_content` (content height is `W_H - TITLE_H - 1`). Every
dialog now derives its height from its lowest element.

**Enter after a run of Tabs** returns to the column the row's entry began in, as Excel
does, instead of walking the cursor diagonally (§81.12).

### Verified

QEMU, driven over QMP, with the screendumps below. Every feature named above was exercised
in the running OS, not just assembled.

```sh
make clean && make                    # all three geometries
python3 tools/os88test.py full        # 19 passed, 0 failed, 0 skipped
python3 tools/checkdocs.py            # 1441 headings, 143 slots, 0 problems
python3 tools/stkbalance.py <those six files>   # 0 unbalanced paths
python3 tools/os88disk.py --verify build/apps360.img
make test                             # then tools/mouse.py + tools/shot.py over QMP
make test TESTAPPS=build/fptest.img   # the soft-vs-8087 acceptance test
```

The only build warnings are nine pre-existing `os88face` 1-pixel-stem notes on the shipped
faces, none from these packages; `nasm` runs with `-w+error`, so an assembler warning is a
build failure.

**All three geometries were checked, not only QEMU.** `SHEET.O88` and `CHART.O88` are
present on the 1.44MB, 720KB and 360KB apps disks, and `os88disk.py --verify
build/apps360.img` reports OK (FAT12, 354 clusters, 316 in use). Worth flagging for
review: that is **38 clusters of headroom on the 360KB disk**, and this branch spent 8 of
them — CHART.O88 grew from 6,486 to 11,468 bytes when it gained `os88fp.inc`. The disk
still builds and verifies, but it is the tightest of the three and the next package to
land on it should check.

Worth noting for review: the pathological cases were built deliberately rather than hoped
for. The `CT_TCAP` fix was proven with a SYLK file holding 300 cells in column 5 followed
by eight in column 1 — the fixed build charts column 1, the pre-change build draws nothing.
The Bar-chart clamp was proven by counting blue row-runs in the framebuffer, not by eye: 30
values give 30 distinct 4px bars where the previous build gave 27. And the stack gate was
proven to bite by adding a stray `push` and watching the row fail.

### Screenshots

28 cropped screendumps, taken from the running OS over QMP. They live on a
[branch of their own](https://github.com/Koriban/os8088/tree/pr-screenshots) in the fork,
so nothing binary lands on the code branch.

**1.** Sheet: the menu bar, formula bar, status bar and both scroll bars

**2.** the name menu — `About Sheet` above `Close` (§12.2)

**3.** doubles and functions: the formula bar reads `=SQRT(2)`, the cell reads `1.41421`

**4.** the Formula menu

**5.** Paste Function — the list dialog, scrolling

**6.** Format Number — buttons inside the frame

**7.** Note — the multi-line field (`os88text.inc`) with its caret

**8.** Column Width — the single-line field (`os88line.inc`)

**9.** absolute references: a fill four rows deep, `=A4` beside a pinned `$A$1`

**10.** Save As asks for the **File Format**, preselected to the open document's

**11.** `.SLK`, `.DIF` and `.BIF` carrying Sheet's icon — all three claimed

**12.** the Gallery — all seven types

**13.** Column, 30 categories filling the plot

**14.** Bar, 30 categories down the canvas — the clamp that used to eat three

**15.** Line

**16.** Area

**17.** Pie, with its legend

**18.** Combination — the second series over its own columns

**19.** a budget whose first column is row headings, charting its **figures** (§82.11)

**20.** `Data > Column` — asking which column instead of guessing (§82.12)

**21.** a `Year, Population` sheet charted by **Column B**: the population, not the years

**22.** `os88fp ALL PASS  soft 0  8087 0`

**23.** error values: `=A1/A2` in the formula bar, `#DIV/0!` in the cell, propagated down the column, with `#NAME?`, `#NUM!` and `#VALUE!` beside it

**24.** the block clipboard: a 2x3 block pasted three columns over, its formula re-indexed to `=H1+H2`

**25.** Sort moved whole **rows** — column B carried by column A's key

**26.** a chart of 43.6, 44.1, 12.25 and 0.5, scaled to `44.1` (§82.13)

**27.** a chart in **millions** — 3,400,000 charts, where a signed word could not hold it

**28.** Pie, with a legend that reads the actual values rather than their truncations


tools/os88index.py
```diff
@@ -0,0 +1,322 @@
+#!/usr/bin/env python3
```
jggonz, line 1: Interesting tool. Lets see how it works out.

jggonz:
Reviewed in depth, and I've pushed **8 commits (50 fixes) to your branch** — the PR stays mergeable and everything is verified: `make`, `make KERN_SMALL=1`, both kernels under their size guards, `os88test.py full` 17/17, all seven floppy images fsck clean, and every headline fix exercised end-to-end in the booted OS (plus a CGA pass, since paint changed).

First: this is a remarkable piece of work. The features are real — I typed `=SQRT(2)`, watched `#DIV/0!` propagate and then clear, and pulled a byte-perfect SYLK file with the cached error in it off the floppy image before reading a line of code. And your PR body's own diagnosis ("a command written before multi-cell selection existed, never taught to read the extent... a register correct when written, destroyed by a loop added underneath") turned out to be the exact map to what remained: most of what follows is stage-4.0's 8-byte widening reaching code that still moved one word. A review team of ten reviewers plus adversarial verifiers confirmed 42 findings (none refuted); the fixes below keep your design everywhere — same idioms, same comment style, SPEC updated in the same edits.

## What was fixed, by commit

### `849233e` — the cell store survives Insert/Delete Row/Column

- **Insert Row after a save destroyed every plain number in the document.** The staging record's value field grew to 8 bytes and the reinsert loop reads 4 words back — but the staging *write* still stored one word, so six of every number's eight bytes came back as whatever the staging segment last held (the file text of your last save, in practice). The write is now the same 4-word loop the read already was. Verified live: 1.5/2.5/3.5, save, Insert Row — all three survive.
- **The same round trip dropped the type tag**, so every label became a number (its text orphaned in the arena) and loaded error cells lost their codes. The staging layout gained TYPE/AUX bytes (free inside the SH_C_SZ stride).
- **File > New and Open kept the old document's notes and borders**, and a stale note offset into the reset arena would show a fragment of some other cell's formula as the note. All three readers and sh_new now clear the note/border tables and the arena together.
- `sh_removecell` still shifted with the pre-widening `12` — in-bounds today by an 8-byte accident, one token (`SH_C_SZ`) to make it true by construction.

### `226523f` — the evaluator answers for what it was asked

- **Any nested function call computed wrong**: `=SUM(10,MAX(1,2))` showed 4, because `sh_pfunc` banked one word of the 8-byte accumulator. It banks all four now; the cell shows 12 (verified in the booted OS).
- **Fill Down truncated 3.5 to 3, and filled a row of headings as the number 0** — the same 81.18 defect class you fixed in Copy, still present in Fill. It now branches on the cell type and stores the full double (or the label).
- **A fixed `#DIV/0!` resurrected itself** whenever the erroring cell was first reached through another formula: the error was raised from the *stored* tag before re-evaluation, then written back. The raise now happens after the evaluation, from the published tag — 81.20's own rule, applied to its own machinery.
- **`=IF(B1=0,0,A1/B1)` showed `#DIV/0!` when B1 was 0** — the untaken branch's parse still raised the sticky error. The condition is decided before either branch parses, so the untaken branch's raises are banked and restored.
- **`=SUM(A1^2,B1)` silently dropped B1** (and `=SUM(A1:A9^2)` answered last-cell-squared): the bare-reference continuation chain had no power level. It does now, and an unparseable argument tail raises `#VALUE!` instead of answering.
- **Nothing bounded parser recursion**, and a formula six memoization-cold cells deep could run ~2.4KB of stack through task 0's 1,024 bytes into kernel disk buffers. A shared nesting budget (SH_PNEST_MAX, sized so your design's own depth-6 fold chain still fits exactly) refuses with `#VALUE!` instead.
- **`=SUM(A1:A16384)` walked all 16,384 coordinates** through a binary search each, per repaint — seconds per keystroke on the 8088. It now searches once and scans the sorted record array across the row span, O(records) instead of O(area × log n).

### `3b421b3` — damage-rect the paint hot paths

- **Every keystroke and arrow move was a full-window repaint plus a recalc pass** — about a second per event on the 4.77 MHz target, which is dropped keystrokes by construction (and invisible in QEMU). A keystroke now redraws only the formula-bar field; a selection move redraws the outgoing and incoming cells plus the bars; the recalc pass only runs when data actually changed.
- **One-row scrolls redrew everything; `OSAPI_GFX_SCROLL` was never called.** Vertical scrolls now blit the surviving rows and letter only the vacated ones (falling back to the full repaint when the clip refuses); horizontal scrolls repaint the grid and headers only.
- **A far Goto on a nearly-empty sheet could crash the machine**: the scroll-thumb math divides `pos*track/total` with pos unclamped, and a 16-bit quotient overflow raises INT 0 on a real 8086. `sh_sbsync` now clamps pos to total−fit at the one choke point both draw and hit-test share.
- **With a chart open, every keystroke re-rasterized the 19KB canvas** for pixels that never changed. A dirty byte set in the two low-level cell mutators now gates the resync, so nothing can forget to set it.
- **`os88text.inc` — the shared include — shipped the exact fill-then-letter flash `os88line.inc` exists to prevent** (its own header even claimed the one-pass design). It draws the opaque run first and fills only the margins now, so every future consumer inherits the right shape instead of the flash. The Note dialog also stops repainting its buttons and labels per keystroke.

### `bf5dd35` — sort keeps labels; macro and Goto fixes

- **Sorting a table with a text header silently replaced the header with 0** — even when the sort moved nothing. Label and error cells now sit the sort out (whole row stays put, the same clip-don't-crash policy your SH_SORT_FCAP already uses). Verified live: "sales" survives, 1/2/3 ordered below it.
- **`=SET.VALUE(B1,A1+A2)` stored parser scratch from AX** — the evaluator's result has lived in `sh_acc` since stage 4.0. It stores the actual (decimal-preserving) value now.
- **`=SELECT(C5)` handed the kernel the macro parse cursor as a window pointer**, white-filling the screen at garbage coordinates and leaving the selection extent stale. It routes through `sh_select` — your design's existing shape.
- **Every Formula > Goto silently defined a bogus name** ("B7") until the name table filled: the `.goto` arm fell through into `.defname`. One `jmp .out`.
- **Pasting a 200-character line into B3 smeared fragments into B4–B6**: the truncation left the parse cursor mid-cell, so the 64th character read as a row terminator. The tail is skipped to the real terminator now.
- **A sort that filled the text arena half-applied its permutation silently** — rows left in two orderings at once. The arena refusal now propagates (CF through sh_settext/sh_setformula/sh_commit) and the sort stops with "Sort incomplete - text area full."

### `a71bf62` — file I/O trusts nothing

- **The SYLK writer's 56-byte room check predated quoted labels**; a label-heavy save could write ~111 bytes past the exactly-32KB staging claim into the next heap block. The reserve is 200 now, sized from the emit paths that exist today (the BIFF cell writer had the same hole 4 bytes wide — NUMBER writes 18 against a 14-byte reserve — fixed alongside).
- **Four hostile bytes hung the app forever**: a BIFF record length of 0xFFFC wraps the 16-bit advance back onto its own header. One carry-checked bound at `.rechdr` now refuses any record that wraps or passes the file end — covering every unknown-opcode skip too. Verified with a crafted file: clean refusal, no hang.
- **A failed two-sheet save reported "Saved - THIS SHEET ONLY"** — the `.warn` tail overwrote the error message. It only upgrades a genuine "Saved" now; verified both ways against an ejected floppy.
- **A truncated save said plain "Saved"** while the tail of the data wasn't in the file. All four writers now say "Saved - TRUNCATED; sheet too large for this format."
- **A foreign .BIF with row ≥ 16384 planted cells on a different sheet** — the row's high bits OR into the packed sheet index. All five BIFF cell paths clamp row/col now (your SYLK reader already did), and the DIF reader gained the matching column bound.

### `e5bb86b` — CHART's bss chain healed

- **`ct_wantcol` was anchored on the wrong array** (`ct_t2val + CH_MAXBARS*2` instead of `ct_w2vals + ...`), so the fp scratch block overlaid live chart data and two output arrays sat up to 271 bytes *past the declared bss* — and the size assertion passed because the literal matched the broken chain. Re-anchored; OS88_BSS 1431 → 1831. Those 400 bytes were always being written — just outside the declaration.
- **CHART had the same 0xFFFC infinite loop as Sheet's reader — but holding the gfx lock**, so the whole desktop froze, not one window. Same carry-checked bounds, plus minimum body lengths for RK/NUMBER; the fix went one step past the prescription because the unrecognized-opcode path needed the bound too. Verified: the hostile file gets "No numeric data found" and the desktop stays live.
- **Opening a one-column file after a two-column one charted the OLD file's second column** — second-series state was never reset between reads. One shared reset at the top of all three readers.
- **A two-series scatter labeled the Y axis with the X series' maximum** — a confidently wrong scale. The value-axis labels are suppressed for that case, the same honesty your horizontal Bar chart already practices (SPEC 82.7.2/82.13 updated).
- Two latent register traps from your own 82.13.1 catalogue: the 8-byte swap resetting the sort's outer CX, and all four dialog name-copy loops counting in the register the copy destroys (the 12-char bound was dead code). Both fixed with the per-site register that's actually free.

### `ae9a621` — os88fp rounds right and shares the 8087 safely

- **Every operation computed a sticky bit and then threw it away** — `fp_pack_a` zeroed it before rounding, so tie-boundary results misround by 1 ulp against the IEEE double your own suite compares with. Each op now jams its lost bits where pack's rounding can see them (into the aligned operand for add/sub — a result-jam rounds the wrong way on unlike-sign ties — and into the result for mul/div). All 28 pre-existing cases are byte-identical on both paths; new cases pin the fix.
- **The 8087 path held live stack state across pre-emptive task switches with no kernel save**, and every instance's `fp_init` runs `fninit` — so recalculating while another fp window repainted could plant a garbage double in a cell, on real hardware only. Every `fpx_*` op (and the `fninit`) now runs inside the repo's `pushf`/`cli`…`popf` idiom — each guarded stretch is well under a tick.
- `fp_atof` clobbered AX against the register-discipline rule (and left SI advanced on a failed parse); `1e600` silently parsed as 1e88 instead of the documented clamp. Both fixed; SPEC 84.2 now states the saturation posture.
- **fptest grew from 61 to 70 cases** — the tie-sticky region, the overflow clamp, divide-by-zero's CF, the subnormal flush, negative exponents — and still reads `ALL PASS soft 0 8087 0`. One subtlety worth recording: the obvious tie-test operand (`…001`) *cannot* pass on a real 8087, which double-rounds through extended precision; the committed operands were chosen so soft, 8087 and host IEEE all agree.

### `cfe7ca6` — the docs tell the truth, the gates see their gaps

- **SPEC 81.1/81.2 still pinned the pre-widening 12-byte cell record** (12/1365/16KB vs the shipped 20/1638/32KB) — the binding contract would have steered the next contributor into writing a word over the type/aux bytes. Rewritten to the shipped layout.
- **The new `chart.bin` rule omitted `os88fp.inc`** — the exact dependency hole your own sheet-rule comment two screens up narrates fixing five times. Added.
- **`make combo` was broken** (sheet+chart add 57 clusters to the 360KB field disk). Both are in COMBO_DROP now with the reason recorded — see the flagged section below, because combo has a second problem that predates you.
- **BIFF-NOTES said Sheet writes only BIFF3** while the workbook path writes BIFF4 0409H/0443H — a maintainer trusting the table would have "corrected" your workbook writer into files nothing opens. The table now covers both.
- The stray SDK paragraph declaring CPU_F_X87 "THE ONE EXCEPTION" to hard rule 1 (contradicted by the rule itself and by your own CPU_F_X87 block twenty lines up) is deleted; the cpu_feat ownership comments in cpudet/xmem now describe the four-bit, two-writer contract.
- **The x87 probe now stores its sentinel *between* FNINIT and FNSTSW** — Intel's canonical FIND_NPX order. With zero spacing, a real 8087 still mid-FNINIT may miss the FNSTSW and the probe would report an empty socket on exactly the machine class the bit exists for. No emulator can show this either way; the reorder costs zero bytes, and SPEC 84.6's listing moved with it.
- `stkbalance` silently skipped routines that end in a tail jmp (3 today) — it now counts and prints them, and the suite row states the honest guarantee. `os88index --check` no longer fails every `make` over an untracked draft in docs/ (it lists tracked files).

## Flagged, not fixed

- **`make combo` still fails at 388/354 clusters with sheet and chart already dropped** — that 34-cluster overrun predates this PR. Which package goes next is the maintainer's call, not a review fix.
- **The pie's rasterization is on the order of ten seconds on the target** (~150K per-pixel plot calls, inside a gfx-lock-held handler). The sweep design is recorded in SPEC 82.6 and the project's rule is measure-before-redesign: a counter in `ch_pie_plot` priced by the table is the next step, and folding the pattern test into `ch_ray`'s loop would buy back roughly a third by call count alone.
- **fptest re-runs the whole suite inside its paint proc** (~1s per expose on the XT) and its fixed 440-line window overflows both 1bpp screens. It's a dev tool that ships on no disk, so it's noted rather than changed.
- **The remaining full-repaint surfaces** (paste, sort, fill, dialogs closing) still go through `sh_repaint`'s fill-then-drawall — those are genuinely many-cell events, but converting the header strips to padded FONT_RUNs would let the whole-window white fill go too.
- The sort could reuse an unchanged label's existing arena offset instead of appending a fresh copy (the honest-stop is in; the thrift is a natural follow-up). One stale comment near the sort staging sizing still says "SH_CELL_CAP=1365" (the conclusion it draws is still true at 1638).
- A range fold nested inside another fold's formula still shares the corner globals — an exposure the pre-rework code had too, unchanged in scope.

## Checked and cleared

No duplicate SPEC § numbers; no API slot moved and the api-abi cross-check (kernel table vs SDK) passes; nothing tracked from build/, no stray binaries (the screenshots branch stays on its own branch — nice touch); the 360KB apps disk really does have the 38 free clusters you flagged; all three geometries fsck clean; both BIFF hostile-file fixes verified against a crafted disk; the CGA selection-frame rendering I checked A/B against your original commit — identical, so nothing in the paint work changed the 1bpp look.

Every defect above lives in code that took real skill to write — five of the six worst are one widening reaching code written before it, which is the hardest kind of change to chase through 19,000 lines. The bug catalogue in your PR body made this review faster and better; thank you for writing it.

https://claude.ai/code/session_01WKeXq4e6nntXSx8MMjTJsb

PR #121, opened 2026-08-30
Elendilon -> Main (TANK ATTACK, Paint Optimization, Boot Splash Completeness, Save Under Pass 3, Bugfixing)

Elendilon:
* Save Under Pass 3
-- 1bbp caches for VGA windows that only use black and white
-- Live cache consideration instead of per open
-- Add newer apps that were missing flags
-- Partial dmg repayment, instead of full window redraws, for windows with workers
* Boot sequence enhancements
-- Full chrome and bar for all adapters
-- Text for all boot segments, instead of just drivers
-- Flicker fixed for the boot animation, and it draws each frame in 1/3rd the time
-- Animation timed to the clock, not progress bar updates
* PS2 mouse will init on more systems
* Added Tank Attack game
* 28 rounds of Paint fixes. Its a real app now.
* Bugfixes
-- Fixed the RNG. Droids can happen in cyclone now.
------------------------------------------------------------
The elendilon-new-2 branch, squashed onto main, with PR #116's review fixes carried across and main's own #117-#120 underneath.

WHAT IS IN IT

- **[vid_w] is the whole desktop, and a screen edge is not** (SPEC.md 11.95.2.1) - a bug main carries today, reported from the field on an 86Box XTDualVideoVGA. Before 39.16 the screen and the desktop were one rectangle, so every "does this reach the edge of the screen" question was asked against [vid_w] and was right. On a two-card desktop [vid_w] is the SUM, and three of those questions in wm.inc silently became about the far side of the OTHER monitor: wm_snap_ax walked a maximized window seven pixels right, wm_flush_ck then read it as not spanning and put the left border back on top of the gap, and wm_snap_w carried the same defect in the other axis. They ask wm_disp_xw now - the window's OWN display's x range, which is what the whole tree had before 39.16. +63 bytes, `tests/dispzoom.py` is the gate.
- A window that spans the screen has no RIGHT border either (SPEC.md 11.95.3). 11.95.2's argument is one sentence - a border separates a window from what is beside it, and at the screen's left edge there is nothing beside it - and it is word for word the argument for the right edge. Dropping only the left left the far edge wherever W_W-1 put it: a maximized window on VGA had 639 content pixels, 79 whole cells and SEVEN COLUMNS that could hold no cell at all. `wm_bord` is the value form ten sites were writing by hand; six are converted to it and the frame's flush case goes from three drawn sides to two. `NOFLUSHR=1` is the A/B for the right border alone. **This reaches main's WEAVE**: `wm_geom` now answers `w` for a flush window, so WEAVE's content grid is 80x17 on CGA and 90x35 on Hercules where it was 79 and 89. WEAVE-SPEC 7.1.1's derivation, `weavesim`'s ADAPTERS table and `tests/weavesmoke.py`'s frame model are updated together - weavesmoke's right-edge check is now conditional on flush exactly as its left-edge check already was, which is the same change 11.95.2 made to it once before.
- TANK ATTACK (SPEC.md 85, `apps/tank/`). A first-person wireframe tank duel that takes the whole screen: a Mode X backend behind SPEC.md 53.7's fence, a raster whose dirty span IS the design, fixed-point geometry with the ridge as a table at infinity, scenery that stops a shell, a difficulty curve over the first ten tanks, GAME OVER lettering that flies in and takes a halo because a 1bpp erase is not a colour, three-initial high scores in a file, and an attract window with a one-stroke outlined logo and a rolling score band. It is the tree's first page-flip: `fsx_page` (SPEC.md 53.10) is the kernel half, and docs/GFX-FSX-PLAN.md is what a vector game found past 53.7 - three apps now carry their own Bresenham, and the largest finding is deliberately NOT proposed as a slot.
- A straddling blit is CUT at the seam (SPEC.md 39.14.7.2). 39.14.7.1's gate was binary: a block that fits one display kept 5.4.1's fast path and a straddling one gave it up on ALL its pixels, not just the ones near the seam. Displays are edge to edge so there is exactly one seam; the block is cut there and each half hooked to its own display. Measured on a VGA+MDA XT with OS8088.GIF straddling: 625 ms against 27,859, a 44.6x. It cost 199 bytes, `NOBLITCUT=1` is the A/B, and `tests/blitcut.py` the gate.
- gfx_blitp asks WHICH DISPLAY before it asks HOW DEEP (SPEC.md 5.4.3.4) - the same motion PR #116's review asked for, arrived at independently here and now written down with the measurement behind it. Both directions were wrong: a VGA block entered while the mono card was current was refused and Paint converted a canvas that could have kept its planes, and a mono block entered while the VGA was current PASSED. Latent, and recorded as latent: on a VGA+MDA XT [vid_cur] was already the block's own display both times, because a window's frame is drawn on its own display earlier in the same lock hold. Safe by accident of draw order, now safe by mechanism.
- Paint resizes in place (SPEC.md 42.19-42.20). The resize walks its row addresses instead of computing them, needs no staging area at all, sets a row up once rather than once per plane, anchors the tool column so a resize does not move it, and the content area may now be WIDER than the canvas.
- A window that resizes itself by name gets 11.90.3 too (SPEC.md 11.90.3.1), and a maximize goes where it says it goes.
- Paint (SPEC.md 42.8, 42.13-42.18). The stroke is banked rather than drawn per sample: a run of leading edges becomes one rect, a diagonal step's two edges are one dab, the bank is paid down in the idle turns and the canvas catches up at the release. Undo moved up a level to once per CHORD, and a row holds AT MOST eight blocks. The canvas resizes - a grow fills its new bands instead of decoding them, a shrink that would lose ink gives back what it can, a pure shrink owes its content nothing - and a 'Save changes to PICTURE.BMP?' dialog stands between the user and the close.
- The loading screen (SPEC.md 15.3, 15.6, 2.9.9-2.9.11). It animates from IRQ0, so a BIOS disk call no longer locks it out; the spinner composes each row and stores it ONCE; the caption and percentage are blitted, which is 5.9 s off a floppy boot. The kernel starts its own screen, and the hard disk has its own boot sector that never heard about the blob.
- The PS/2 mouse (SPEC.md 9.9). The 8042's auxiliary port, probed after the serial ports: bit 5 of the command byte rather than silence from 0xA9, the port drained the moment it is enabled, and int 09h no longer eats the probe's replies. vm/386-ps2 is the machine and tests/ps2mouse the row.
- Save-under and raise caches (SPEC.md 11.96.16-11.96.18, 28.8-28.11). A window may declare its content TWO COLOURS and its cache then holds ONE plane; the renderer is told whose content it is drawing; a drag or a zoom over another window banks it; TeXPad, the Browser and the Task Manager each hold a cache on their own terms, and the Task Manager repairs at the restore rather than repainting.
- gfx_spans (SPEC.md 5.10): a run of scan lines in one call, with a gate against its own documented fallback, and gfx_sp_next answering in registers.
- Cyclone (SPEC.md 67.8, 67.9, 67.22): the AI droid hunts, the powerup drop gate stopped reading the LCG's dead bits, and a fully ascended enemy can be shot again.
- Memory pressure (SPEC.md 18.8, 50.3, 66.5): a FAT window is a CACHE the machine may take back and the kernel's own is a PIN; a donated claim moves, with the kernel as the other half of the callback; a driver's second image belongs at the top of the heap.

SPEC.md 85, NOT 81 - THE ONE THING A REVIEWER SHOULD LOOK AT

TANK ATTACK was written as SPEC.md 81 on this branch, which main has since given to SHEET (81-84 are Sheet, Chart, text input and software floating point). It is renumbered to 85 here: the section and its 24 headings, its 13 in-section cross-references, and every citation in apps/tank/, tests/tank.py, tests/suite.py, docs/GFX-FSX-PLAN.md, the Makefile rule and four references from SPEC.md 17 and 53. Sheet's own 81.x citations are untouched, and so are the two `db` rows in apps/tank/ that merely contain the number 81.

PR #116's REVIEW FIXES, CARRIED ACROSS

All nine commits from elendilon-pr are present. Seven applied to code this branch had not moved and were taken verbatim. Two needed hand-merging, because both sides had changed the same routine:

- kernel/vga12.inc. Main moved gfx_blitp's [vid_mono] test to AFTER GFXDENTER, because it is a question about a display and asked above the hook it answers for whichever display was last active. This branch had refined the same block's clip refusal so that 11.3.3's damage cull is not treated as a clip region (SPEC.md 5.4.3.3) - refusing on it cost Paint its four planes for the session. Both are kept: the pre-hook block is the cull-aware clip test with no [vid_mono] in it, and main's post-hook test stands as written.
- Makefile, $(BUILD)/boothd.bin. Both sides fixed this rule for different halves of one bug. Main made the extraction its own step (so a refusal fails here with a traceback instead of reaching nasm as an empty string) and gave os88sym $OS88_BUILD, so a BUILD= sub-make's map is checked against its own kernel. This branch passed $(VIDDEF) through to kernsize, which is the half that was silent: kernsize re-assembles to measure, so --json with no defines described the SHIPPED kernel and a knob build got a VBR pointing its blob at the wrong heap floor (SPEC.md 52.10.2.1). kernsize now takes both --build $(BUILD) and the knobs; it needs each.

One fix was checked and deliberately left as main wrote it: tests/textsites.txt's kernel/wm.inc reason. Main corrected it to say the composed title bar is a KNOB, and that is true on this branch too - BANDCOMP is gated on BAND in all three trees - so the corrected reason stands.

MAIN'S CLAUDE.md is taken wholesale. It does not carry this fork's "Working in this fork" section; that is main's call and the file is theirs.

SIZE

kernsize[big]: sections   text 57,149 +0  bss 5,955 +0  cold 40,784 +0  lowbss 8,982 +0  vgabuf 848 +0  ovl 3,969 +0   (sum +0)
kernsize[big]: rungs      image 63,488 +0 (384 left, was 384)   cold 40,960 +0 (176 left, was 176)   low 10,240 +0 (234 left, was 234)   vgabuf 1,024 +0 (176 left, was 176)
kernsize[big]: accrued    image 128/512 (25%)   cold 336/512 (65%)   low 278/512 (54%)   vgabuf 336/512 (65%)   - spent into the current rung and NOT YET BILLED
kernsize[big]: footprint  KERN_SIZE 120,320 of KERN_BUDGET 129,536 -> 9,216 spare (18 steps), was 9,216  [+0]
kernsize[big]: boot       KERN_SIZE 120,320 of 189,952 -> 69,632 before it cannot BOOT on 196KB (guard 5)
kernsize[big]: segment    .text+.bss 63,104 of KERN_CODE_MAX 65,536 -> 2,432 left
kernsize[big]: ladder     KERNEL 0x0060  COLD 0x0fe0  FAT 0x19e0  LOW 0x1b00  VGABUF 0x1d80  HEAP 0x1dc0 = 119.0 KB   (heap KB = int 12h - 119.0)
kernsize[big]: unchanged

kernsize[small]: sections   text 48,415 +1,051  bss 5,442 +7  cold 37,870 +371  lowbss 8,470 +0  vgabuf 0 +0  ovl 3,886 +60   (sum +1,489)
kernsize[small]: rungs      image 54,272 +1,024 (415 left, was 449)   cold 37,888 +0 (18 left, was 389)   low 9,728 +0 (234 left, was 234)   vgabuf 0 +0 (0 left, was 0)
kernsize[small]: accrued    image 97/512 (18%)   cold 494/512 (96%)   low 278/512 (54%)   vgabuf 0/512 (0%)   - spent into the current rung and NOT YET BILLED
kernsize[small]: footprint  KERN_SIZE 106,496 of KERN_BUDGET 107,520 -> 1,024 spare (2 steps), was 2,048  [+1,024]
kernsize[small]: boot       KERN_SIZE 106,496 of 120,320 -> 13,824 before it cannot BOOT on 128KB (guard 5)
kernsize[small]: segment    .text+.bss 53,857 of KERN_CODE_MAX 65,536 -> 11,679 left
kernsize[small]: ladder     KERNEL 0x0060  COLD 0x0da0  FAT 0x16e0  LOW 0x1800  VGABUF 0x1a60  HEAP 0x1a60 = 105.5 KB   (heap KB = int 12h - 105.5)
kernsize[small]: *** the image rung CROSSED: 104 -> 106 steps of 512 - 1,024 bytes of every machine's RAM, gone ***

docs/KERNEL-MEMORY.md is blessed for both variants, and the hand-written "Where it goes" table and the kern_small paragraph are re-derived from these same sections and rungs lines, as the section above them instructs.

TESTING

make: green - checkdocs 1699 headings, 151 slots, transparent-text registry 59 sites in 20 files, 0 problems.

os88test fast: 28 passed, 0 failed, 0 skipped in 4.3s (budget 30s). os88test FULL: 35 passed, 0 failed, NO SKIPS in 162.6s (budget 600s) - the whole gate, with a C toolchain, MartyPC and QEMU all present, so no row was excused. buildmatrix (71.9s) is every knob kernel and kern_small assembling; bootsmoke boots this kernel on a cycle-accurate 4.77MHz 8088; weavesmoke runs main's WEAVE against it on both 1bpp adapters.
Soak, for the rows this round moved or added: tank 41.6s, blitcut 262.9s, dispzoom 53.4s, sizesnap 19.1s, wmartifact 211.9s, tmdmg 59.3s - all green.

kernsize reports the blessed baseline unchanged after the commit.

Claude-Session: https://claude.ai/code/session_01R14MXKyKvhZFmuF9nxhmzQ


jggonz:
Reviewed this against `main`, merged `main` in, and pushed two commits — `19d75e9` (the merge) and `64f66de` (all the fixes below). **The PR is mergeable.**

The merge was one commit — main's #122, which added `vm/xt-weave` and `vm/386-weave`. No conflicts, no duplicate `§` numbers on either side, and nothing of main's was lost: I compared every Makefile target, every `Row(` in `tests/suite.py`, every SPEC heading, every `%error` in `kernel.asm` and every top-level label in the kernel and driver files against the merge base and the difference sets are empty. Both of PR #116's hand-merges survived with both halves. The §81 → §85 renumbering is complete and Sheet's own citations are untouched, and the API table grew by append only — no published slot moved.

Ten agents reviewed the merged tree unit by unit and an adversary tried to refute every finding. **15 of 31 findings did not survive** and are not below — including two I would have bet on: `gfx_blit4`'s `[gfx_blit_hk]` ordering (identical in main, so not yours) and `CDGRAY` breaking the 1bpp promise (`gfx_pen_dis` already rounds it, and the stipple carries §47's signal). What follows is what stood up.

---

## The one that stops a merge

**Paint's tool palette goes blank after a canvas resize.** Change the width, click Apply, and the whole left column — tools, W/H boxes, Apply — is white. It stays blank until something forces a full repaint; dragging the window brings it back, which is exactly the kind of "it fixed itself" that never gets reported properly.

`pt_fsbed` lays the tool column's white bed on `pt_dmg_hit` — *does this paint owe this rect* — while §42.19.3's new gate runs `pt_draw_pal` on `pt_dmg_anch` — *did the anchored rect **move***. A resize that leaves the origin where it is answers yes to the first and no to the second, so the bed is erased and nothing draws on it. The bed asks the same question as the furniture standing on it now.

I photographed this before the review found it by reading, which is the only reason I'm confident it is the resize and not my mouse.

## Memory

- **A long stroke could corrupt the "picture has changed" flag, so Paint would close without asking to save.** `pt_bank`'s full path is documented as "spend exactly ONE sample and drop it", but with `[pt_bnkd]` still 0 the whole budget pays the press dab, `pt_flush_n` returns with `[pt_bnkh]` at 0, and `pt_bnk_compact` — which drops nothing when the head has not moved — leaves the bank full. The store after it is unconditional, so it wrote four bytes **past** `pt_bnk`, over `[pt_ikeep]`, `[pt_dirty]` and `[pt_blankc]`; the release replay then read them back as a chord endpoint. It is a loop now, and two turns is the ceiling. (§42.8.8.1 says so too.)
- **The mouse pointer left a solid black or white square behind it** on any window that promised `OSAPI_SAVEU_1BPP`. `gfx_save`/`gfx_restore` call `cur_unlazy` at the *top*, above their own disarm, so the arrow's four-plane save-under was restored through the one-plane arm's Map Mask 0Fh. `vga12.inc`'s own comment said this could not happen "because gfx_save and gfx_restore put them back" — that is the half that is not true. `cur_saveu`/`cur_restoreu` bank the pair now, at the one point both cursor paths pass through (§11.96.17).
- **A free-space figure read out of memory something else had taken.** `dsk_free_clus` banks `[dsk_fatseg]` in ES for a whole walk. That was safe while the window was `MEM_K_FATW` and pinned; as `MEM_P_FATW` it is shed by any other task's ordinary claim, and `mem_fatw_dirty` does not cover a count. `sch_lock` over both banked runs — released between FAT sectors, because the next one does disk I/O.
- **The hard disk's listing claim was declared movable one call too early.** `OSAPI_MEM_MOVABLE` ran before `OSAPI_VOL_ADD` wrote the base down, so a compaction in that window corrected `HDV_LSEG` and left the *stale* DX to go into the kernel's own `DV_SEG`. The declaration moved below the `VOL_ADD`, with the volume index reloaded because `hd_lst_reloc`'s address goes through AX.
- **TANK ATTACK's ninth game refuses to start, permanently.** `tk_r_setup` claims 16KB of shadow at every Play and never gives it back — on CGA and Hercules only, which are the machines with the least heap to lose. Once per instance now; the "the kernel frees it with the instance" comment is true again.

## Input and the ISR rules

- **A serial mouse may not be detected at boot, and the machine just has no pointer.** `spl_stepq` raises `[spl_quiet]` around `mouse_init`'s identify window because a 28.7 ms spinner frame loses three bytes of a 1200-baud burst an 8250 holds one at a time — and `sch_isr`, which is the handler that actually fires there (and deliberately does *not* `sti`), drew straight through the flag with interrupts off. The flag was a `.boot2` byte read `cs:`; both IRQ0 handlers are kernel `.text`, where `cs:` names the kernel segment. It lives in `viddet.inc` beside `[spl_live]` and `[spl_busy]` now.
- **The boot spinner could paint backwards.** `spl_isr` reaches `spl_spin`'s `rep movsb` with no `cld`, so DF belongs to whatever it interrupted — during the load, a BIOS `int 13h`. `sched.inc` carries the same line above its own in-ISR blit.

## Redraw

- **The pencil drew every stroke twice on Hercules and CGA.** `pt_segdo` stored 0 to `[pt_noscr]` rather than restoring it. That was right while `pt_stroke` was its only caller; §42.8.8's replay is a second one and arrives with the flag *set* — the live pass already put the ink on the glass. The clear also leaked past the routine, so `.perpix` chords after it drew per-pixel to the screen too. This is PERFORMANCE.md's double-draw flash, on the default tool, on the machine the fast path exists for.
- **Holding Backspace on an empty Rename box repainted the whole listing at every key repeat.** `fm_editkey`'s empty-buffer exit answered `.out` — "the caller owes the whole window" — which was free until this PR connected that answer. It answers `.nochg` now, like the four refusal sites beside it.
- **The Tank attract screen could draw its logo over whatever had come to the front.** `tk_at_step` asks `OSAPI_WM_OBSCURED` above the gfx lock, on a worker; `wm_obscured`'s own header says it is only trustworthy under the lock. Lock first, then ask.
- **The Browser lost its save-under for the session after a bad URL.** `br_go`'s error tail settles to `BN_ERR` without re-stating the promise the `BN_OPEN` transition withdrew — and the fetch was aborted, so neither of the two settles that own the set ever runs. One `br_promise` call (§71.11).

## Correctness

- **A window dropped on the second monitor lost its 8-pixel alignment for the session**, and with it `font_run`'s single-store fast path on every text row. `wm_snap_ax` asks `wm_zoom_xmax`, which asks `wm_disp_of`, which reads the window **record** — and `ui_drag` calls the snapper before it stores, because its own "did it move" test still needs the old x. The candidate goes into the record for the length of that one call and comes straight back out.
- **`fsx_page` turned on the Hercules second-page decode unconditionally, including for page 0.** `vid_setmode` leaves that bit clear and says why at the write: the page is at B8000, where a CGA in the same machine lives, so enabling it puts two cards on the bus for the same addresses. Refused on a dual-card machine (`vid_dual_ok`, not `[vid_ndisp]` — `vid_fsx_enter` has already forced that to 1), and page 0 puts the register back. A lone-Hercules XT still gets the page.
- **Answering "Save" on Paint's close question could overwrite the original of a cropped picture.** File → Save fences on `[pt_trunc]` and says so; the close question offered the same Save with no such test, and `.close` follows straight on — no second chance. It raises Save As instead, which already sets `[pt_qclose]` so the close still finishes.
- **A height-only window drag welded the alignment padding onto the document.** §42.20's entitlement test was read as one question when it is two: with the width test passing and the height changed, the combined `je` fails and the width comes back from the content — 466 becomes 472, six white columns in the file, and stable afterwards. The width answer is banked per axis now.
- **A drag across an alert left a button drawn pressed.** `os88ui_adrag` was written as "`os88ui_aclick` without the arm", so it pressed whatever the pointer crossed — and a press that started on the alert's *message* arms nothing, so the release took `or ax,ax / jz .out` and never put it back up. It reads the arm through `os88ui_armed`'s peek, and the release un-draws first and unconditionally (§13.8.2's third consequence).
- **Cyclone's droid drew itself in the wrong place after a level change** onto a narrower web — `cy_clearboard` clears neither `[cy_pw_droid]` nor `[cy_dlane]`, and rows 12..15 of the vertex tables still hold the previous shape's coordinates. Clamped in `cy_home_ck`, which exists for exactly this timing.
- **cword told you your RTF was not an RTF file** when the read had actually failed, and said nothing at all when the folder could not be listed. `lo` is the dialog's size hint and the association path has none, so the guard written against it was dead there; `os88_ferr()` is the discriminator `os88.h` documents for this.

## The documents and the gates

Every one of these is the code's own contract disagreeing with the code.

- **WEAVE-SPEC §7.1.1** — the prose above says the content of a flush window is `w`, and the derivation below still said `CW = floor(([vid_w]-1)/8)` with **79×17 / 89×35 / 79×52**. `weavesim` and `weavesmoke` already said 80/90/80. §12 says the 8086 must reproduce those grids exactly, so this is the number the wave-3 implementer would have been handed.
- **`tests/sizesnap.py`** modelled the superseded `W_W-1` rule and would have reported a correct build OFF GRID.
- **SPEC.md §15.1 and §2.7** still said `MIN_RAM_KB` is 128. It is per configuration now — 196 for big — and that is a real change of what the shipped system claims.
- **SPEC.md §28's Task Manager tag table** still listed `MEM_K_FATW`, three lines above the `MEM_P_WSAVE` + slot row that is already in the shape it needs.
- **`tools/dsegaudit.py`** is a 0/1 gate and nothing ran it — not `make`, not the suite, and `t_registry` cannot see it because its walk is over `tests/`. It is a fast row now. **`NOFLUSHR`, `FATWGATE` and `FDDSLOW`** were in `$(KNOBS)` with no build-matrix row and no test, so nothing was keeping them assembling; all three are in `t_buildmatrix`'s roster.
- **`kern_small`'s kernsize baseline was never blessed**, so every `make small` printed a rung crossing against main's figures. Both variants are blessed and the two prose figures in KERNEL-MEMORY.md re-derived.

## Flagged, not fixed

- **`[pt_ikeep]` is read by `pt_wipe` and written nowhere**, so §42.18's band-preserving wipe is inert and `pt_iresize`'s stated precondition is false. Harmless today — `pt_resize` takes `pt_cvwipe` on the carrying path and `pt_wipe` only on `.allnew`, where emptying the table is right. I left it because the fix is a question only you can answer: was the mechanism meant to exist, or is the flag the leftover? Either way it wants deciding rather than patching.
- **`tests/blitcut.py` never runs the BELOW layout**, so `.cutv` — the only branch with the source-stride multiply — is at zero coverage. I read it line by line and found no defect (the `mul`/`imul` low-word argument holds, the stack is balanced on the odd-cut refusal), so this is coverage rather than a bug. `dispcp.set_mode` already takes `"below"`.
- **`gfx_blit4` clears `[gfx_blit_hk]` after its two zero-size guards**, so a zero-width blit can run `GFXDLEAVE` for a hook it never took and underflow `[gfx_dnest]`. This is **main's**, byte for byte — your `gfx_blitp` comment describes exactly this class and puts its own clear in the right place. Not yours to fix in this PR, but it is real and I have raised it separately.

## What I checked and cleared

`make`, `make KERN_SMALL=1` and `make small` all clean; `test-fast` 29/29; all seven images fsck clean; boots to a desktop on VGA and on CGA. **TANK ATTACK** launches, the attract band rolls its six rows in order, F enters Mode X and draws the horizon, the tank and the radar, and Esc brings the desktop back intact — the bracket's exit path is sound. **Paint** draws a multi-segment stroke with no gaps, saves, and the BMP header is exactly 448×280, 4bpp, 62,838 bytes at offset 118 with the disk clean afterwards. The PS/2 probe passes its own gate (`p2st 9, port 04, pointer exact, keyboard intact`). The C toolchain builds `cword` through its four-rule gate.

**What I could not run: MartyPC is not installed here**, so `tank`, `blitcut`, `dispzoom`, `sizesnap`, `tmdmg`, `spantest` and the Paint rows — every gate this PR adds — are unrun by me. Your own run of them is the evidence for those.

## One number worth your attention

`kern_small` is at **107,008 of its 107,520-byte budget: 512 bytes, one step.**

The image rung (1,024 bytes) is this PR's own. The cold rung (512) is my `sch_lock` fix, and it crossed by **one byte** — the PR had left 18 bytes in that rung and the fix is 19. I have not touched a guard; both baselines are blessed and `make small` reports `unchanged` again. But small has one step left, and the next 512 bytes it grows is a conversation rather than a build fix. If you would rather have that step back than the FAT12 half of the guard, say so — the FAT16 half is the one that matters most and it is half the bytes.

Nothing above is a criticism of the work. This is a large, careful PR and most of what the review threw at it bounced off.

PR #123, opened 2026-08-31
kernel: OSAPI_JSLOT's jump must be STRICT near - main does not assemble

Koriban:
## What

`main` does not assemble. One word fixes it.

```
kernel/kernel.asm:3317: error: os8088 API jump table must be exactly 153 8-byte slots
```

## Why

The table has exactly **153 entries** and both slot macros are eight bytes by
construction, so the count is right. The **length** is wrong: 1,223 bytes — one
short, and not even a multiple of eight.

`OSAPI_JSLOT` is `jmp near %1` plus a fixed `times 5 db 0`, and its own comment
says why that adds to eight: *"E9 rel16 = 3 bytes"*. NASM shortens the jump to
`EB rel8` anyway when the target is within 127 bytes, and `near` alone does not
stop it — `strict` does. `api_icon_draw` came within range in #116, so its slot
assembled as **seven** bytes. The listing names it outright:

```
2292 000004B8 EB33            <1>  jmp near %1
2293 000004BA 00<rep 5h>      <1>  times 5 db 0
```

The padding is a fixed count, so it cannot absorb the difference. Every slot
after the shortened one slides down a byte and answers at the wrong address —
`gfx_blitp` at `0x04C0` would have been reached at `0x04BF`. That is exactly
what the length assertion is there to catch, and it caught it.

## Scope

Bisected to **f3a41d1 (#116)**; `4f415e7` — the previous `main` — still builds.
Reproduced at every commit from #116 to `37db5c2`, on a clean checkout with
none of my own work in the tree.

## SPEC

No section moves. This restores the behaviour §20.1's ABI already specifies —
a fixed 8-byte slot at a fixed offset — rather than changing it.

## Verification

```
git checkout 37db5c2 && make build/kernel-full.bin     # fails, as above
git checkout <this branch> && make -j4                 # builds
python3 tools/os88test.py full                         # 25 passed, 0 failed
```

Measured by replacing the assertion with the `times` idiom, which reports the
delta rather than just failing: `TIMES value -1 is negative`, i.e. 1,223 bytes
against 153 x 8 = 1,224.

No screenshots: there is nothing visual here, and before the fix nothing boots
to photograph.

### DIFF
--- a/kernel/kernel.asm
+++ b/kernel/kernel.asm
@@ -2289,7 +2289,14 @@ dbg_reg_at:                     ; 0060:000E - THE DEBUG REGISTRY (SPEC.md 57)
 %endmacro
 
 %macro OSAPI_JSLOT 1                ; a cell that defers to a longer stub
-    jmp near %1                     ; E9 rel16 = 3 bytes
+    ; STRICT, and the whole slot depends on it. Without it NASM shortens the
+    ; jump to EB rel8 whenever the target is within 127 bytes - which
+    ; api_icon_draw became - and the slot emits 2 + 5 = SEVEN bytes. The
+    ; padding here is a fixed `times 5`, so it cannot absorb the difference:
+    ; every slot after the shortened one slides down a byte and answers at the
+    ; wrong address. That is the whole reason the length assertion below
+    ; exists, and it is what caught this.
+    jmp strict near %1              ; E9 rel16 = 3 bytes, always
     times 5 db 0
 %endmacro
 


jggonz (approved):
Verified independently and approving — the fix is correct, and it's exactly the right one. Nothing needed changing, so I pushed nothing to your branch.

What I ran, on your head (`e3b68f9`) with current `main` (`f6e00ac`, #121) merged in locally:

- `make` — green, fast tier 29/29.
- `python3 tools/os88test.py full` — **32 passed, 0 failed** (4 skips that need MartyPC).
- A byte-for-byte comparison of `kernel-full.bin` built with and without the fix: the only differing bytes are the commit-count build number (§14.2). The API table is identical, so the change costs nothing on the current toolchain — it just makes the encoding §20.1's ABI already requires unconditional.

One nuance worth recording here for whoever reads this later: **the failure is NASM-version-dependent.** On NASM 3.02 (what this side runs), `jmp near` already emits `E9 rel16` and pre-fix `main` assembles fine; your listing's `EB 33` shows your NASM's optimizer shortens non-strict `near`. That makes the title's "main does not assemble" true on some toolchains and not others — and makes `strict` the right fix regardless, since a table whose byte layout depends on the assembler version is exactly what a fixed-slot ABI cannot tolerate. Your analysis of the mechanism (fixed `times 5` padding that can't absorb the shortened jump, every later slot answering one byte early) checked out line for line, and `OSAPI_SLOT` is confirmed safe by construction since `call rel16` has no short form.

Nice catch, and thank you for the bisect and the listing excerpt — they made this a fast review.

PR #138, opened 2026-09-01
EGA 640x350 support (VID_EGA) - closes #136

### LINKED ISSUE #136: Add EGA support
My 5160 XT machine has an IBM EGA adapter. It's maximal graphics mode is 640x350 with 16 programmable colors from 64.

It can still work with VGA 640x480 mode, but since EGA has only 350 pixels, the bottom part of the image is inaccessible.

Image

I can use a CGA mode, but then it's a waste of a card.

Image

pentagram666:
## What & why

Closes #136 — a real IBM 5160 with a genuine IBM EGA and a 350-line monitor.
`vid_detect` classes every EGA as `VID_VGA`, so the kernel sets mode 12h and
drives a 640×480 desktop into a framebuffer the card shows 350 lines of; the
bottom 130 rows (part of the dock included) are off the tube. CGA works and
wastes the card.

Mode 10h (640×350×16) is mode 12h's framebuffer at a shorter height —
planar, four planes, `A000:0000`, 80-byte stride, not banked. So `VID_EGA` (3)
is **not a fourth renderer**: `[vid_mono]` stays 0, every `vga12.inc`
primitive is reused unchanged, and `gfx_rowbase`/`gfx_nextrow` reduce to
`y*80` / `add di,80` with `bmask=bshift=0` exactly as on VGA.

## SPEC.md sections

- **New §39.24** — the `VID_EGA` contract and the full list of what differs
  from VGA.
- §39 adapter table (fourth row), §39.1 (the detection branch + two scope
  cuts), §39.6 / §39.23 (mode 10h + the 350-row clear), §39.7 (`wm_fit` on a
  350-tall screen), §39.8 (`OSAPI_VIDEO` DL can be 3 — branch on bpp not
  kind), §39.9 (`VIDEO=ega` and `make xt-ega`).
- New design record **`docs/EGA-PLAN.md`** — why EGA is a kind and not a flag
  on `VID_VGA`, the rejected alternatives, and what is deferred.

## The change

`vid_tab` row + a `vid_detect` branch + one `vid_setmode` arm, then the
handful of sites the fourth kind reaches. Files:
`kernel/{viddet,vidsel,vga12,wm,ctrl,apps,fsx}.inc`, `apps/os88api.inc`,
`apps/{missile,tank,tracker,cyclone}` + `apps/cc/os88.h`, `Makefile`,
`README.md`, `CLAUDE.md`, `docs/TESTING.md`, `tests/unit/t_buildmatrix.py`,
`vm/xt-ega/86box.cfg`.

## The Colour system theme now runs on EGA (commit `b7d227c`)

Added as a separate commit after review. `thm_set`'s eligibility check for
`THM_COLOR` was `cmp [vid_kind], VID_VGA` — an **identity test standing in for
a colour capability**, written while VGA was the only colour adapter.
`VID_EGA` is a four-plane 16-colour adapter that is not a VGA, so the proxy
broke and refused Colour on the one machine class §76.12 is patterned on.
The theme needs nothing an EGA lacks: it is six palette *indices*
(0/1/3/7/8/15) copied into an 8-byte block; no DAC, no Attribute Controller,
no palette programming anywhere. So `thm_set` and `cp_thm_colgrey` now gate
on `[vid_mono] == 0` (a four-plane adapter; `= 1` still refuses, §39.4).
**Zero kernel bytes** — both `cmp` forms are 5-byte `80 3E addr imm8`,
`kernel.bin` byte-identical. SPEC.md §76.12 / §76.12.1 / §76.12.4 updated.

Still deferred: `VID_EGA` in a `KERN_BIG` per-display pair, an EGA on a
monochrome display (§39.1's scope cut — `int 10h AH=12h` `BH=1` falls through
now), and a 64 KB EGA.

## Open question for the maintainer

Should `OSAPI_FSX_CAPS` answer a depth in `DH`? It answers a *per-display*
kind and no depth, so a package genuinely cannot follow the SDK's "branch on
bpp, not kind" rule from that slot — which is why the four packages below had
to be patched by hand. `DH` is unused in the slot's contract, so it is purely
additive. Not done here — it is a published-slot contract change and a
separate decision from this PR.

## How verified

- **`make` (default and `small`)** — clean, `-w+error` on. Both the
  auto-detect kernel and a `VIDEO=ega` kernel assemble with zero warnings.
- **`make test-full` — 33 / 33**, `buildmatrix` (the knob kernels incl.
  `kern_small`) included, with a new `video-ega` row. `os88disk --verify` on
  all six images.
- **`kernsize`** — `.text` +168 bytes total, **no rung crossed**, `KERN_SIZE`
  unchanged (120,320 / 129,536), both guards untouched, on both variants.
- **86Box `ibmxt` + IBM EGA** (two floppies, SB 2.0) — the §39.1 detection
  branch and the mode 10h set run on 86Box's period EGA BIOS; comes up to the
  640×350 desktop. This is `vm/xt-ega/86box.cfg`'s machine.
- **Real period hardware** — booted from the **360 KB** floppy pair on a
  *TOP Super Turbo III* (AMD 8088-1 @ 10 MHz, 8087, 640 KB) with an **ATI EGA
  Wonder**: detection on a real (clone) EGA BIOS, `int 10h AX=0010h` sets
  mode 10h, 640×350 desktop.
- **Driven on the glass at 640×350** — desktop / menu bar / drive column
  (A:, B:) / dock all inside 350 rows, nothing below; menus open; a Disk
  window opens and fits; About reads `EGA - 640x350 - 16 colors`; the Display
  page lists Cga + Ega with Ega selected; a full **EGA → CGA → EGA primary
  round trip** (runs `vid_blank_kind`'s EGA arm, sets mode 10h at run time);
  and **Cyclone launches in colour** (fix #1, below, photographed).
- **`VIDEO=cga` boot** — the 1bpp path is undisturbed by the `vid_apply`
  re-keying.

**Still open:** a boot on the **IBM 5160 + genuine IBM EGA** from #136 —
@Bs0Dd to confirm the colours and layout on IBM's own card (86Box's IBM-EGA
emulation already passes). Three questions for them are in `EGA-PLAN.md §8`.

## Maintainer review — folded into the branch (`e0355b4`)

Second pass on the edges the fourth kind reaches:

1. **Four packages rendered monochrome on the 16-colour card.** Missile
   Command, Tank, Tracker and Cyclone decide "is this colour?" with
   `cmp dl, VID_VGA` against `OSAPI_FSX_CAPS`'s kind — `DL = 3` sent every one
   down the mono arm. Each gained the second compare.
2. **The AC blank arm** (`vid_blank_kind`/`vid_unblank_kind` EGA path) is one
   `vid_ac_pas` now: a `pushf`/`cli` around "reset the flip-flop, write the
   port" (a task switch in that gap left the toggle mis-phased), and a second
   3DA read afterwards so 3C0h is not left expecting *data* with palette 0
   selected. +14 bytes, no rung crossed.
3. `VGADIRTY=1`'s dirty-fill now takes `EGA_H` like the clear it mirrors.
4. `t_buildmatrix.py` gains a `video-ega` row (it was the only `VIDEO=` value
   nothing kept assembling).
5. SPEC.md §39.9 said `make test VID_FORCE=4` — not a make variable; it is
   `make test VIDEO=ega`. And the framebuffer does come back 640×350 under
   QEMU, not 480. Both corrected.
6. `make xt-ega` added to README (×3), CLAUDE.md and TESTING.md alongside the
   other 86Box targets; `apps/cc/os88.h` gains `OS88_VID_EGA`; two stale
   comments fixed.


jggonz:
Reviewed and merged nothing — this branch is cut from `main`'s current tip, so there was no catching-up to do. I pushed **one commit** (`e0355b4`) to `ega-support` with the fixes below.

First, the headline: **the design is right.** `VID_EGA` really is `VID_VGA`'s planar path at a shorter height, and I went looking for the geometry sites that would need the live `[vid_ch]` and didn't have it — the usual way a "one binary, N adapters" change goes wrong. I couldn't find one. `vga12.inc` was already reading `[vid_ch]`/`[vid_chm1]` nearly everywhere, which is exactly why only two constants needed changing; that's a good sign about the change, not a suspicious one.

I also re-ran the tier you couldn't run on Windows. **`make test-full` is 33/33 at your HEAD**, `buildmatrix` included, and your kernel-size figures are exact: `.text` +150/+138, no rung crossed, `KERN_SIZE` unchanged.

---

## What I fixed

**1. Four games rendered in black-and-white on your 16-colour card.**

This is the one that would have stung. Boot the EGA build, launch Missile Command, Tank, Tracker or Cyclone, and every one of them comes up monochrome — on the very card the whole PR exists to light up. You'd reasonably conclude the EGA support wasn't really working.

The mechanism: those four ask `OSAPI_FSX_CAPS` which display they're on and then do `cmp dl, VID_VGA` to mean "is this colour?". Perfectly true when there were three kinds; with `VID_EGA` = 3 they all fall to the mono arm. Each now has the second compare, in the shape your own `wm_display` change already uses.

Worth saying plainly: **this wasn't a thing those packages could have got right.** Your SDK note says "branch on bpp, not kind", and it's the correct rule — but `OSAPI_FSX_CAPS` answers a *per-display* kind and no depth, and on a two-card machine `OSAPI_VIDEO`'s depth is about the *primary*, which is the wrong monitor. So the kind was the only thing they could ask. See the open question at the bottom.

**2. The Attribute Controller blank could have left the desktop the wrong colour.**

`vid_blank_kind`'s new EGA arm reaches the AC's video-enable bit correctly, but it's missing the two things `vid_seq1_blank` one screen above it spells out for the VGA's arm:

- **No critical section.** "Reset the flip-flop, then write the port" is exactly the reachable gap that routine's header documents — a task switch landing between the two leaves the toggle in the wrong phase.
- **The flip-flop is left in the data phase.** 3C0h is one port for both index and data, and which one a write lands in is a toggle. Writing the index and stopping leaves it expecting *data*, with palette register 0 selected — so the next code to write an index there writes it into palette 0 instead, and the desktop's black quietly becomes some other colour. Two bytes to put it back.

Both arms are now one documented `vid_ac_pas` that does the `pushf`/`cli` and re-reads 3DA afterwards. Net cost across all my changes: **14 bytes, no rung crossed.**

**3. `VGADIRTY=1`'s dirty-fill still wrote 480 rows.** You made the *clear* EGA-aware; the fill 48 lines above it, which exists to dirty exactly what that clear cleans, still used `SCREEN_H`. Test-only knob, so nothing shipped — but the two halves of one routine have to agree, and on a real 128KB EGA that build would write past the plane.

**4. `VIDEO=ega` was the only `VIDEO=` value nothing kept assembling.** `t_buildmatrix.py` has `video-cga` and `video-herc` rows; there's a `video-ega` one now. That file's own header is about exactly this rot.

**5. SPEC.md §39.9 told the reader to type a command that does nothing.** It said `make test VID_FORCE=4`. `VID_FORCE` isn't a make variable — the Makefile derives it from `VIDEO=`, and only `VIDEO=` is in the stamp — so that command sets a variable nothing reads, leaves `kernel.bin` up to date, and boots the *previous* configuration. Which reads exactly like the feature being broken. It's `make test VIDEO=ega`.

Same paragraph said mode 12h is still set under QEMU so the framebuffer stays 480 tall. It isn't: your EGA arm asks for `0010h` and QEMU's VGA honours it — **my screendumps came back 640×350.** The knob tests more than you gave it credit for. Corrected both.

**6. `make xt-ega` was documented nowhere** — missing from README (three places), CLAUDE.md's machine list and TESTING.md's two lists, where every other 86Box target appears. Added, plus `ega` in CLAUDE.md's `VIDEO=` row.

**7. Two small ones.** `apps/cc/os88.h` never learned `OS88_VID_EGA`, so the C and asm SDKs disagreed about how many kinds exist (nothing in tree needed it — `apps/weave/wcanv.c` is the only C package that asks and it correctly tests `bpp == 4`). And two comments that stopped being true: `cp_vid_slot`'s "0..2" and `vid_probe_avail`'s header claiming `[vid_kind]` is `VID_VGA` on a machine with an EGA BIOS.

## What I checked and left alone

- **The `cmp … VID_VGA` sweep**, all of it. Beyond the four above, the remaining ones are right: Paint, Frotz and Tracker's *depth* tests all use `cmp dh, 1`; `os88ui.inc`'s `UI_ISMONO` reads `[vid_mono]`; the browser tests *for* CGA specifically; `vid_text` falls through to mode 3 for EGA exactly as your §39.6 text claims.
- **`wm_pref_take`'s clamp** — I checked it's the single site that indexes the package's table by kind, so clamping there is complete.
- **`vid_probe_avail`'s ordering** — `[vid_mono]` really is published by `vid_apply` at kmain:4157 before the probe at :4172, so re-keying that test was safe.
- **The three-row Display list.** I convinced myself a fourth row could never be drawn into a band laid out for three: `vid_avail` starts as `1 << vid_kind`, so the VGA and EGA bits are mutually exclusive and three is still the ceiling. Left the geometry alone; the comment now says so.
- I also chased an apparent "can't re-select EGA in Control Panel" bug for a while. It was my own `--screen` mismatch on a CGA-mode screen. Not a defect — mentioning it so nobody else loses the same twenty minutes.

## Flagged, not fixed

**The Colour theme is greyed on EGA.** Your `EGA-PLAN.md §7` defers it deliberately and the greying is correct and honest — I photographed it. But it's worth saying out loud that this is the most visible thing @Bs0Dd will notice: a 16-colour card that gets Bright/Dark only. Might be worth a line in the PR body so it lands as a known deferral rather than a surprise.

**Should `OSAPI_FSX_CAPS` answer a depth in `DH`?** That's the root fix for #1 — it would make "branch on bpp, not kind" actually followable per-display, which today it isn't. `DH` is unused in that slot's contract so it's purely additive. I did *not* do it: it's a published-slot contract change, it costs kernel bytes, and it's your call and @jggonz's, not a review repair. The four packages are patched either way.

## How it was verified

`make` and `make small` clean under `-w+error`; `os88test full` 33/33 with the new row; `kernsize` on both variants (+14 bytes on your figure, no rung crossed, `KERN_SIZE` and both guards untouched); `os88disk --verify` on all six images; a `VIDEO=cga` boot to confirm the 1bpp path is undisturbed by the `vid_apply` re-keying.

And driven on the glass at 640×350: the desktop, menu bar, drive column and dock all inside 350 rows with nothing below; menus open; a Disk window opens and fits; About reads `EGA - 640x350 - 16 colors`; the Display page lists Cga and Ega with Ega selected; Colour correctly greyed on the Theme page; a full **EGA → CGA → EGA primary round trip**, which is what runs your `vid_blank_kind` EGA arm and sets mode 10h at run time rather than from the splash — clean repaint both ways; and **Cyclone launched in colour**, which is fix #1 photographed.

Nice piece of work — the "not a fourth renderer" framing is what made it small, and the SPEC-before-code ordering made it reviewable. Everything above is a second pair of eyes on the edges, not a verdict on the change.

jggonz:
## Correction, and a second round of fixes (`90357cc`)

**I need to walk back the headline of my last comment.** I said "the design is right" and that I couldn't find a site that needed the live `[vid_ch]` and didn't have it. The second half stands — the geometry really does look right, and I still can't find a bad clip. The first half was too generous: **this PR does introduce one regression**, and I missed it because my verification was structurally unable to see it.

Four more findings. Fixed and pushed.

---

### 1. `vid_dual_ok` stopped being invariant, and `vid_disp_init` was written on the assumption that it couldn't

This is the real one.

Before this branch, `vid_dual_ok` read **nothing but `[vid_avail]`** — which `vidsel.inc`'s own header calls "fixed for the life of the machine". So the predicate was constant across a boot, a refusal could never follow a non-refusal, and `vid_disp_init`'s `jc .out` returning without doing anything was safe by construction.

`VID_EGA` adds `[vid_kind]` to that predicate, and `vid_switch` **moves** `[vid_kind]` at run time (`vidsel.inc:1000`).

So on a `KERN_BIG` machine with an EGA *and* a Hercules — the config your own comment at the refusal invites — this sequence:

1. Boot EGA. Dual refused, one display.
2. Display → Set Primary → **Cga**. Dual now *passes*. The Desktop row appears.
3. Turn on **Extend**. `[vid_ndisp]` = 2, display 1 placed and lit.
4. Display → Set Primary → **Ega**. `vid_disp_init` refuses — **and tears nothing down.**

`[vid_ndisp]` stays 2 with display 0's `vid_ctx` still describing the *CGA*, so the kernel paints a 640×350 desktop into B800 at a 640×200 stride. The EGA goes dark, pixels land on the CGA aperture, the Hercules keeps a frozen half-desktop, and windows can be dragged into a display region that no longer exists.

The refusal arm now checks `[vid_ndisp]` and, when something is live, collapses it through the SINGLE arm that already knows how — other card darked, count back to 1, desktop re-unioned. `[vid_dmode]` is **kept**, so the arrangement is impossible on this primary rather than forgotten, and switching back restores it.

**Not exercised.** It needs a second card; QEMU can't provide one and 86Box can't be driven. Reasoned, built, gate-green — and flagged rather than claimed. Worth a look on real hardware if anyone has an EGA + mono pair.

*(And a note on my own testing: I ran exactly that EGA → CGA → EGA round trip on the glass last round and reported it clean. It was clean — because with no Hercules present `vid_dual_ok` refused at every step for the other reason, so the bug was unreachable. A passing test on the wrong machine.)*

### 2. An EGA on a monochrome display was being sent to a mode it doesn't have

`int 10h AH=12h` — the very call that detects the card — also returns **`BH` = 0 for a colour attachment, 1 for monochrome**, and the branch threw it away.

That answer is binding, because an EGA's mode list follows the display its DIP switches name. With a 5153/5154 you get modes 0–6 and 10h. With a **5151 you get modes 7 and 0Fh and nothing else** — no mode 10h, and no mode 6 either. So the old branch set a mode the card doesn't have and then painted `A000` (a blank tube, not a degraded picture), and `vid_probe_avail` went on to offer that machine a CGA mode it also doesn't have — which, if picked, is a black screen with no way back, persisted to `SYSTEM.CFG` for the next boot too.

`BH = 1` now falls through to the equipment word. That isn't a *good* answer — it's an honest one. The right answer is a **mode 0Fh setup** (640×350 mono, planes 0 and 2, at `B000`), which is a second adapter setup rather than a flag on this one, and I've written that up in `EGA-PLAN.md §7`. SPEC §39.1 asserted the opposite in as many words — "the only planar mode it has" — and that's corrected.

### 3. The EGA blank was writing a bit the IBM EGA does not have

Last round I fixed the *shape* of that arm (critical section, flip-flop). I should have questioned the mechanism: **bit 5 of the AC index — Palette Address Source — is a VGA addition.** On the IBM EGA the index register is bits 4–0 with 7–5 unused, and SR01 bit 5 (Screen Off) is a VGA addition too. So the EGA has *neither* enable bit, and the arm as written blanks nothing at all on the one card it exists for.

What an EGA can do is show black. `vid_ac_pal` puts all sixteen palette registers down and mode 10h's own sixteen back. It keeps the property the other three arms have and `vid_blank`'s header states — CRTC untouched, both syncs still going out, the tube never re-acquiring — so a background task may still draw into a blanked EGA.

**And this one cost a round trip, which is the part worth reading.** My first version blanked beautifully and then *never came back*. Every index write carried a bare `0..15` — bit 5 clear. On the EGA that's a don't-care, but this same code runs on a VGA under `VIDEO=ega`, and there a clear bit 5 means "the host is loading the palette", which **disables video as a side effect of addressing it**. So the restoring pass put the colours back correctly into a tube that was still switched off. Every index write carries `0x20` now.

I only caught it because I compared against your HEAD: yours blanks and restores correctly under the harness, mine blanked and didn't. Nothing but the glass shows that one.

### 4. SPEC §53.4 was still describing the pre-PR world

Its mode table had three adapter columns and its mask list read "VGA 0x1EF, CGA 0x00F, HERC 0x011" — while `fsx_capstab` has had a fourth row since your first commit. So the EGA's `0x000F` appeared in **no** SPEC section, and a package author greying a mode menu per §47 had nothing to read. Added the column and the mask, and at §39.18.2 and §53.8 where the same three are enumerated.

The paragraph there saying `vid_detect` "admits EGA-class cards as `VID_VGA`… if those machines ever matter, this is the routine that learns the finer probe" is gone. **This branch is that finer probe** — nice to be able to delete it.

---

## Where it stands

`make` and `make small` clean; `os88test full` **33/33**; kernel `.text` +237/+206 on `main`, **no rung crossed**, `KERN_SIZE` and both guards untouched — though the image rung is down to 105 bytes (big) and 172 (small), which is tighter than this branch started and worth knowing before the next thing lands. All six images `--verify` clean.

On the glass at 640×350: desktop, menus, a window, About, the Display page, a mode round trip, Cyclone in colour, and the screen saver blanked to **100% black** and woken back to a pixel-identical desktop.

**Still open for the field test** — none of it blocking:

- The `vid_disp_init` teardown, which needs two cards.
- Whether an IBM EGA really is missing both enable bits. I'm going on the documented EGA/VGA register difference; @Bs0Dd's card is the thing that would settle it, and the palette blank is correct either way.
- The Colour theme still greyed on EGA (your `EGA-PLAN.md §7` deferral, correct and honest — just the most visible thing a 16-colour card owner will notice).
- Whether `OSAPI_FSX_CAPS` should answer a depth in `DH`, which is the root fix for last round's item 1 and is @jggonz's call, not mine.

Sorry for the two-part review — the second reader's report reached me after I'd already commented. The change is in good shape; the EGA-specific edges were just deeper than one pass found.

pentagram666:
Added **`b7d227c` — the Colour system theme now runs on EGA**, as a separate commit on top of your round-2.

The gate was `thm_set`'s `cmp [vid_kind], VID_VGA` — an identity test that was standing in for a colour *capability* while VGA was the only colour adapter. `VID_EGA` broke the proxy: it refused Colour on the one machine class §76.12 is patterned on ("because that is all an EGA palette has"). The theme itself needs nothing an EGA lacks — six palette **indices** (0/1/3/7/8/15) copied into the 8-byte live block, drawn with the same `OSAPI_SET_COLOR` + `gfx_fill`/`font_run` path that already renders Solitaire and Cyclone at 16 colours on the card. No DAC, no Attribute Controller, no palette programming.

So `thm_set` and `cp_thm_colgrey` gate on `[vid_mono] == 0` now (a four-plane adapter; `= 1` still falls back to Bright per §39.4). `[vid_mono]` is published by `vid_apply` → `vid_depth_set` before every `thm_set` caller — same lifecycle `vid_probe_avail`'s re-key relies on.

**It touches §76.12's contract** ("VGA's alone" → a depth capability), so it's a policy change on top of the geometry work — flagging it for your call. §76.12.4 (extended desktop) is unaffected and now says so: an EGA is never a *live* extended-desktop primary (a switch to it collapses the span via your `vid_disp_init` teardown), so the primary there is always the VGA and `[vid_mono] = 0` means exactly what `VID_VGA` did.

**Kernel size: zero delta.** `cmp byte [vid_mono], 0` and `cmp byte [vid_kind], VID_VGA` are both 5-byte `80 3E addr imm8`; `build/kernel.bin` is byte-identical to your `90357cc` (110,951). `KERN_SIZE`, both guards, every rung unchanged.

Verified locally (nasm 3.02 / QEMU 11.1 on Windows; `make test-full`'s `buildmatrix` needs a real `make`, otherwise `make` + `make small` clean `-w+error`, `test-fast` 30/30, `kernmods` green): booted `VIDEO=ega` at 640×350 — Theme page shows **Colour selectable, not greyed**; selecting it colours the desktop (`CTEAL`), menu bar + dock (`CLGRAY`) and the focused title bar (`CBLUE`/`CWHITE`); Solitaire launched and renders its own green/red/blue/black palette correctly under the theme (chrome themed, content untouched); 640×350 intact, nothing clipped. VGA (`auto`) build: Colour works identically at 640×480. `VIDEO=cga` build: Colour row still greyed.

PR #139, opened 2026-09-02
EGA: the Colour system theme is a depth capability — follow-up to #138

pentagram666:
## What & why

Follow-up to #138 (EGA 640×350, merged as `e4bf59f`). That PR left the Colour
system theme (§76.12) greyed on EGA, on the reasoning that it needed an
Attribute-Controller palette path. **That reasoning was wrong** — the Colour
theme does no palette programming at all — and @jggonz's review flagged the
greyed row as the most visible thing a 16-colour EGA owner would notice. This
enables it, as a one-predicate capability change.

## The mechanism it turned out not to need

The Colour theme is **six palette indices** — `CBLACK` `CBLUE` `CTEAL`
`CLGRAY` `CDGRAY` `CWHITE` (0/1/3/7/8/15) — that `thm_set` copies into an
eight-byte live block. Every chrome site then draws with the same
`OSAPI_SET_COLOR` + `gfx_fill` / `font_run` path that already renders
Solitaire and Cyclone at 16 colours on an EGA. No DAC, no Attribute
Controller, no palette programming anywhere in the theme, and the default
16-colour table is identical under EGA mode 10h and VGA mode 12h.

## The change

`thm_set`'s eligibility check for `THM_COLOR` was `cmp byte [vid_kind],
VID_VGA` — an **identity test standing in for a colour capability**, correct
only while VGA was the sole colour adapter. `VID_EGA` is a four-plane
16-colour adapter that is not a VGA, so the proxy broke and refused Colour on
the one machine class §76.12 is patterned on ("because that is all an EGA
palette has").

Both that check and its mirror in `cp_thm_colgrey` (the single predicate
behind the Control Panel row's greying, its glyph and its click refusal, §47
rule 1) now read **`cmp byte [vid_mono], 0`** — a four-plane adapter, VGA or
EGA; `= 1` is a banked 1bpp one, where §39.4 still reduces the palette and
"Colour would be Bright with extra steps" as before. `[vid_mono]` is
published by `vid_apply` → `vid_depth_set` before every `thm_set` caller (the
boot resolve, `drv_cfg_unpack`, `vid_switch`, the Control Panel) — the same
lifecycle `vid_probe_avail`'s re-key relies on.

No palette code added. No theme or graphics refactor. `THM_COLOR` and its
`thm_tab` row already existed (`KERN_BIG` only — `kern_small` has no Colour
theme on any adapter, §76.12.3).

**SPEC.md §76.12 / §76.12.1 / §76.12.4** rewritten to state the depth
capability rather than "VGA's alone". §76.12.4 (extended desktop) is
unaffected and now says so: an EGA is never a *live* extended-desktop primary
(a switch to it collapses the span via `vid_disp_init`, §39.24), so the
primary there is always the VGA and `[vid_mono] = 0` means exactly what
`VID_VGA` did. `docs/EGA-PLAN.md §7` corrected — the earlier
Attribute-Controller claim is gone.

## Kernel size — zero delta

`cmp byte [vid_mono], 0` and `cmp byte [vid_kind], VID_VGA` are both 5-byte
`80 3E addr imm8`. `build/kernel.bin` is **byte-identical** with and without
this change (110,947 bytes). `KERN_SIZE`, `.text`, `.text+.bss`, both guards
and every rung: unchanged.

## How verified

nasm 3.02 / qemu-system-i386 11.1 (native Windows; `make` via GnuWin32, so
`test-full`'s recursive `buildmatrix` tier could not run here — the merged
branch was 33/33 on Linux):

- `make` and `make small` clean, `-w+error`.
- `checkdocs` / `docindex` pass; `make test-fast` 30 / 30; `kernmods` green
  (`VIDEO=cga`, `VIDEO=herc`, `kern_small` all still assemble).
- **Booted `VIDEO=ega` at 640×350**: Control Panel → Theme shows **Colour
  SELECTABLE, not greyed**; selecting it colours the desktop (`CTEAL`), the
  menu bar and dock (`CLGRAY`) and the focused title bar (`CBLUE`/`CWHITE`);
  **Solitaire launched** and renders its own green/red/blue/black palette
  correctly under the theme — chrome themed, window content untouched;
  640×350 geometry intact, nothing clipped.
- **VGA (`auto`) build at 640×480**: Colour works identically — the change is
  byte-neutral there (`[vid_kind] == VID_VGA` ⟺ `[vid_mono] == 0`).
- **`VIDEO=cga` build (1bpp)**: the Colour row is still **greyed**.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -43584,7 +43584,7 @@ chain, `wm_fit`, the chrome, `desk_rowcalc` and the cursor all follow from
 | `vid_setmode` | `int 10h AX=0012h` + A000 clear of `SCREEN_H` rows | `int 10h AX=`**`0010h`** + A000 clear of `EGA_H` (350) rows (§39.6, §39.23) |
 | `vid_depth_set` | colour: `[vid_mono]=0`, `[vid_planes]=4` | identical — `VID_EGA` takes the colour arm |
 | `vga_vline_core`, `vga_xor_hline` | clip y to `SCREEN_H` (a VGA-only constant) | now clip to `[vid_ch]` — a no-op edit on VGA, byte-checked (§39.9) |
-| Colour theme (§76.12) | DAC, `int 10h AX=1002h` on a VGA | **VGA only** — an EGA machine gets Bright/Dark; an Attribute-Controller path is deferred (`EGA-PLAN.md §7`) |
+| Colour theme (§76.12) | eligible | eligible — the theme is six palette *indices*, no DAC or AC programming; `thm_set` / `cp_thm_colgrey` gate on `[vid_mono]`, not `VID_VGA` (§76.12.1) |
 | `OSAPI_WM_PREFER` kind index | `vid_kind` = 0 | clamped to 0 — an EGA takes VGA's preference row (§39.8) |
 | the idle **blank** (§64) | SR01 bit 5, Screen Off | **the PALETTE** — `vid_ac_pal`, all sixteen AC registers to black and mode 10h's own sixteen back. An EGA has neither enable bit: SR01 bit 5 and the AC index's bit 5 (Palette Address Source) are both VGA additions, so writing either on an EGA lands in a reserved field and blanks nothing |
 | `vid_dual_ok` | `[vid_avail]` alone | **also `[vid_kind]`** — and that is what makes the predicate non-invariant, below |
@@ -43611,9 +43611,10 @@ restores it.
 
 **Out of scope**, recorded so the next reader does not re-open it: `VID_EGA`
 as a member of a `KERN_BIG` per-display pair (§39.12 — it may be the primary
-that *ends* one, per the paragraph above, but never a half of one), the
-Colour theme on EGA, an EGA on a **monochrome display** (§39.1's second scope
-cut — mode 0Fh is the right answer and is a second setup), and a 64KB EGA.
+that *ends* one, per the paragraph above, but never a half of one), an EGA on
+a **monochrome display** (§39.1's second scope cut — mode 0Fh is the right
+answer and is a second setup), and a 64KB EGA. (The Colour system theme
+*is* supported on EGA — §76.12, a depth capability now.)
 
 ## 40. apps/fractal — the progressive renderer and its restore cache
 
@@ -75708,7 +75709,7 @@ a user's Dark choice **round-trips untouched** through a `kern_small` boot
 instead of being reset by one. It is the same argument `[wm_animon]` is
 declared outside `%ifdef WM_ANIM` for (§76.11).
 
-### 76.12. Color — the third theme, and VGA's alone
+### 76.12. Color — the third theme, and the colour adapters'
 
 Bright and Dark are two arrangements of one pair of pens. **Color is the first
 theme that needs the distinctions the chrome actually has**, and adding it is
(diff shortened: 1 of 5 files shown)


pentagram666:
Field-tested on real hardware — **TOP Super Turbo III** (AMD 8088-1 @ 10 MHz, 8087, 640 KB) with an **ATI EGA Wonder**, booted from the 360 KB floppy pair:

- Control Panel → Theme: **Colour is selectable** (not greyed), and selecting it colours the desktop / menu bar / dock / focused title bar as expected.
- The four packages fixed in #138 — **Missile Command, Tank, Tracker, Cyclone — render in colour** on the EGA, under both the default theme and the Colour theme.
- 640×350 desktop, dock reachable, geometry correct.

So both the merged #138 work and this follow-up check out on an actual period EGA card.

Elendilon:
I'm jealous of the ATI EGA Wonder! Always wanted one.

Want to check on extended desktop with EGA? I've not run it yet to test (in the middle of an optimization pass), but there is a bunch of "Here is how I fit on this mode" and "here is how I straddle or switch adapter types", and EGA is kind of "hercules vertical sizing + vga color".

jggonz:
Reviewed, verified and pushed one fix commit (`7c9800c`) on top of yours. Nothing to merge from `main` — the branch is already cut from its tip — and the PR is mergeable as it stands now.

The change itself is right: the Colour theme really is six palette indices, and an EGA really can show it. I booted your `VIDEO=ega` build in QEMU and got exactly what you described — Colour selectable, teal desktop, grey bar and dock, blue title bar at 640×350 — and the `VIDEO=cga` build keeps the row greyed. Both kernels assemble under `-w+error`, the full regression tier passes, and `kernel.bin` is byte-identical to `main`'s bar the build number, as your body said.

**One thing I changed: which byte the test reads.**

*What could have gone wrong for a user.* Someone with a VGA and a Hercules side by side, desktop extended, drags the Control Panel onto the Hercules and opens Theme. The Colour row shows greyed, and clicking it does nothing. Drag the panel back onto the VGA: Colour is live again and works. Same machine, two answers, depending on which monitor the panel happened to be sitting on. On `main` that machine accepts Colour from either monitor.

*Why.* `[vid_kind]` is the primary adapter and never changes while the machine runs. `[vid_mono]` looks like the same fact but is not: on a two-card desktop every drawing call re-derives it from the display the drawing lands on, and inside one lock hold nothing puts the primary's value back (SPEC §39.14.3, §39.14.6). The Control Panel evaluates `cp_thm_colgrey` right after lettering the Dark label, and on the click's release edge right after redrawing the control — both on the panel's own display. So with the panel on the Hercules the byte says "1bpp", and the row greys and the click cancels. The three other `thm_set` callers (boot, `drv_cfg_unpack`, `vid_switch`) all go through `vid_apply` first and were never affected — this is only the Control Panel, and only on a VGA+Hercules extended desktop.

*What I changed.* Both sites now test the primary's kind for the two four-plane values — `VID_VGA` or `VID_EGA` — which is the same pair `wm.inc` already uses to derive a plane count. Your capability is unchanged (EGA is eligible, 1bpp adapters refused); it is just read as a machine fact rather than a display fact. Cost is 7 bytes in `thm_set` and 7 in `CTRL.DRV`, no rung crossed. SPEC §76.12.1 / §76.12.4 and EGA-PLAN §7 are reworded to match, with the account of why `[vid_mono]` is the wrong byte there so nobody re-derives it.

Measured under MartyPC on the two-card `os8088_xt_vga_herc` machine, reading `[thm_kind]` back after clicking Colour:

| panel on | `main` | your commit | with the fix |
|---|---|---|---|
| VGA | accepted | accepted | accepted |
| Hercules | accepted | refused, row greyed | accepted |

**Flagged, not changed**

- The kernel size baseline in `docs/KERNEL-MEMORY.md` was 233 bytes stale from #138 (it did not bless). I blessed both variants in the fix commit and the message says which bytes are whose.
- Your commit message says 110,951 bytes and the PR body 110,947; the binary is 110,947. Cosmetic.
- Elendilon's question about EGA in an extended desktop is EGA-PLAN §7.1, still deferred, and out of this PR's scope.

Thanks for the field run on the EGA Wonder — a real 350-line card confirming both #138 and this is worth more than any of the emulator rows.

PR #143, opened 2026-09-03
Hibernate: save the whole machine to the hard disk and pick up exactly where you left off

jggonz:
## What this gives you, in one sentence

You can now put the machine to sleep on disk. Pick **Hibernate...** from the chip menu, confirm, and everything that is open — your windows, your documents, the programs running behind them — is saved to the hard disk. Switch the machine off. When you turn it on again, it asks whether you want to carry on where you left off, and if you say yes, you are back at the same desktop within moments.

## Why it matters

A 4.77 MHz machine takes a long time to boot and longer to get everything open again. Hibernate turns "start over" into "continue". It is the same idea a laptop uses, built for an IBM PC/XT with a hard disk.

## How to try it, step by step

1. Boot a machine that has a hard disk. `make xt-mfm` is a new 86Box machine with a 20MB drive for exactly this; the live USB image works too.
2. Open a couple of things. An About box is enough to see the effect.
3. Chip menu (top left) → **Hibernate...**. A small window tells you where the file will go, for example `C:\HIBERNAT.IMG`. Click **Hibernate**.
4. The screen goes to text and says it is safe to switch off. Do that, or press a key to restart.
5. On the next boot the desktop appears with one question: **Resume** or **Discard**. Click Resume. Your windows are back.

Nothing is asked about file names or folders. The file always goes to the root of the disk you booted from, so the next boot can find it by itself.

## What to expect, and what not to

- **Works on:** a hard disk the machine's BIOS knows — the boot partition, or one served by the hard-disk driver. That covers the live USB, an installed disk, and the new MFM machine.
- **Greyed out, with the reason in the menu:** when there is no hard disk (`Hibernate (No HD)`), or when a program is holding extended memory (`Hibernate (XMS)`), which no shipped program currently does.
- **Refused with a message:** a disk the BIOS cannot address directly, or not enough free space. Nothing is written in either case.
- **Does not survive:** network sessions and sound-card state. Those drivers are unloaded before the save and loaded again after the resume, like any power cycle. The clock is taken from the boot that just happened.
- **Safety:** a hibernation file from a different build of the OS, a different memory size, or a different video card is recognised and refused rather than loaded. Discard deletes the file. Resume deletes it once it has been used.

## What it cost, and why that was acceptable

The feature lives almost entirely in a file on the system disk, `HIBER.DRV`, that is loaded only when you pick the item or when a saved file is found at boot. The resident kernel grew by about 670 bytes, which spent two of the eighteen spare steps in the memory budget. That budget is exactly what those steps are held for. This PR also adds the rule to CLAUDE.md that new features consider this "load it only when needed" approach before growing the kernel.

## How we know it works

- A new automated test boots an emulated XT with a hard disk, opens a window, hibernates by clicking the button with the mouse, restarts, resumes by clicking again, and then checks the machine's memory directly: the window is back, the locks are released, the clock is ticking, and the hibernation file is gone. It then repeats the whole thing through the keyboard and takes Discard instead. Both the boot-partition and the driver-served disk configurations pass.
- The full pre-merge test suite passes.
- Two bugs were found and fixed on the way by that test: a click that never landed on a button (a register clobber in the hit test), and a scheduler counter that was captured mid-write and left the resumed machine unable to switch tasks.

## Also in this PR

- `make xt-mfm`: an 86Box IBM XT with a 20MB ST-225 MFM drive, the period configuration this feature is meant for.
- A small addition to the hard-disk driver's interface so the resume can ask a driver-served volume where its sectors physically are.
- SPEC.md §86 is the full design record, including what was deliberately left out (extended memory, IDE drives the BIOS does not see) and why.

## What I would like from a reviewer

If you have a real XT or AT with a BIOS-visible hard disk, try steps 1 to 5 above. The emulators cannot tell us how long the save and the restore take on real iron; docs/FIELD-MACHINES.md is where that number would be recorded.

https://claude.ai/code/session_016VfNrZhaL3bF2tDFAKbEPh

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,7 @@ vm/xt-c64/nvr/
 vm/286-c64/nvr/
 vm/xt-weave/nvr/
 vm/386-weave/nvr/
+vm/xt-mfm/nvr/
 
 # Page snapshots the Playwright MCP server drops in the working directory
 .playwright-mcp/
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -238,6 +238,8 @@ sees an up-to-date `kernel.bin`, boots the previous configuration, and it reads
 exactly like the feature being broken.
 
 86Box targets for period hardware, one per `vm/` directory: `xt`, `xt-640`,
+`xt-mfm` (a 20MB ST-225 on a Xebec MFM controller — the machine to install
+and hibernate on; `build/mfm20.img` is created blank and kept),
 `xt-cga`, `xt-hercules`, `xt-ega`, `xt-multimon`, `xt-sound`,
 `xt-sound-1.44`, `286`,
 `286-sound`,
@@ -345,6 +347,17 @@ learned.
   one 64KB window because offsets are 16 bits. They are relieved by different
   mechanisms. **Raising either is a decision to take with whoever asked for the
   feature, not a build fix.**
+- **Before spending a resident byte, ask whether the feature is an ON-DEMAND
+  MODULE** (§2.8, `kernel/mod.inc`, docs/ONDEMAND-PLAN.md §1's test): kernel
+  code that ships as a file (`CTRL.DRV`, `FORMAT.DRV`, `CLONE.DRV`) and is
+  read into a heap claim when the feature is asked for, freed when it is
+  done. A feature qualifies when the system disk is already required to use
+  it, or can be required without interrupting what the user was doing. When
+  it qualifies, the resident part is the menu item, the greying predicate and
+  the thunks; everything else goes in the module. Only a feature that fails
+  the test may grow `.text`/`.cold`, and moving code to `.cold` or the boot
+  overlay relieves `KERN_CODE_MAX` but not `KERN_BUDGET` — a module relieves
+  both. Check `tools/kernsize.py` before and after: a byte costs a byte.
 - **A heap claim can MOVE, and the default is that it may not** (§66). A record
   is born `MC_RLOC` = 0, PINNED; `OSAPI_MEM_MOVABLE` opts one in and takes a
   relocation **proc**, not the address of the word naming the block — a holder
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -349,7 +349,8 @@ learned.
   feature, not a build fix.**
 - **Before spending a resident byte, ask whether the feature is an ON-DEMAND
   MODULE** (§2.8, `kernel/mod.inc`, docs/ONDEMAND-PLAN.md §1's test): kernel
-  code that ships as a file (`CTRL.DRV`, `FORMAT.DRV`, `CLONE.DRV`) and is
+  code that ships as a file (`CTRL.DRV`, `FORMAT.DRV`, `CLONE.DRV`,
+  `HIBER.DRV`) and is
   read into a heap claim when the feature is asked for, freed when it is
   done. A feature qualifies when the system disk is already required to use
   it, or can be required without interrupting what the user was doing. When
--- a/Makefile
+++ b/Makefile
@@ -57,6 +57,7 @@ endif
 
 VM    := $(CURDIR)/vm/xt
 VM640 := $(CURDIR)/vm/xt640
+VMMFM := $(CURDIR)/vm/xt-mfm
 VMCGA := $(CURDIR)/vm/xt-cga
 VMHERC := $(CURDIR)/vm/xt-hercules
 VMEGA := $(CURDIR)/vm/xt-ega
@@ -1301,7 +1302,7 @@ $(shell mkdir -p $(BUILD); \
         [ -f $(VIDSTAMP) ] || { rm -f $(BUILD)/.video-* $(BUILD)/kernel.bin \
                                       $(BUILD)/kernel-full.bin \
                                       $(BUILD)/ctrl.drv $(BUILD)/format.drv \
-                                      $(BUILD)/clone.drv \
+                                      $(BUILD)/clone.drv $(BUILD)/hiber.drv \
                                       $(BUILD)/boot.bin $(BUILD)/boot360.bin \
                                       $(BUILD)/hdd.bin $(BUILD)/hdd.drv \
                                       $(BUILD)/hddtool.bin $(BUILD)/hddtool.drv \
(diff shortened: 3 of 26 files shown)


jggonz:
Review of #143, one pass: five reviewers over the module, the resume path, the kernel glue and driver ABI, the window and module lifetime, and docs versus code; every finding read again by eye before it was fixed. One commit pushed to `hibernate`, 5fd566f, on top of 37f7797; main has not moved, so there was nothing to merge. `make`, `make small`, `test-full`, both hibernate rows (26 checks each now), the Ethernet gate, an fsck of every image and a boot on VGA and CGA all pass on the pushed tree.

**About the mouse report.** I could not reproduce it. On QEMU's VGA with its Microsoft serial mouse the Hibernate button and the Resume button both fire on a click, an instant click, and a click that jitters between press and release; Discard by mouse works too; and the MartyPC gate clicks both buttons with a real serial packet and passes. The hit test that broke the click ("only Enter worked") is fixed in 37f7797 itself, so a build from before that commit shows exactly what you describe. Your `build/` is from two minutes before the commit and `build/mfm20.img` is still blank, so I could not tell which binary you clicked on. If it still fails after `make` on the pushed tree, say which machine and which disk it booted from and I will chase it there.

**Fixed, in 5fd566f**

1. **Every driver from a previous release would have lost its newest features under this kernel: the Ethernet driver its whole socket surface, so the browser, Telnet and the FTP server would say there is no network.** The driver table grew from 36 to 38 bytes for `DSV_GEOM`, and the kernel's copy was "all of it, or the first 28" — so a driver whose header says 36 fell to 28 cells. It now copies exactly the length the header claims, clamped, over a zeroed table, which is what SPEC 51.2 already said it did. Checked with an A/B: a driver patched to claim 36 keeps its sockets and fetches a page.
2. **A second hibernate after a boot question that was closed with its box could resume into the wrong place.** The old pointer file survived, the new image replaced the old one, and until the new pointer landed (or for good, if that write failed) the old pointer named the new image with the old session's wake address. The old pointer and image are now deleted before a byte is written, pointer first, and the root is entered only after the locks are taken, because detaching the drivers yields and another task could have moved the current folder under the write.
3. **A floppy machine with a RAM disk showed `Hibernate...` live and then refused on the click.** The fixed-volume test counts a RAM disk or a network volume as fixed. One predicate now decides for the menu item, the boot probe and the window, and it skips those.
4. **Every boot with a hard disk paid a full remount of every fixed volume's root, listing and icon harvest included, and a floppy boot with the disk driver paid the floppy's remount back.** The probe now looks in the one root a hibernation can be in, with the quiet navigation and directory walk the driver loader uses.
5. **The memory tag reused 0xFF0B, the retired tag the comment beside it says to leave unused**, so the heap map tool labelled the extent list as the boot overlay and the heapmap soak row would fail any sample taken during a resume. It is 0xFF0C now, and the tools and the SPEC table know it.
6. **Closing the Hibernate window by its close box, Close Window or the Task Manager left the module resident** until a later Cancel. `app_close_win` now drops it on those routes; the module's own closes set the mode first so the hook stays out of their way. Verified on MartyPC: segment held while open, zero after the box, reloaded on the next open.
7. **If a resume read failed, the stub's "Resume failed" message overwrote the loop printing it**, because the text page it prints on is the page the stub runs in and the cursor was at row 0. It prints from row 8, over the dead IVT copy.
8. **A key held at "press any key to restart" repeated through the reboot into the boot question, where Esc is Discard.** The ROM's keyboard buffer is emptied before the question opens and before that prompt.
9. **On an XT, the resume wrote to port 0xA1, which on a 5150/5160 aliases the NMI mask register.** It is left alone below the AT tier, the mouse driver's rule.
10. **The clock copy took twelve bytes where the clock is ten**, so the Control Panel's 12-hour and seconds toggles came back as the fresh boot's. Ten now.
11. **Resume skipped the Control Panel flush a restart does**, losing a setting changed on the fresh desktop; **a refused resume or hibernate left the desktop standing at the disk's root**; **an unreadable disk was reported as "not enough room"**. All three take the restart path's shape now.
12. **`make xt-mfm` booted nothing on a fresh clone**: the config named the system floppy only in 86Box's recent-files list, not as the mounted disk. `fdd_01_fn` added, and `vm/xt-mfm/nvr/` ignored like every other machine's.
13. **Documentation and tools**: SPEC 12's command-id block and 29's kind block carry the kern_big split and `KIND_HIBER`; the kernsize baseline is blessed for both variants (it was quoted in prose but not written into the baseline, so the next change would have been billed for this one); `os88sym.py` resolves symbols in `.modh`; stale mentions of the removed file dialog are gone.
14. **The test** now checks that both files reached the disk before asserting them gone, checks the image as well as the pointer, and waits for the question itself rather than for a settled screen, which the BIOS's disk probe can fake on a loaded host.

Resident cost after the fixes: 259 bytes of .text, 94 of .bss, 365 of .cold, inside the same two rungs the PR already crossed; 16 steps spare, unchanged.

**Flagged, not fixed**

- An XTIDE-only machine (rung 1) shows `Hibernate...` live and refuses on the click. The transport kind is a stable fact, so a third greyed string is possible; SPEC 86.2 documents the click-time refusal, so I left the design.
- Esc on the boot question is Discard with no confirmation. With the keyboard drain it can no longer happen by accident; whether a deliberate Esc should delete a session is a UX call.
- The text page shows glyph garbage for the length of the restore (the staging area is the visible page). Moving the extents to 0x1000 would free rows for a "Resuming" line.
- One extra white fill per repaint of the window, and the vacated rect painted twice on wake. About 2 ms on the 5150, one time.
- The 4KB spare in the room check is smaller than one cluster on large-cluster FAT16 volumes.
- Picking `Hibernate...` while the boot question is up fronts the question instead of greying.

**Checked and cleared**: 512-byte alignment and 64KB-page bounding on every transfer both ways; the image write's data-then-FAT-then-directory order and its rollback; the pointer's self-description and the size check against the directory; the stub's layout, stack, CHS arithmetic (matching both writers), interrupt masking and the IVT going back last; the return frame into the UI task and the lock counts on wake; the window's register contracts on all five callbacks and the hit test; the seven shims, the eleven far entries and the module slot budget; `HIBER.DRV` on all three geometries and the live USB; kern_small assembling with the stub module; the menu-id shift touching nothing a package or test reads by number.

PR #144, opened 2026-09-03
Audio Player (AUDIO.O88) + SB Pro v1/v2 high-speed output

pentagram666:
## What this adds

1. **`AUDIO.O88` — a streaming Audio Player** (new package, `apps/audio/`, `SPEC.md 86`). Lightweight background music for the IBM XT: a WAV file — unsigned 8‑bit PCM, or IMA/DVI 4‑bit ADPCM decoded straight to PCM8 — is **streamed from disk, never loaded whole**, through the existing Sound Blaster ring‑stream infrastructure. It owns no hardware: not the DMA controller, not the SB IRQ, not the DSP, not the DMA page register. It is a client of the `OSAPI_SND_STREAM` verb contract, the architecture `SPEC.md 34` / `45.2` were built for.

2. **SB Pro (v1/v2) high‑speed output in `SOUND.DRV`** (`drivers/sound/sb.inc`). The driver's wide‑rate regime (24–44.1 kHz) was gated at DSP ≥ 4.00 — a Sound Blaster 16. An 8‑bit ISA XT has no SB16, so that whole tier was unreachable on the target hardware. This adds a third output regime — SB Pro high‑speed mono — for DSP ≥ 3.00.

3. Two small **Windows‑host build‑tooling fixes** (`tools/os88ovlchk.py`, `tools/os88index.py`) that were blocking `make` on Windows, unrelated to the feature but needed to build it.

This branch is stacked on `main` and contains **only** the audio/sound/tooling commits — no EGA work.

---

### Why streaming, and why a look‑ahead stage

Tracker and ModPlug load the whole module into RAM. A music file does not fit that model, so the player streams. But `SPEC.md 34.5`'s pinned rule says a stream fed live from disk **will** underrun: an `int 13h` read holds `sch_lock`, which pauses both the kernel refill task and the package worker. The answer is a **look‑ahead stage** in front of §45.2's ring:

```
disk --OSAPI_FILE_READ_AT--> 32 KB look-ahead ring (heap claim) --decoder-->
     2048-byte halves --verb 6/1--> 16 KB SND ring --> SB DMA
```

- The **UI task** opens/closes the stream, parses WAV headers and does every disk read (file slots are UI‑task only, `SPEC.md 20.6` rule 7).
- The **worker** decodes and feeds the ring lock‑free on the any‑task verbs 1/3/6. When the look‑ahead ring runs low it raises `[ap_req]` and `OSAPI_WM_WAKE`s the UI task, which services one read in `ap_onwake` and clears `[ap_req]` last — the FTPD handshake (`SPEC.md 77.1`).
- `[ap_mixing]` brackets a feed pass; `ap_stream_close` drains it before any UI‑side reset touches engine state (`SPEC.md 45.2`).

### The RIFF/WAVE parser (`apwav.inc`)

Does **not** assume `fmt ` is followed by `data`, a fixed header size, or the absence of `LIST` / `fact` / `INFO` chunks. Walks the chunk list, honours the RIFF odd‑length pad byte, skips every chunk it does not recognise, slides a fresh cluster window forward past a large preamble (`APW_WINDOWS` windows — a header larger than that is refused, not mis‑read), and bounds‑checks every size the file claims against the file length before trusting it. Rejects non‑mono, unsupported tags/bits/rates, bad block alignment and bad ADPCM samples‑per‑block, each with a specific reason on the glass.

### The decoder (`apdec.inc`)

PCM8 is a pass‑through. IMA/DVI ADPCM is decoded from the **public IMA specification** — the 89‑entry step table and 16‑entry index table only; nothing vendored. The decoder is **resumable**: `apd_pull(want)` keeps predictor / index / block cursor / nibble phase across calls, so a 2048‑byte stage half never has to line up with an ADPCM block boundary. There is no intermediate 16‑bit PCM buffer. Verified byte‑identical to ffmpeg's `adpcm_ima_wav` over a whole 1.8 M‑sample file.

### What real hardware taught us (and what changed)

The first cut was **heavier than ModPlug** at PCM8/8 kHz on an emulated XT — suspicious, since PCM8 has no decoder. Profiling with a compile‑time counter build (`-DAPROF`, `make … AUDIOPROF=1`, `AUDIOP.O88`) found it: `ap_refill_chunk` read **one 2 KB cluster per `int 13h`, ~5×/second**, and every one of those reads takes `sch_lock`. ModPlug's module is resident — zero `int 13h` during playback. Invisible on an emulator, which is exact about work and useless about time.

Round of fixes, each its own commit, each measured:

| change | effect (measured, 20 s of PCM8/8 kHz) |
|---|---|
| **16 KB gulp reads** instead of one cluster | `FILE_READ_AT` 106 → 14; the worker→WM_WAKE→onwake handshake 182 → 14 |
| adaptive worker sleep, feed early‑out, throttled reap | worker no longer wakes ~18×/s when idle |
| change‑gated + throttled redraw, skip when obscured | progress redraws ~18–24/s → ~2/s |
| **don't decode a half without a full half of input** | the ADPCM decoder is stateful — a short pull consumed input then discarded the output, heard as a stutter |
| **ADPCM input buffered 512 bytes at a time** + per‑sample `push`/`pop` stripped from the IMA loop | ~190 cycles/sample of stack traffic removed — roughly a third of a 4.77 MHz 8088 |

### File workflows

- **One add path.** File ▸ Open, `.WAV` association, and the startup `OSAPI_ARG_FILE` all converge on `ap_add_file(name, dir, vol, flags)` (`SPEC.md 86.2.1`). It validates the extension, de‑duplicates on name + dir + vol, and starts per the flags.
- **File ▸ Open does not interrupt** a playing track — it adds and selects, and only starts if the player is idle (`SPEC.md 86.3`).
- **Single‑instance handoff was built, then disabled by default** (`-DAP_HANDOFF` to re‑enable). It works under QEMU, but on a 4.77 MHz XT the running instance's poll has to *remount the system root* to read the rendezvous file — seconds per check — and a playing ADPCM track polls often enough that the machine spends much of its time remounting. `SPEC.md 86.11.1` records the design and why it's off; a second double‑click opens a second window, the OS default.
- **Drag‑and‑drop from the File Manager is not possible** — the window manager has no cross‑app file‑drop event. Proven from source, documented as a limitation, no kernel change.

### The window

Layout is computed from the **live** content geometry every paint (`SPEC.md 39` — three adapters, one binary); drawing is deliberately plain (no visualiser, no animation). The About menu opens a small card: name, `Version 0.6b`, blurb, and the accepted rate set. All transparent‑text sites use `OSAPI_FONT_RUN` (ink 0 / paper 15) so nothing is drawn twice.

---

## The SB Pro high‑speed driver change

`drivers/sound/sb.inc` now has three output regimes, chosen in `sbl_v_open` by rate and DSP version, all mono unsigned‑8:

| rate | regime | rate cmd | start cmd | pause / stop |
|---|---|---|---|---|
| ≤ 22,222 Hz | time constant (any DSP) | `40h` + TC | `1Ch` / `14h` | `D0h` / `D4h` |
| > 22,222, DSP ≥ 4.00 | wide (SB16) | `41h` + Hz | `C6h` + mode | `D0h` / `D4h` |
| **> 22,222, DSP ≥ 3.00** | **SB Pro high‑speed** | `40h` + TC | **`90h`** | **mask/unmask 8237 ch 1; DSP reset on close** |

A new `sbl_hisp` flag, set for 22,223–44,100 Hz on DSP ≥ 3.00. It borrows the wide regime's 4 KB double buffer and `SBL_WD_WIDE` watchdog but programs a TC byte, not `41h`. `sbl_hw_start` gains a `.hispeed` branch: block size (`48h`) then `90h`. Because `90h` locks the DSP out of commands until a reset:

- `sbl_halt` masks 8237 channel 1 (`out 0x0A, 05h`) instead of writing `D0h` — one OUT, ISR‑safe, and the underrun path in `sbl_isr` uses it too;
- `sbl_go_on` unmasks (`out 0x0A, 01h`) instead of `D4h`;
- `sbl_stop_stream` masks in the `cli` window, retires the IRQ and releases the record, then after `popf` runs the new near `sbl_dsp_reset` to leave high‑speed mode for the next open.

**Everything DSP < 3.00 and every rate ≤ 22,222 is byte‑for‑byte the old code.** `SPEC.md 34.5` gets the three‑regime table.

The player's accepted rates are `{8000, 11025, 16000, 22050}` on any SB plus `{24000, 32000, 44100}` on an SB Pro or SB16 (`SPEC.md 86.15`). `apw_parse` accepts all seven and lets the driver be the authority; a `verb 0` refusal above 22 kHz is reported as *"Rate needs a Sound Blaster Pro"*.

---

## Test results

**Physical hardware — IBM PC/XT‑class, 8088 @ 10 MHz, real Sound Blaster 2.0:**

| Format | Sample rate | Result |
|---|---|---|
| PCM U8 mono | 8 kHz | ✅ |
| PCM U8 mono | 11.025 kHz | ✅ |
| PCM U8 mono | 16 kHz | ✅ |
| PCM U8 mono | 22.05 kHz | ✅ |
| IMA ADPCM mono | 8 kHz | ✅ |
| IMA ADPCM mono | 11.025 kHz | ⚠️ depends on the track |
| IMA ADPCM mono | 16 kHz | ❌ |
| IMA ADPCM mono | 22.05 kHz | ❌ |

This is not "an emulator thinking it works" — PCM streaming was played to 22.05 kHz on a physical 8088 machine.

**86Box — 8088 @ 10 MHz, Sound Blaster Pro v2 (validates the new driver path):**

| Format | Sample rate | Result |
|---|---|---|
| PCM U8 mono | 24 kHz | ✅ |
| PCM U8 mono | 32 kHz | ✅ |
| PCM U8 mono | 44.1 kHz | ❌ |

24 and 32 kHz on an SB Pro confirm the high‑speed regime works on the glass; 44.1 kHz does not sustain (CPU feed rate / driver edge — under investigation, and outside what an XT would ask of it).

**86Box — 4.77 MHz (the stock‑XT baseline; these runs are 86Box, not physical):**

| Machine | CPU | Audio | Result |
|---|---|---|---|
| IBM XT (1986) | 4.77 MHz | PCM 8 kHz | ✅ |
| IBM XT (1986) | 4.77 MHz | PCM 11 kHz+ | ❌ not adequately |
| DTK PIM‑TB10‑Z | clocked to 4.77 MHz | PCM 8 kHz | ✅ |
| DTK PIM‑TB10‑Z | clocked to 4.77 MHz | PCM 11 kHz | ✅ |
| DTK PIM‑TB10‑Z | clocked to 4.77 MHz | PCM 16 kHz | ✅ |

**QEMU** (functional verification only — exact about work, useless about time): 28 s of PCM8 @ 22 kHz gap‑free with 14 s of it while another window holds the focus; both codecs; all transport controls; the SB16 wide‑rate path (24/32/44.1 kHz) plays with pause/resume/stop intact.

### Reading of the results

- **PCM8 on a stock 4.77 MHz XT is the sweet spot; higher rates depend on the machine.** The IBM XT's slow bus doesn't sustain 11 kHz; a faster‑bus XT clone (the DTK) does 16 kHz. At 10 MHz PCM is solid through 22.05 kHz on real silicon.
- **IMA ADPCM decode is ~half a slow 8088.** 8 kHz is fine everywhere; 11 kHz is marginal (track‑dependent) even at 10 MHz; 16 kHz+ is over budget. The decoder is correct and optimised (verified against ffmpeg, per‑sample stack traffic removed) — this is the irreducible cost of the IMA step maths on an 8088, not a bug. A 286 clears it with headroom.
- **SB Pro high‑speed works** for 24/32 kHz mono on an SB Pro v2; this is the path an XT/286 without an SB16 needs.

---

## Deliberately not done

No visualiser / spectrum bars; no seeking (an ADPCM seek must respect block boundaries and decoder state); no `.M3U`; no clickable playlist rows; no arm‑on‑press / fire‑on‑release for the transport buttons. Single‑instance handoff is built but off by default (see above). Drag‑and‑drop needs a `W_ONDROP` OS event that does not exist. SB Pro high‑speed at 44.1 kHz needs more work on the feed/driver edge.

---

## Project‑rule checklist

- **`SPEC.md`** updated *before* the change: §86 (the Audio Player's binding section, with §86.2.1 / 86.3 / 86.5 / 86.5.1 / 86.5.2 / 86.6 / 86.7 / 86.10 / 86.11 / 86.11.1 / 86.12 / 86.15 / 86.16 / 86.17 / 86.20), and §34.5 (the three output‑rate regimes).
- **`docs/INDEX.md`** regenerated by `tools/os88index.py` (the AUDIO PLAYER row + §86); it is checked by every `make`.
- **`docs/AUDIO-PLAN.md`** is the design record — the audit, the codec comparison, the performance round with measurements, the file‑workflow round, the SB Pro driver rationale, and what real hardware said.
- **8086‑only**, near model, register discipline, section discipline, adapter‑agnostic (`[vid_w]`/`[vid_h]`/`[vid_stride]`), every disk‑visible base 512‑aligned. Checked against a 1bpp adapter (opaque text, no greyed glyphs).
- **Pure NASM assembly**, following the Tracker / ModPlug / Recorder precedent for this app class (the C SDK deliberately does not wrap the stream verbs). The audit's preferred C‑shell split was not possible because the C cross‑compiler could not be bootstrapped in the build environment; a future C rewrite keeps this engine as the shim verbatim.
- **No binaries committed.** `make build/audio.o88` / `make audio-hdd` reproduce everything; the toolchain is deterministic.
- **Separate logical commits** (18), each building and passing `make test-fast` (30/30).
- `make` (which runs `test-fast`) is clean on this branch. `test-full`'s `buildmatrix` and `ps2mouse` rows fail on this Windows host for environment reasons (GnuWin32 `make` pattern rules; QEMU‑on‑Windows has no `-daemonize`) — they fail identically on a clean checkout, unrelated to this change.

https://claude.ai/code/session_01PyZDLaPNxmmcSDXVKGNs3o


pentagram666:
I've been used to having music playing in the background on my desktop computers since the early 2000s. For me, using a computer for any length of time without music playing actually feels a little strange. That was the main motivation behind my idea for the Audio Player.

I remember reading somewhere in the project that playing music directly from a CD-ROM should work, or is something that is expected to work. I actually have an IDE CD-ROM drive, but I don't have any CDs available to test it with right now, and I don't even know yet whether that drive will work properly in my XT. At some point I'll get some recordable media, burn a disc and give it a try.

So, since CD audio isn't something I can actually use or test on my XT at the moment, I started thinking about playing music files instead — and that's where the Audio Player idea came from.

What I'd ultimately like to do is leave a playlist playing in the background while I connect to BBSes, browse around, download files and games for DOS, CP/M or C64, and then run them on the same XT.

That's really the experience I'm looking for: not just demonstrating that an 8088 is capable of playing digital audio, but actually using my XT the way I'm accustomed to using a computer — with music playing in the background while I'm doing other things.

jggonz:
Thanks for this - a streaming player with a resumable ADPCM decoder, on hardware, is real work, and the PR body's honesty about what the 4.77 MHz machine can and cannot do made it much easier to review.

I merged `main` into the branch and pushed two commits to `audio-pr` on your fork; the PR is mergeable now.

- **0d178fb** - the merge. Both sides had added SPEC.md **§86** (your Audio Player and #143's Hibernate). The branch's numbering wins at the squash, so Hibernate became **§87**; every reference #143 made moved with it (99 lines, scoped to exactly the lines that commit added), `docs/INDEX.md` regenerated, and the precedent recorded in `docs/UPSTREAM.md` beside the other two. Nothing of yours was renumbered.
- **e9a43e5** - the fixes below. Every one was confirmed by reading the code, and where a screenshot could show it, on the glass in QEMU (with an SB16, capturing the card's output to a WAV so I could check the tone that came out was the one that went in - it was, for both codecs).

## What I fixed, and what it would have done to you

**A large track, or one bad file, froze the whole machine.** The progress bar divides the track's total byte count by the bar's width in pixels, in 16-bit arithmetic. Past about 14 MB - a ten-minute track at 22 kHz, which is exactly what the `audio-hdd` image is for - the quotient no longer fits, the 8086 raises interrupt 0, and the UI task never comes back: menus stop dropping, buttons stop answering, the playing tone just stops. A WAV whose `data` size is near 4 GB got there too, because the parser's clamp adds size to offset in 32 bits and never looks at the carry, so the wrapped sum compared small and passed. I reproduced the hang with such a file, and after the fix played a 15 MB track from the HDD image (11:53, bar filling). The clamp now tests the carry; the bar halves total and played together until the division fits.

**The player painted over its own Open dialog.** Every repaint the player starts itself - a track ending, the launch document, the settled first paint, the repaint after a menu command - draws under the gfx lock without `OSAPI_WM_CLIP_SET`, and §11.3's primitives are absolute, so the whole content landed on top of the dialog the moment a track ended or the dialog opened. Reproduced twice. `apu_repaint` now arms the clip, which is what your worker's tick already did correctly.

**File ▸ Open in another folder ended the song.** Two reviewers landed on this independently. The dialog moves the instance's current folder (§38.10, and every navigation, cancelled or not), and the refill reads the playing track *by name*. Standing in the other folder, "not found" reads as end of data - the song ends within one gulp and the playlist advances - and a same-named file there gets spliced in as audio. `ap_open_track` now banks the track's folder and `ap_refill_chunk` re-stands there with `OSAPI_FILE_GOTO_QM` when the two differ (a word compare inside one volume, a quiet mount across). Verified: a 60-second track kept playing through six refills while the dialog sat in another folder, and a track picked from that folder played afterwards.

**Every track open was a full remount, under the gfx lock.** `apl_goto_cur` used `OSAPI_FILE_GOTO` - directory scan, sort, an icon read per file, which your own §86.11.1 measured at seconds on the XT - at every track boundary, including the automatic one between two songs, with the UI frozen behind the lock for that long. `OSAPI_FILE_GOTO_QM` is the quiet stand that *does* move the instance, which is what the comment there asked for.

**The read buffer's alignment was luck.** `ap_scratch` is the `int 13h` target, and its offset was wherever the bss accumulator happened to reach (259 mod 512). CLAUDE.md's hard rule exists because one load position in four puts a 64 KB physical boundary inside a 16 KB read, the BIOS answers error 09h, and the refill takes that as end of data - the song stops a few seconds in, only on the machine whose heap put the package there, and it flips with the next unrelated edit that changes the image size. The accumulator is rounded to 512 before the buffer and the image end padded to 512 (image 8,762 → 9,216 bytes).

**The 8 KB ring tier could never play.** The pre-roll was a fixed six halves (12 KB); the driver refuses a total larger than the ring, so on the small tier every track said "Cannot play this file". The pre-roll now follows the ring it got (ring/2048 − 1, capped at 6): six on 16 KB, three on 8 KB - the Tracker's rule.

**Two silent failures now say what happened.** A refused `OSAPI_TASK_SPAWN` (seven dynamic slots, shared - §20.6 calls it ordinary) left the stream open under "Playing" with nothing feeding it, forever; it is now a refusal with FTPD's sentence, "No free task - close an app and retry". A disk error mid-track (bad sector, pulled disk) read as a normal end and quietly advanced; the status line now says "Cannot read the file".

**On the SB Pro driver - a refused second open wedged a playing high-speed stream.** `sbl_v_open` zeroed the *live* `sbl_hisp` at entry, before the busy refusal. Any second `verb 0` while an SB Pro played 24-44.1 kHz - another instance pressing Play, any package opening sound - was refused, correctly, but left the playing stream flagged legacy: its next underrun wrote `D0h`, which the DSP ignores in high-speed mode, so the stale ring looped audibly; its close skipped the DSP reset, so the DSP stayed in high-speed and every later open was silent until reboot. The comment three lines up explains exactly this hazard for `sbl_ropen`, so the fix is that idiom: a temp `sbl_hsopen`, committed to `sbl_hisp` at `.irqok`. The same commit means a refused high-speed open no longer leaks the flag into a later Recorder capture.

**44,100 Hz played at 45,454 Hz on the SB Pro.** The time-constant division truncates: 1e6/44100 = 22, TC 234, and the card runs at 1e6/22 - half a semitone sharp, a three-minute track ending 5.6 s early, and the same file at two pitches on the two cards. The high-speed branch now rounds (TC 233 = 43,478 Hz, the nearest a byte can say); the ≤ 22,222 path keeps its bytes. §34.5 states the quantisation.

**Smaller things in the same commit:** `sbl_dsp_reset` returns CF on a timeout and the close retries once (its comment said "the next open resets again", and nothing does); the RIFF walker's 16-bit window arithmetic tests the carry (a 64 KB tag block before `data` wrapped the cursor into the window); a 32 KB-cluster volume is refused as "Cluster too large for this player" rather than "Cannot read the file"; ADPCM `nBlockAlign` below 8 is refused, because the 4-byte block header is not resumable across a dry ring; `ap_prime`'s one-gulp start never actually fired (a full gulp is never quite 16 KB of audio, so every track start was three disk visits) - the test is now `AP_RD_CHUNK − AP_HALF`; the worker's progress tick filled the bar white then black (the double write PERFORMANCE.md rule 2 names) and now draws only the strip it grew by; File ▸ Diagnostics is greyed in the build with no counters (§47 rule 5); `make audio-hdd` also picks up `*.WAV`; §86.5.2 said `AP_LA_LOW` is 12 KB (it is 16), and §86.12's size row was stale; ModPlug's and Tracker's "44 kHz needs a DSP 4.x card" - now shown only to an SB 2.0 - say "an SB Pro or SB16".

## Flagged, not fixed

- **Tracker's and ModPlug's 44 kHz items now open on an XT with an SB Pro** where they used to be refused with a message. The mixer cannot feed 44.1 kHz on a 4.77 MHz 8088, so that user gets an underrun-pause loop instead of the old status-line refusal. Greying the item on tier-0 hardware is a design call for those packages, not this PR.
- **22 kHz from a 360 KB floppy on the XT.** A 16 KB gulp from a 9-sector-per-track drive is ~4 track reads holding `sch_lock`, during which nothing feeds the DSP; the SND lead at 22 kHz is 0.56-0.74 s. The look-ahead ring cannot help while the scheduler is frozen. Your table shows 16 kHz on the DTK and 8 kHz on the IBM XT, both consistent with this; worth one 86Box run at 22 kHz on the `xt` target before §86.5.2's "no underruns" is read as covering the 360 KB case.
- **No automated gate for the player** (no WAV fixture, no decode-against-known-vector). Tracker, ModPlug and Recorder have none either, so this is the norm rather than a regression; the fixtures I used are trivially generated and would make a `tests/` row if you want one.
- A fmt chunk shorter than 16 bytes is refused, but as "Bad WAV block alignment" rather than a reason of its own. Harmless.

## What I checked and found right

The heap claim is pinned and never held across a call; the worker/UI handshake bytes and ring indices are single-writer and atomic; the close drain absorbs a worker mid-pass; the worker uses only any-task verbs and takes no lock inside the mixing bracket; callbacks read no window fields and use ES for kernel buffers; teardown is covered on every close route including the Task Manager; the IMA/DVI decoder matches the public tables (and ffmpeg, as you said - my 660 Hz ADPCM fixture came out at 665 Hz, same level as the PCM one); the playlist's capacity, copies, de-dupe and shuffle; the 16 KB gulp coalesces to one or two `int 13h` calls; every §86 constant, rate gate and error string matches the code; the allapps and live payloads pick `AUDIO.O88` up automatically; `-DAPROF` never reaches the shipped binary; the two tooling fixes are no-ops on macOS and Linux; the driver's command order before `90h`, the mono TC formula, auto-init being the right mode for the ring, the ISR's register and ACK discipline, the watchdog gating during a pause, and that no verb, constant or header changed for existing clients. `make`, `make small`, both size guards (the kernel is byte-unchanged), every geometry's fsck, `test-full`, a CGA boot (the greyed item and the pressed Shuffle both read correctly at 1bpp), and the hard-disk image with a 15 MB track.

https://claude.ai/code/session_01WjVc9R91MDagQy3aWcXNzG

Issue #142, opened 2026-09-02
Chart window keeps stale pixels when it is moved or raised


jggonz:
Found while driving the UI to take screenshots for the v1.0.20260902 release pages. QEMU, VGA, `build/os8088.img` + `build/apps.img` at `49ccea13f`.

## What happens

A **Chart** window (SHEET's `Data > Chart Column...`, and presumably `CHART.O88` too) draws its picture once, when it opens. After that it does not repaint:

1. **Dragged** — the frame moves, the chart image is re-blitted, and everything in the window *outside* the chart image keeps whatever was on the glass before the move. Fragments of the Sheet window underneath ("Dat" from its Data menu, and Sheet row headers "9" and "10") were left sitting inside the Chart window's frame.
2. **Raised** — clicking the Chart window's title bar makes it the active window (the menu bar changes to "Chart") but it is still drawn *underneath* the Sheet window. It never repaints, so the Sheet's pixels stay on top of it.
3. The stale pixels **persist**. They survived a 6-second wait and a raise-and-repaint of the window in front of it, so this is not a mid-repaint capture.

## Repro

```
make test                                     # apps.img in B:
```
- Disk B: -> `APPS` -> `SHEET.O88`
- Type anything into a few cells of column B
- Select them, `Data > Chart Column...`  -- the chart draws correctly here
- Drag the Chart window by its title bar

The chart's own bars, its `Column B` caption and its scale all redraw. It is the rest of the window's interior that does not.

## Guess at the cause

SPEC.md §82.1: the chart is composed into a private 4bpp buffer and put on screen with one `OSAPI_GFX_BLIT4`. That blit looks like it happens only when the data changes, not on a damage event -- so a move or a raise repaints the frame and re-blits the picture, and the band of window interior the picture does not cover is never repainted at all.

Dragging the **Sheet** window in the same session is clean, so this is specific to the window whose content is a blitted buffer rather than the window manager generally.

## Not verified

- Whether `CHART.O88` standalone has it (only Sheet's live chart window was driven)
- Whether it reproduces on the 1bpp adapters
- Whether a resize does it too

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MLd5kxqkiH4M1WGVpqbMPy

Koriban:
Had a look at this one. **The first symptom is real and fixed; the second did not reproduce.**

[...]

PR #145, opened 2026-09-03
Chart window: paint the interior, not just the picture — #142's move symptom

Koriban:
Fixes the **move** symptom of #142. The **raise** symptom in that issue did not reproduce and is not addressed here — detail below, and in the issue.

## The defect

`sh_chart_paint` is the `OSAPI_GFX_BLIT4` and nothing else, and that is not enough to paint a window.

The frame is `SH_CHARTWIN_W` × `SH_CHARTWIN_H` = 260 × 200, so its content is **260 × 181**. The image is `CH_W` × `CH_H` = **240 × 160**. That leaves a **20-pixel band down the right and 21 along the bottom** which belongs to the window and was never written by anything — the *"little margin around the CH_W x CH_H canvas"* that `SH_CHARTWIN_W`'s own comment describes. The margin was deliberate; painting it was not considered.

So a move redraws the frame, re-blits the picture, and leaves whatever was on the glass in between.

The issue's guess was close but not exact: the picture **is** re-blitted on a damage event. It is only the band the picture does not cover that never was.

## The fix

Fill the content rectangle before the blit.

**One fill, not four strips round the edges.** A fill costs 756 µs whatever it covers (PERFORMANCE.md), so covering the content once is cheaper than computing and drawing the four the picture leaves, and the blit lands on top of it.

**`OSAPI_WM_GEOM`, not `W_W`/`W_H`.** Those are the *outer* frame, and filling to them paints over the window's own two-pixel border. `WM_GEOM` answers the size of the box `WM_CONTENT` gives the origin of, and `sh_geom` a few hundred lines up already uses it for the same reason. My first attempt used the frame and erased that border; a screenshot **diff** between the two builds is what caught it — the 2px sliver was invisible to looking at the picture.

## Verification

Driven through MartyPC, dragging the same chart across the SHEET window with one script against both builds:

- **VGA** — `screenshots/issue142-before.png` has `rma` from SHEET's *Format* menu and `40`/`90`/`15`/`84`/`9 TOTAL 12795 14229` from its cells sitting inside the Chart frame. `issue142-after.png` is white there. **1,530 differing pixels.**
- **CGA** — `screenshots/issue142-cga.png`, because a drawing change is not done until it has been seen on a 1bpp adapter (CLAUDE.md, §39.4/§47).

`make test-fast`: 30 passed.

## What this does not fix

**The raise symptom did not reproduce.** Driven through the reported sequence — chart on top, raise SHEET over it, click the Chart's title bar — the window comes forward and draws correctly, and the fixed and unfixed builds are **identical to the pixel**. The interior fill neither causes nor cures it.

A theory I could not confirm: what was seen may *have been* this same defect. The window really was raised — the issue notes the menu bar changed to "Chart" — but with only the 240×160 picture repainting, the margin still showed SHEET's pixels, which reads as "still drawn underneath". If so, this covers it. If a session genuinely leaves the window behind, that is a separate defect.

Also untested: **resize**, and `CHART.O88` standalone, which shares this paint proc.


jggonz:
Reviewed and merged nothing yet — one commit pushed to your branch (`967a318`), so #145 now carries your fix and my follow-up. No merge from `main` was needed: you branched from its tip and it hasn't moved.

**The diagnosis is right and the evidence is the good kind.** You found a band nobody had ever written, you proved it with a pixel diff rather than by looking, you caught the border problem with a *second* diff against your own first attempt, and you said plainly which half of the issue didn't reproduce instead of quietly claiming both. That last part is worth more than the patch.

Three things about the answer, and one bookkeeping note.

---

### 1. Filling the whole content paints the picture twice

The blit lands on top of the fill, so all 240×160 pixels of the chart get written a second time. On this machine that isn't just wasted work — it's a *visible* one. The picture goes white and then wipes back in as the blit works down it, and on a 4.77 MHz XT that gap is long enough to watch. PERFORMANCE.md calls it out as one of three defects that are invisible in an emulator and only show up on real hardware, and Set 33 measured the identical pattern in Paint at **2,617 ms of white**. It's also the exact pair `WF_OWNBG` (SPEC.md §11.90.1) was added to *remove* from `wm_draw_win`, so putting it back inside `W_PAINT` runs against the grain of the thing.

And on SHEET this isn't a rare path. `sh_chart_paint` is called directly by the `sh_chartdirty` resync at `sheet.asm:857` — so it fires when you edit a charted cell, not only when you drag the window.

**Changed to:** fill only the two bands the picture doesn't cover, so no pixel is written twice.

### 2. The cost argument came out backwards — and it's two strips, not four

This one's a trap the repo has stepped in before, which is why PERFORMANCE.md says it in so many words:

> **do not quote "756 µs" as the cost of a call, or as a floor a design has to beat.** It is the fixed part of a small one. Measure the call you mean.

Its measured model is 177 µs a scan line plus 0.28 µs a pixel — and that model reproduces PERFORMANCE.md's own published `GFX_FILL 256x128` = 31.7 ms figure exactly, so it's trustworthy:

| | |
|---|---|
| one fill of the whole 258×181 content | **45.1 ms** |
| right band (18×181) + bottom band (240×21) | **38.0 ms** |

So the strips are cheaper *as well as* flash-free. And it's **two** of them, not four — the picture is anchored at the content's top-left corner, so there's nothing above it and nothing to its left. The bottom band stops at the picture's right edge so the corner the right band already covered isn't written twice either.

### 3. `CHART.O88` had the same defect, on the same margin

`ct_paint` in `apps/chart/chart.asm` was its own copy of "one `OSAPI_GFX_BLIT4` of the already-rasterized buffer, nothing else" — on a window of exactly the same size (`CT_WIN_W` × `CT_WIN_H` is the same 260×200 around the same `CH_W` × `CH_H`, carrying the same "a little margin around the CH_W x CH_H canvas" comment). Your PR body listed it as untested; it was actually unfixed, because it's a separate proc.

So the fix now lives in **`apps/os88chart.inc` as `ch_margin`** — the file that exists precisely so these two packages can't drift into two opinions about one drawing — and both `W_PAINT`s call it. It keeps the content origin in `SI`/`DI` rather than asking either package for two more bss words, so no caller convention changed (the two packages each assert their own bss size, and adding words there breaks the build in a way that's fun to diagnose).

### 4. The content is 258 × 181, not 260 × 181

Minor, but SPEC.md is the contract. `wm_geom` takes `TITLE_H` + 1 off the height *and one border column off each side* — `wm_bord` answers 0 or 1 (`kernel/wm.inc:4527`). So the content is 258 wide and the right band is **18** px, not 20. Your height is right. The code was never affected, since it reads `WM_GEOM` rather than the constant — which is the reason using `WM_GEOM` was the right call in the first place.

---

### On the screenshots

I dropped `screenshots/` from the tree. They're review evidence and they're already right here on the PR, which is where CONTRIBUTING.md §7 asks for them — and nothing else in the repo tracks a `screenshots/` directory. Not a criticism of taking them; keep doing that.

### On the raise symptom

Leaving your finding exactly as you wrote it. "Driven through the reported sequence, the two builds are identical to the pixel" is a result, and your theory that what was seen *was* this same defect wearing a different hat is a good one. It stays an open question on #142 rather than something this PR claims.

### How I checked it

Built and driven over QMP against a build of the pushed commit:

- **VGA** — SHEET's chart window dragged across the SHEET window: both bands white, no cells or menu fragments in them, and the window's own 1px border still intact. Then `CHART.O88` launched standalone and dragged over the APPS window: the same.
- **CGA** — because a drawing change isn't done until it's been seen on a 1bpp adapter. `CWHITE` comes out as the adapter's white, and the window running off the bottom of a 200-line screen takes the short-content path without misdrawing the border.
- `make` clean, `make test-fast` 30 passed, all seven images verify.

Over to @jggonz for the approval and the merge.

https://claude.ai/code/session_013Ue18GLizEWCbqgvakYY7x

jggonz:
Addendum — a scope note on this fix, and a separate defect underneath it.

After pushing, I had the move path instrumented directly rather than reasoned about, and the result narrows what #145 can claim. **The fix is real and stays**, but it does not cover every route to the reported symptom.

### `W_PAINT` does not always run on a move

`wm_draw_win` reaches its content fill and the app's `W_PAINT` only when the raise cache misses (`kernel/wm.inc:11165-11177`):

```
call wm_su_try     ; CF=0 -> the content came back from the cache
jc .content        ; ...only CF=1 reaches the fill and W_PAINT
call wm_su_bget
jnc .paint
jmp .growbox       ; whole content restored: no fill, NO W_PAINT
```

With counters on those four exits, a drag of the Chart window took the last branch: the content was restored bit-for-bit from the cache, and `sh_chart_paint` was never called. On the glass, SHEET's pixels were carried verbatim from the Chart's **old** position to its new one — which only a cache restore can do.

So a fill inside `W_PAINT` — yours or mine — helps when the cache **misses** (a frontmost window that was never covered has nothing banked, which is the case in all three scenarios I verified) and is simply not reached when it **hits**.

On the **raise** path the kernel's own `.content` fill does run — confirmed with a marker colour visible under the blit — which is the other half of why filling the whole content app-side was the wrong shape: there it would have been a second fill of a rectangle the kernel had just filled.

### Where the garbage actually comes from

`sh_repaint` (`apps/sheet/sheet.asm:835`) does a bare `OSAPI_GFX_FILL` of SHEET's whole content plus `sh_drawall` with **no clip region armed**, and `sh_docmd_chart` calls it *after* the Chart window has been shown and painted. SHEET paints straight over a window sitting on top of it. The raise cache then banks that corrupted content, and the drag restores it faithfully.

So the sequence is: SHEET overpaints the Chart → the cache banks the corruption → the move replays it. That is a bigger and older defect than this PR, it has 34 call sites, and it is not something to fold into a fix for #142 — I've opened it separately.

### What #145 is, then

A correct fix for the paths where `W_PAINT` runs — which includes the open, the raise, `CHART.O88` standalone, and every drag of a window that was never covered — plus the removal of a double-draw and the same fix for `CHART.O88`. It is not the whole of #142, and #142 should stay open with a pointer to the new issue.

None of this changes the code on the branch. Your original diagnosis of the unwritten band was right; it just turns out to sit on top of something else.

https://claude.ai/code/session_013Ue18GLizEWCbqgvakYY7x

PR #146, opened 2026-09-03
Host tooling: guard the package overlay boundary, check the shared-include list, and two portability fixes

Koriban:
Four host-side changes, no guest code touched. Two are gates that catch a defect class nothing checked before; two are portability fixes to tools that only worked on one platform.

## `os88ovlchk`: the package overlay boundary

§68.10 rule 1 — *no near call across the boundary* — was enforced for the kernel and not for packages. A package's `.modc` is a different segment at run time, so a near `call` from module code into a resident routine lands at that offset **inside the module** and runs whatever is there. It assembles cleanly and there is no error anywhere; the symptom is a feature that silently stops working.

The walk now inlines a package's `%include`s the way NASM does, tracks which section each line lands in, and reports any `call`/`jmp` whose target is on the other side. `call far [vec]` is the way across and is not flagged.

Mutation-tested on this tree: a near `call wd_entry` inside WORD's `.modc` is reported; without it the row passes.

A package listed and not present is **skipped** rather than an error — a fork may carry one `main` does not — and the count printed is what was actually walked, so a skipped package cannot read as a checked one.

## `os88index`: the shared-include list is checked against the tree

`INCLUDES` is hand-written, and a hand-written list is exactly what goes quietly stale. `check_includes()` compares it against every `apps/os88*.inc` a package can opt into (`os88api.inc` excluded — it is not opted into, it *is* the API).

Mutation-tested: dropping an unlisted `apps/os88*.inc` into the tree fails the row with *"is in the tree and not in INCLUDES"*.

Worth saying that this gate fired on its own author. The change arrived from a fork carrying a ninth include; on `main`, where that file does not exist, it immediately reported its own list as stale. The row for the fork-only file is dropped here and the gate stays.

## `os88pix`: read a non-PNG on Linux too

The converter shelled out to `sips`, which is macOS-only, so any input that was not already a PNG failed on Linux with a missing-binary error rather than a useful message. Falls back to Pillow when `sips` is absent.

## `zpic`: find the Inform 6 compiler by either of its names

The Z-machine picture fixture looked for `inform` only. Debian and Fedora both ship it as `inform6`, so `make zpic` failed on a machine that had the compiler installed under the other name.

## Checks

`make test-fast`: 30 passed. No guest binary changes — `build/` untouched, nothing under it tracked (§16).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4877,15 +4877,38 @@ zgfx: zh zpic $(BUILD)/stories.stamp
 	python3 tools/zharness.py --all --graphics
 	python3 tools/zharness.py $(ZPICDIR)/zpictest.z6 --graphics
 
+# ...and the PICTURE half on its own. zgfx needs the story fetch, so on a
+# machine with no network - or when the question is only about the drawing
+# path - this is the part that can still run: the v6 fixture is the only thing
+# in the tree that asks for a picture at all (SPEC.md 61.7, 61.14), and it
+# needs neither a story nor a reference interpreter.
+zgfxpic: zh zpic
+	python3 tools/zharness.py $(ZPICDIR)/zpictest.z6 --graphics
+
 # The v6 picture fixture: a story that draws, and three flat blocks to draw.
-# Needs `inform` (`brew install inform6`), which is host-side only.
+# Needs an Inform 6 compiler, which is host-side only.
 ZPICDIR := $(BUILD)/zpic
+
+# WHICH IS NOT ALWAYS CALLED `inform`. Debian's inform6-compiler and Homebrew's
+# inform6 both install it as `inform6`; older and hand-built ones use `inform`.
+# This rule hard-coded the maintainer's name for it, so everywhere else it did
+# not degrade - it died as `make: inform: No such file or directory`, which
+# reads like a broken Makefile rather than a missing package. Look for both,
+# and let INFORM= name a third.
+INFORM ?= $(shell command -v inform6 2>/dev/null || command -v inform 2>/dev/null)
 
 zpic: $(ZPICDIR)/zpictest.z6 $(ZPICDIR)/zpictest.PIX
 
 $(ZPICDIR)/zpictest.z6: tests/frotz/zpictest.inf
 	@mkdir -p $(ZPICDIR)
-	inform -v6 $< $@
+	@if [ -z "$(INFORM)" ]; then \
+		echo "zpic: no Inform 6 compiler found (tried inform6, inform)."; \
+		echo "  Debian/Ubuntu: sudo apt install inform6-compiler"; \
+		echo "  macOS:         brew install inform6"; \
+		echo "  or name one:   make zpic INFORM=/path/to/inform6"; \
+		exit 1; \
+	fi
+	$(INFORM) -v6 $< $@
 
 $(ZPICDIR)/zpictest.PIX: tools/zpicgen.py tools/os88pix.py
 	@mkdir -p $(ZPICDIR)
--- a/Makefile
+++ b/Makefile
@@ -1385,7 +1385,7 @@ KERNEL_INC := $(wildcard kernel/*.inc) apps/os88ui.inc boot/boot2.asm
         bench field combo combo144 combo720 stackprobe trklog trkscrl npbench clicktest marty \
         comscan lptlink calcref \
         fonts fontsheets fontlist \
-        stories zdisk ztest zh zhboot zcheck zgfx zpic zscreens xt-z 386-z \
+        stories zdisk ztest zh zhboot zcheck zgfx zpic zgfxpic zscreens xt-z 386-z \
         worddisk wordcheck xt-word 386-word \
         cc-note chello covl cword cworddisk 386-c-word runcpm runcpmdisk \
         runcpm-src cpmsw rcz80test rcmemtest rczex 386-runcpm \
@@ -3371,7 +3371,8 @@ $(BUILD)/mppmove360.img: $(BUILD)/heapfrag.o88 $(BUILD)/modplug.o88 \
 # drive than ModPlug.
 $(BUILD)/zt/ZOPS.Z5: tests/frotz/zopstest.inf
 	@mkdir -p $(BUILD)/zt
-	inform -v5 $< $@
+	@$(INFORMCHK)
+	$(INFORM) -v5 $< $@
 
 $(BUILD)/zmove360.img: $(BUILD)/heapfrag.o88 $(BUILD)/frotz.o88 \
                        $(BUILD)/zt/ZOPS.Z5 tools/os88disk.py
@@ -4781,7 +4782,8 @@ ztest: $(ZTESTDIR)/gold3.txt $(ZTESTDIR)/gold5.txt $(ZTESTDIR)/gold8.txt \
 
 $(ZTESTDIR)/zopstest.z%: tests/frotz/zopstest.inf
 	@mkdir -p $(ZTESTDIR)
-	inform -v$* $< $@
+	@$(INFORMCHK)
+	$(INFORM) -v$* $< $@
 
 $(ZTESTDIR)/gold%.txt: $(ZTESTDIR)/zopstest.z%
 	dfrotz -w 80 -h 200 -p $< | grep -E '^(PASS|FAIL|TEXT|RESULT)' > $@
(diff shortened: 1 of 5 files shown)


jggonz:
Reviewed and pushed one commit (`5d73b1e`). I merged `main` into this locally to check the combined result — no conflicts, nothing to resolve — and the PR is `MERGEABLE`; it's only waiting on approval now.

Short version: **both gates do what they say, and neither portability fix regresses macOS.** I went looking for the "fixes Linux, breaks the maintainer's Mac" failure and didn't find it. Details below, because a gate is only worth what you can prove it bites on.

## What I checked

**The overlay gate bites.** I mutated it independently of your test — a `call wd_entry` inserted right after `section .modc` in `word.asm` — and got:

```
apps/word/word.asm:19592: .modc -> .text, near: wd_entry  (apps/word/word.asm)
os88ovlchk: 1 package call(s) cross a section boundary near - SPEC.md 68.10 rule 1
```

exit 1, right file, right line, right diagnosis; removing it goes back to exit 0. That is the whole point of the change and it lands.

**The INCLUDES gate bites both ways.** Dropping an unlisted `apps/os88zzz.inc` into the tree: *"apps/os88zzz.inc is in the tree and not in INCLUDES"*, exit 1. Moving `os88fp.inc` out: *"INCLUDES lists os88fp.inc, which is not in apps/"*, exit 1. Both messages name the file, which is the part that saves the next person ten minutes.

**Coverage is complete, and worth stating out loud.** `PKGS` lists only `apps/word/word.asm`, which reads thin until you check — WORD is the *only* hand-written assembly package in the tree with a `section .modc`. `LOOM` and `covl` also produce an `.OVL`, but they are C packages whose overlay boundary is already guarded by `tools/cc8086.py` (§73.14), and `apps/cc/crt0.asm` is the runtime rather than a package. So the list is right; it just doesn't look right at a glance.

**`expand()`'s include search matches what NASM is actually given.** The `word.bin` rule passes `-I apps/ -I apps/word/`, and `expand()` tries the including file's directory then `apps/` — same two places, and no filename exists in both, so the ordering difference can't bite.

**The label regex is the same one the kernel half already uses** (`^([A-Za-z_]\w*):`, lines 128 and 421). I checked every source WORD pulls in for an indented label that would slip past a column-0 anchor: there are none, and `-w+error` turns NASM's `orphan-labels` warning into an error, so a colon-less label can't exist either. Consistent with the file, and sound on this tree.

## The portability half, on an Apple Silicon Mac

This is where I looked hardest, since two of the four changes are Linux fixes and I'm reviewing on macOS.

**`zpic` — no regression, and it turns out this machine is the interesting case.** This Mac has `/opt/homebrew/bin/inform` and *no* `inform6`. So the detection falls through to the second name and finds it, and `make zpic` runs clean: `3 picture(s), release 7, 1920 bytes`. The old hard-coded `inform` also worked here, so nothing changed for me — but the ordering (`inform6` first) is the right way round, and the failure message names the package for both platforms instead of leaving `make: inform: No such file or directory` to be misread as a broken Makefile. That reading is exactly right; I've watched people go into the Makefile over it.

**`os88pix` — the macOS path is untouched, and I checked the fallback for a hazard you didn't claim.** `sips` is on PATH here, so the `if shutil.which("sips")` branch is taken and always exits through `return` or `fail()`; it can't fall through into the Pillow code. Fine.

The thing I wanted to be sure of is **determinism** — CLAUDE.md pins this toolchain as byte-reproducible, so a `.PIX` built on Linux and one built on a Mac had better be the same file. I built the same image as a GIF, a BMP, a greyscale TIFF and an 8-colour paletted GIF, ran each through the `sips` path and then again with `sips` hidden to force Pillow, and compared:

```
IDENTICAL  src.gif      IDENTICAL  src.bmp
IDENTICAL  grey.tiff    IDENTICAL  pal.gif
```

Byte-identical all four. The quantiser downstream washes out the colour-management difference between the two decoders, so the fallback is genuinely equivalent and not just close. Worth knowing, since it's the kind of thing that would surface a year later as an unreproducible release.

## What I changed — `5d73b1e`

One commit, ten lines deleted, no behaviour change. Two leftovers in `check_pkgs()` from what looks like the first shape of the double-include fix:

- `walked += 1` sits between two identical `os.path.exists(pkg)` guards, so the second can never be false.
- `last`, `first_of` and `dup` are built with two full passes over the expanded stream plus a scan per file — and `dup` is then never read. The dedup that actually runs is the `seenfile` pass right below it, which is also what your comment describes: *"the code-bearing include is the LAST one"* is `seenfile[(f, n)][-1]`, not this block. On WORD's 33,043-line expansion `dup` evaluates to the empty set.

Nothing else moved, and the comment above the removed block still describes the pass that remains, so it didn't need rewording. I re-ran the same mutation at the new commit and got the identical diagnostic and exit code — the gate is unchanged, there's just less of it.

## Flagged, not fixed

Neither of these is a defect today; they're both places where the next edit could quietly go wrong.

1. **The dedup keeps both copies when a file is included twice into the *same* section.** `len(seenfile[(f, n)]) > 1 and sect != seenfile[(f, n)][-1]` skips a row only when the sections differ, so two `.text` inclusions both survive. Harmless for the label map (same label, same section), but a genuine bad call inside such a file would be printed twice. No file in the tree does this, so I left it alone rather than touch working dedup logic.

2. **"The code-bearing include is the LAST one" is an assumption, not a rule.** It's true for `os88img.inc` and its `OS88IMG_CONSTS_ONLY` guard (§85), and since this deliberately doesn't evaluate `%ifdef` there isn't a cheap way to do better. But a package that ever includes code first and constants second would have its labels mapped to the wrong section, silently and in the safe-looking direction. Might be worth a line in the comment saying the assumption is load-bearing.

## Combined-tree check

Merged `main` in locally (it had gained `79a436f` AUDIO and `665f5c0` the chart-window fix) and ran the pre-merge gate on the result:

- `make` — exit 0, fast tier 30 passed / 0 failed
- `python3 tools/os88test.py full` — **33 passed, 0 failed**, 205s (the 4 skips are this box lacking the C toolchain and MartyPC, not failures); `buildmatrix` green, so the knob kernels and `kern_small` still assemble
- `os88disk.py --verify` on all seven images — clean
- `git ls-files build` — 0, and no binaries or `.claude/` rode in
- no duplicate `§` headings

The two doc claims check out against the source too: `OS88UI_SBDRAG` is real in `os88ui.inc`, and `os88chart.inc` does have Scatter and Combination, so "seven" is right where "five" wasn't.

Nice piece of work — the overlay gate in particular is the good kind, where the thing it catches has no other symptom. The two false positives you hit and documented in the commit message (`.cold` not being a second address space in a package, and a file included twice being one file) are the part a reader will thank you for; they're the reasoning that's normally lost.

jggonz:
Follow-up, and a correction to my comment above.

Two reviewers I had running came back after I'd posted, with findings I'd missed. I re-derived each one myself rather than take them on trust, and **three were real**. I've pushed two more commits (`ca86b2c`, `8cf804f`). Everything still green — `make` exit 0, `os88test.py full` 33 passed / 0 failed.

## Correction to what I said above

I claimed the Pillow fallback was byte-identical to `sips`. **That was incomplete, and the incompleteness was in my test, not your code.** I compared flat-colour synthetic images — a GIF, a BMP, a greyscale TIFF, a paletted GIF — and all four genuinely are identical. A *photographic* JPEG is not: I measured 20 differing bytes of 14,032 on a 200×140 q75 image, the chroma-upsampling difference between the two decoders surviving the 16-colour reduction.

So the comment at `os88pix.py:158-160`, "nothing non-deterministic reaches the archive", is now true only *within* one host. The blast radius is genuinely small — the only archive that reaches is `BRONZE.PIX` from Bronze.zblorb's JPEG cover, which lands only on `build/zork2.img`, built on demand from stories that are never committed and compared by nothing. No code change; I've made the comment say it.

## The one that mattered: the dedup was hiding a real crossing call

This is the finding I'm glad came in late rather than not at all.

The double-include dedup keeps the **last** copy of a file, on the reasoning that the code-bearing include comes last. Reverse the order and it drops the wrong one. I built the fixture: a `lib.inc` containing `lib_fn: call resident_a`, included into `.modc` first and `.text` second. Its `.modc` rows get discarded, so the near call — a genuine `.modc → .text` crossing, the exact bug this walk exists to catch — is reported **clean, exit 0**.

```
before:  os88ovlchk: 1 package(s) keep every overlay call far
after:   lib.inc:2: .modc -> .text, near: resident_a
         os88ovlchk: 1 package call(s) cross a section boundary near
```

And it isn't buying anything here. Its justification names `apps/os88img.inc` and `OS88IMG_CONSTS_ONLY` "(SPEC.md 85)" — that file isn't in this tree, nothing includes it, and §85 is TANK ATTACK. I walked WORD's 33,043-line expansion and counted: **no file in a package's expansion is included twice at all.** So the block is gone (`ca86b2c`), both copies are now judged in the section they actually land in, and the comment records that a future dedup would have to keep the `.modc` copy rather than the last one.

I suspect this is a fossil from a different tree — the same shape as the `scribe.asm` entry and the `os88chartbss.inc` row you already caught and dropped. It's the fourth thing in this PR that came across from a fork where it was true.

**Also in `ca86b2c`:** `walked == 0` printed `0 package(s) keep every overlay call far` and exited 0, so renaming or moving `word.asm` would have silently disabled the package half of the gate — the same "a skip must not read as a check" you fixed for the count in `406343b`. It refuses now.

## The `inform` fix reached one rule of three

`8cf804f`. There are three rules in the Makefile that compile Inform source, and the fix landed on one:

- `Makefile:3374` — `$(BUILD)/zt/ZOPS.Z5`, which feeds `zmove360` and `tests/editmove.py --app frotz`
- `Makefile:4784` — `$(ZTESTDIR)/zopstest.z%`, which is `make ztest`

On the Debian machine this change is written for, both still die with exactly the `make: inform: No such file or directory` you set out to remove. Both go through `$(INFORM)` now, and the refusal is a shared `$(INFORMCHK)` rather than a third copy of the five-line message — it names `$@`, so it tells you which target wanted the compiler.

**And the comment had Homebrew backwards**, which is worth more than the fix itself because it's the reason anyone would trust the ordering. It says Homebrew's inform6 installs as `inform6`. On this Mac, `brew list inform6` (formula 6.44-r7) installs `/opt/homebrew/bin/inform` and `inform-6.44` — **no `inform6` at all**. So macOS fires the *fallback* name and Debian the preferred one: the two platforms want opposite names, which is a considerably better argument for your change than the one written down. Reworded.

Small one in the same commit: `zgfxpic` joins the `.PHONY` list its whole family is already on (`zh zcheck zgfx zpic zscreens`, Makefile:1388). Not a live bug — its prerequisites are phony so the recipe runs anyway — but it'd stop the day a file of that name existed.

## Flagged, still not fixed — this one needs your call

**`expand()` drops an unresolvable `%include` silently**, at `os88ovlchk.py:517-524`: the `for cand in (...)` loop has no `else`, so if neither candidate path exists the line is discarded with no message. Every label in that file is then absent from `where`, `where.get(tgt)` returns `None`, and every near call into it is skipped while the gate prints success.

There is exactly one such include in the tree today — `apps/word/word.asm:19333`, `%include "wdbench.inc"`, and the file isn't present. That one is *harmless*, because it sits inside `%ifdef WDBENCH` and NASM doesn't include it either.

Which is exactly why I didn't just fix it: the obvious `else: report and fail` would fire on `wdbench.inc` on every build. Telling a guarded-out include from a genuinely missing one means evaluating `%ifdef`, which this deliberately doesn't do. The options I can see are a warning to stderr that doesn't fail (noisy, and a warning nobody can clear gets ignored), tracking `%if`/`%endif` depth just far enough to skip guarded includes (more machinery, but it would also fix the `os88ui.inc` misclassification below), or leaving it and writing the assumption down. That's your design call, not mine to make inside your PR.

Related and lower stakes, both shared with the pre-existing kernel half so arguably house style rather than defects: `section` is matched but not `segment` or `[section .modc]`, and the label anchor is column-0 so an indented `label:` is invisible. Both fail *open*. Neither form occurs anywhere in the tree, and `-w+error` turning NASM's `orphan-labels` into an error means a colon-less label can't exist, so there's nothing live here.

One more worth knowing: `%ifdef` not being evaluated means `os88ui.inc`'s compiled-out `section .cold` misclassifies about 3,300 lines of the stream. It's harmless *only* because `fold` collapses `.cold` into `.text` — a guarded-out `section .text` sitting inside `.modc` would fail open the same way. That's the same root as the include question above, which is part of why I think it's worth one decision rather than three patches.

## Standing

- `make` exit 0; fast tier 30/30
- `os88test.py full` — 33 passed, 0 failed (4 skips are this box lacking the C toolchain and MartyPC)
- all seven images verify; no duplicate `§`; `git ls-files build` = 0
- the WORD mutation still bites at HEAD: `apps/word/word.asm:19592: .modc -> .text, near: wd_entry`, exit 1
- all three Inform rules build for real here, and `INFORM=` gives the refusal

Nothing above changes my view of the PR — the overlay gate is the good kind, and the dedup bug is the sort of thing you only find by building a fixture that tries to defeat it, which is precisely what the gate now protects everyone else from having to do.

PR #147, opened 2026-09-04
Elendilon -> Main (Kern Big Shrinks 12KB, Kern Small is Real, .o88 Package Parts, 12 Usable Worker Stack Slots, Cursor Unfrozen in Disk I/O)

Elendilon:
# Elendilon -> Main (Kernel Shrinks 12KB, Kern Small is Real, .o88 Package Parts, 12 Usable Worker Stack Slots, Cursor Unfrozen in Disk I/O)

The elendilon integration branch, squashed onto main, with main's own
Hibernate (#143) and Audio Player (#144) merged underneath rather than
overwritten.

CLAUDE.md is main's, wholesale. The integration branch carries a
branch-specific "Working in this fork" section that is the fork owner's
standing preference rather than a property of the project; taking main's file
drops it along with the pointer to it. Its knobs table is the one edit — this
branch has 27 build knobs where main documents 12, and all 12 keep main's
wording.

---

## The headline: the kernel is 11,776 bytes smaller

Measured as **main today against this branch today**, so main's own spending
cancels: both trees carry Hibernate and the Audio Player, and neither is part
of this number.

| `kern_big` | main | this branch | |
|---|---:|---:|---:|
| `.text` | 57,696 | 50,704 | **−6,992** |
| `.bss` | 6,049 | 6,003 | −46 |
| `.cold` | 41,168 | 37,154 | **−4,014** |
| `.lowbss` | 8,982 | 9,182 | +200 |
| **`KERN_SIZE`** | **121,344** | **109,568** | **−11,776** |
| `.text`+`.bss` (the 64KB window) | 63,745 | 56,707 | **−7,038** |

`KERN_SIZE` is the whole resident footprint — what the kernel costs every
machine before a single package loads — so this is 11.5 KB of RAM handed back
to the heap on every boot. The 7,038 bytes off `.text`+`.bss` matter for a
different reason: that guard is the 64KB segment, offsets are 16 bits, and it
cannot be raised by anyone.

Three things paid for it. A line-by-line size pass over all 44 kernel files;
a second half of the boot overlay (`.ovlw`) for bodies that are dead after the
first mount, which trades reclaimed memory for resident memory and is why
`.ovl`+`.ovlw` is *larger* here while `KERN_SIZE` is smaller; and moving
features out of the kernel image entirely — `SYSTEM.CFG`'s parser and the
file-association cache both stopped being resident.

The `.lowbss` row is the one thing that went up, and deliberately: it is the
worker stacks below.

## `kern_small` is 28.5 KB lighter, and the 128KB machine is real now

| `kern_small` | main | this branch | |
|---|---:|---:|---:|
| `.text` | 48,678 | 39,218 | **−9,460** |
| `.cold` | 37,900 | 25,853 | **−12,047** |
| **`KERN_SIZE`** | **107,008** | **77,824** | **−29,184** |

Free heap is whatever the ladder leaves, so that drop *is* the heap gain, byte
for byte. Booted on an actual 128KB machine — not arithmetic, an 86Box/MartyPC
profile with 128KB in it — the floor machine reaches a desktop with **50.5 KB
of free heap**. `MIN_RAM_KB` had been an unmeasured claim since the day it was
written; `tests/small128.py` is now the gate, and it checks the desktop is
really there before it believes the number.

Most of that came from deciding what a 128KB machine does not need rather than
from shaving. Cut/Copy/Paste and the Standard File dialog became on-demand
modules that ship as files and are read in when asked for (`FILECP.DRV`,
`FDLG.DRV`); file associations are gated out; and several packages gained a
small build — Note Pad 20,379 → 13,520 bytes an instance, Paint 31,352 →
26,465, the Calculator 7,351 → 5,563. It is not a second ABI: a small-built
package calls the same API table and runs on `kern_big` too.

The small disks also stopped carrying what could never run there. Eight
packages need a driver `kern_small` does not ship, so they are not on the
floppy at all — a program that cannot reach its driver cannot even refuse in
its own words.

## Packages can have parts (SPEC.md 20.12)

A `.O88` may now carry more than its own segment: named parts inside one file,
sized, claimed, loaded on demand, optionally into XMS, and refused with an
arithmetic the program states itself. A package over 64KB is still a package.

The first consumer is the C64 emulator, whose 20 KB of KERNAL, BASIC and
CHARGEN used to be a `C64.ROM` sidecar — a file a copy could separate from the
program, which made a machine that could not start. It is part 0 of `C64.O88`
now.

**The whole kernel bill is 85 bytes** (`.cold` +70, `.bss` +15). Everything
else — sizing, refusing, claiming, XMS, lazy loading, the C SDK's support — is
package code in `apps/os88parts.inc`, which is the point rather than a
coincidence: an earlier kernel-side version of this was built to five waves,
measured at 2,560 resident bytes, and thrown away.

## Thirteen worker slots where there were seven, for 128 bytes

A task stack used to be a flat 384 bytes because that was the deepest worker.
It is a class now — 128, 192, 256 or 384 — looked up rather than derived, and
a task asks for the smallest that fits.

| | main | this branch |
|---|---:|---:|
| worker slots | 7 | **13** |
| bytes of slice | 2,688 | 2,816 |

Nearly double the concurrent workers for 128 bytes, because the sizes came
from walking every shipped worker rather than from the worst case: 192 is the
modal class, eleven of the twenty land there.

What made the small classes safe is that a slice is not sized by what the task
*does*, it is sized by what *lands on it* — and about 82 bytes of every stack
belonged to no program at all. The ROM's `int 08h` chain and both mouse ISRs
now run on shared private stacks in `.lowbss` instead of on whichever task they
interrupted. That is the `+200` bytes in the table above, and on the bench it
takes the floor a slice is cut from **82 bytes down to 32**.

## The pointer keeps up with your hand during disk I/O (SPEC.md 7.4)

A file operation freezes the machine, and the arrow used to freeze with it —
then, once an operation had moved three sectors, **leave the screen entirely**
for the rest of it. Copying a file meant watching a still picture with no
cursor on it.

The CPU inside a BIOS disk call is not busy: it has handed the transfer to DMA
and is spinning on IRQ6, with interrupts enabled the whole time. So the mouse
ISR draws the arrow in a gap the machine was going to spend idle. It is
*tracking*, not merely un-hiding — a lit-but-frozen arrow was shipped here
once and reported back from the field as a stutter, so the only version worth
building is one that follows the hand.

**180 bytes**, and the knob build is byte-identical to the kernel before it.
`NOCURDISK=1` is the A/B and `tests/curdisk.py` the gate: 28 and 37 pointer
moves across two scenarios, against 0 and 0.

## Paint's canvas can hold one bit a pixel (SPEC.md 42.23)

A third canvas format, and the first that is about the document rather than the
card. On a monochrome adapter the picture *is* one bit, so storing four was
costing four times the claim for nothing.

| canvas | 4bpp | 1bpp |
|---|---:|---:|
| Hercules default 448×258 | 56.6 KB | **14.2 KB** |
| what 13.5 KB of heap buys at 448 wide | 61 rows | **245 rows** |

That is the difference between a letterbox and a picture on the floor machine.
Storing it `1 = white` agrees with the BMP palette convention *and* takes
`rep movsw`'s 12.5 clocks a byte instead of a complementing loop's 17 — the
intuitive polarity would have cost 36% of every blit forever.

---

## Also in it

- **A straddled window no longer loses a character of unaligned text**
  (39.14.11) — reported from the field. A display seam is a multiple of 8, so
  aligned text lost nothing and unaligned text lost exactly one cell: the cell
  the seam crossed was dropped whole. It is cut at the seam now, and because
  the cut lands where both renderers already split, neither gains an
  instruction.
- **`SYSTEM.CFG` costs 2 resident bytes instead of 14**, the parser having
  moved into the boot overlay — worth 1,024 bytes of every machine's RAM.
- **The software renderer's plane loop is gone** on 1bpp adapters (39.26),
  where it was drawing plane 0 four times.
- **The Audio Player ships at 360KB** (24.4.1). Main left it off that geometry
  on the arithmetic that it fitted with one cluster to spare; with this
  branch's packages on the disk it does not fit at all — `os88disk` refuses the
  build at 360 clusters against 354. MODPLUG and RECORDER come off that
  geometry alone and AUDIO goes on: 337/354, seventeen clusters spare. Tracker
  stays, being the only one of the four that authors anything. The 1.44MB,
  1.2MB and 720KB disks are untouched, and AUDIO is on neither small apps disk
  — `kern_small` ships no `SOUND.DRV`.
- **`.modh` is registered with `os88ovlchk`.** Hibernate's module section was
  never added to the overlay gate's list, so every label in it was filed as
  `.text` and a near call from the module into the kernel — the exact defect
  that gate exists to refuse — read as legitimate and passed in silence.
  A/B'd on an injected call both ways. The tree is clean with it registered.

## Testing

`make test-full` passes: 52 rows, 2 skipped for want of the optional C
toolchain, including `kern_small` booting to a desktop on both 1bpp adapters,
the knob-kernel build matrix, and the resident-memory gate. The soak rows for
everything above pass, and main's own `hibernate` and `hibernatedrv` rows pass
on the merged tree, resume included.


jggonz:
I merged `main` (#143 Hibernate, #144 Audio Player, #145 Chart, #146 host tooling) into this branch and pushed five commits on top of yours: the merge (`8463342`), the adaptations that make the merged tree build and pass its own gates (`799c5a9`), a batch of fixes from a twelve-unit review (`5ee929f`), the SPEC/plan updates that follow them (`33a1ee3`), and one test follow-up (`70c54d0`). The PR is mergeable now. Nothing of yours was reverted or rebased.

This is a big piece of work and the review reflects that: most of what twelve reviewers and ten adversarial verifiers read came back clean, and the clean lists are worth having on record (the ISR private stacks cannot nest, the seam cut is exact in both renderers, the DMA staging repair holds, the loader's new size fence admits nothing the file cannot hold, the 1.2MB geometry is right end to end, every boot-sector knob combination still fits). What follows is what did not survive, and what I did about each.

## First, three things about the PR as pushed

- **The head you pushed (`0d610c66`) did not carry the merge the body describes.** It has no `kernel/hiber.inc`, no `apps/audio/`, no `.modh` in `os88ovlchk`, and no SPEC 24.4.1. The body's size figures were measured on a tree that never reached the fork. I did the merge instead (commit `8463342`); Hibernate took module index 3 on both kernels with main's kern_small stub, and your `MOD_FCP`/`MOD_FDLG` moved to 4 and 5. The 360KB apps disk keeps main's arrangement (AUDIO off, MODPLUG and RECORDER on) and fits at 344 of 354 clusters, so the "does not fit at all" sentence in the body is not true of this tree.
- **The head did not assemble on NASM 3.02.** `kernel/mod.inc`'s `add di, mod_fp - mod_tab*7` is a label times a constant, which NASM 3.02 refuses as an invalid operand in every spelling. Your NASM accepted it, so your gate was green and mine was not. `MOD_TAB_OFF equ mod_tab - $$` is the same address as a number (`.text` has vstart=0), and `mod_fpr` multiplies that instead (`799c5a9`).
- **Your new stack-balance gate flagged main's code**, correctly: Hibernate's `hbm_perform` enters its stub with push/push/`retf`, and the Audio Player has four itoa loops. Both now carry the `STKBALANCE-OK` / `STKBALANCE-LOOP` markers the gate asks for. That is the gate working as designed.

## What I fixed (commit `5ee929f`)

Each one: what a user would have seen, then the mechanism, then the change.

**The pointer-and-menu machinery**

- *For 59 seconds of every minute the UI believed its event queue was never empty.* `evq_count` became a byte in your size pass, and two readers in `ui.inc` still compared it as a word, which pairs it with `clk_sec` beside it in `.bss`. Every idle pass spent all 16 drain turns on an empty ring, and the lost-release safety net in `ui_arm_chk` never ran. Both compares are `cmp byte` now.
- *A dock tile could draw another file's icon.* `inst_icon_ptr` was made the "fifth tenant" of `dsk_ico` on the claim that all five run on task 0 - but `dock_draw_tile` is reached from `wm_destroy` on a dying package's worker, while task 0 can be inside a mount's icon pass with no lock held (`osapi_file_goto`). It has its own 64-byte buffer again. That costs kern_small one 512-byte image rung (17 bytes were left in it); if you would rather recover those bytes elsewhere, that is your call and the reasoning is in the commit.
- *A package could corrupt the API jump table through the new `OSAPI_WM_NOANIM` slot.* It set a flag bit through whatever BX it was handed, twelve lines from `wm_saveu`, which documents exactly that incident. It takes the same range test now.

**Stacks**

- *A second tick arriving inside the ROM's `int 08h` chain ran the sound driver's tick and the splash animation on the 128-byte private chain stack*, 58 bytes on top of the chain's own 56, with no canary under them and the mouse ISR's stack directly below. A nested tick now skips those two calls and keeps the accounting and wake scan SPEC 8.5.1 promises. This cannot show under QEMU (SeaBIOS hops stacks for IRQ0), which is why nothing caught it.
- *`sch_switch` could park a ROM's stack pointer into a task record.* SeaBIOS serves `int 16h` AH=01h, which `ui_task` polls with the lock clear, on its own stack, and the new task-0 canary read would then go through that segment. The switch now declines a tick when SS is not `LOW_SEG`, the same test your chain swap already makes. Main had the same window without the canary read, so this is a hardening rather than a regression, but it is the first time the window is survivable.
- I measured the 128-byte class with `SOUND.DRV` loaded and its tick armed, since the plan's floor was taken without a driver: the idle task used 48 bytes and a ticking Timer 58. The classes hold.

**Disk**

- *At one kernel size in 128, every mount would have failed with "Disk error".* Narrowing the listing stride from 32 to 24 moved `dsk_secbuf` to `LOW_SEG+2816`, which is 256 bytes into a sector, so an `int 13h` read into it straddles a 64KB DMA page at exactly one position per page. Nothing in the tree asserted the alignment; guard 6 checks segment bases only. The buffer is the window's first bytes now, and a `%if` against the rung base holds it. To be fair to the branch: main's buffer was misaligned too, at a different residue. What the branch changed is that kern_small sat 9 KB of shrink from the fatal position where main's sat 42 KB away, on the arm you just took 29 KB off.
- *A Cut/Copy/Paste "replace?" question you walked away from left the paste suspended for good*: the copy buffer (up to 64 KB) claimed, `fcp_busy` set, and on kern_small `FILECP.DRV` held. `fm_edit_end` had arms for modes 5, 6 and 7 and none for 4. It answers "stop" for an abandoned question now, and a second question of the same paste (two conflicting names) is no longer read as an abandonment - the first version of this fix would have dropped the module under a live paste, which the verifier caught.

**The parts standard**

- *A lazy part fetched on any floppy with 1KB clusters could never be dropped.* `op_fetch` banks the part's base past the head slack so `op_seg` can answer it, and `op_drop` handed that biased word to `OSAPI_MEM_FREE`, which matches the claim base exactly and refused; nobody read the CF, so the claim leaked until close. `tests/mseglazy.py` could not see it because its fixture's lazy part happens to land on an even sector. `op_drop` runs the slack arithmetic backwards now.
- *A package whose last file-backed part is not a whole number of sectors was refused at every launch* with "Cannot read my parts": `op_want` asked for sector padding the packer never writes after the last body. The C64 escapes because its ROM is exactly 40 sectors. `op_size` remembers the last row's padding and `op_want` no longer asks for it. `OP_BSS` went 71 to 73; the mirror tracks it.

**The small builds**

- *The small Calculator wrote past its own bss on every click.* `cal_layout` still laid out the history pane's nine rects into a table `APP_SMALL` had shrunk to the keypad's, 72 bytes past the array and 46 past the instance's bss. The block is under `CALF_HIST` like its siblings.
- *The small Note Pad's Edit menu read its fifth item out of the string "Cut  ^X".* Five declared, four supplied. The count is gated with the list.
- *`make smallapps` built a floppy kern_small refuses to mount.* The 1.44MB apps disk lacked the `--fatcap 2` its paired system disk gets, so mount rule 10 refused its 9-sector FAT and B: came up "No os8088 disk". Nothing caught it because both small rows mount the 360KB pair. Fixed, and `AUDIO.O88` (arriving from main) is on the small-disk omit list with the other three sound packages.

**Packages**

- *Chart's and Fractal's About cards drew off arbitrary memory on any repaint.* Chart's `ct_paint` spends SI on the blit's source offset (118) and then handed that SI to the card as its window; Fractal's `fr_redraw` clobbers BX and the card never reloaded it. Every About card in the tree is new in this branch, so the verifier swept all eleven: the other nine are right.
- *Backspace in Paint's text tool restored the wrong pixels on a one-bit canvas.* `pt_upix` read the undo image as packed nibbles whatever the canvas format. It decodes through `pt_getpx`'s own three arms now - which also fixes the planar case that predates this branch.
- The C64 no longer frees the ROM segment on its two refusal paths; it is the parts loader's claim now. The self-extracting `OS88NET.COM` clamps a match copy to what the image still owes, so a damaged file cannot run the decoder over itself. Two stack-class notes now state the arithmetic of the class they declare.

**Gates and build**

- `buildmatrix` and `bmshare` both drive `make` against `build/` and both ran in the shared parallel lane while 47 rows read the build; a `builds=True` row leaves that lane now whatever `serial` says.
- `t_stkclass` read only a package's top-level `.asm`, so TANK and BROWSER - the two packages that declare classes - were never sized and the gate said nothing. It reads the `.inc` files too (13 workers measured became 16) and reports a spawner it cannot size.
- The `RAMKB` matrix rows target the boot sector, which is the arm the knob reaches; `NOHEDGE` no longer puts a budget exemption on a kernel it does not touch; `tests/curdisk.py` builds its knob arm into `build/curdisk` instead of over `build/` with an `atexit` restore a kill never runs; `tests/fdlgdrop.py` reads `MOD_FDLG` out of `mod.inc`; `.modh` has the section-tail guard its siblings have; `t_lowwin` asserts the sector buffer's alignment.

## SPEC and the plans (commit `33a1ee3`)

Where the documents and the code disagreed, the code was the fact: §8's `MAX_TASKS` and the slice layout, §20.6 rule 6's flat 384 (a worker's stack is the class its header declares), §20.2's byte +15, §20.12.3's 6-byte parts row (it is 8), `OP_SCRATCH` (it is `OP_ZERO`), §2.8.1's `MOD_NENT` 6 (it is 7), §7.4.4's paragraph saying the driver path is never bracketed (it is), the stage-1 RAM floors (115 KB and 84 KB, derived), and Paint's phantom `pt_lit`/`PT_LIT16` names.

## Flagged, not fixed

These are yours to decide; each is a design choice or a measurement rather than a repair.

- **Flood-filling with the grey swatch on a one-bit canvas re-scans rows it has already filled** (Paint). The dither ink keeps half the run at the terminator colour, so `pt_frow`'s shoulders push spans inside filled rows; a shoulder across a 448-wide canvas is about 224 redundant `GFX_FILL` calls (~170 ms). The picture comes out right. Refusing the swatch would delete a capability 42.23.1 argues for, so I left it; greying the fill tool alone for that ink, or documenting it beside the converse case 42.23.1 already blesses, are the two honest options.
- **A one-bit canvas whose width is not a multiple of 8 loses the fast blit on every adapter**, because the width check in `pt_blit_1` is all-or-nothing; a 466-wide picture opened from a file repaints one blit call per row (~227 ms) instead of two banded calls. The fix is a two-pass restructure, not a hunk.
- **The small Paint's repaint is one blit call per row on the machine it was built for** (kern_small has the slot and not the body). Not a regression - the small build has no prior number - but PERFORMANCE.md Part 5 is owed a row.
- **The parts standard's grant check omits the head slack and the cluster rounding**, so on a 1KB-cluster floppy it can claim 1 KB more than it checked and shed the caches 20.12.4 says it will not; the claim itself is right, so the cost is a correct refusal. And the packer's 128-sector carve bound is slack-blind: the runtime ceiling is 124 sectors at 1KB clusters, 112 at 4KB, and the packer admits 127.
- **`op_xload` never counts the bytes a read delivered**, so a truncated OP_XMS span loads as success with its tail from uninitialised staging. No shipping package declares OP_XMS and no 8088 has a store.
- **A parts refusal no longer names the package.** The C64 used to refuse in its own words; `op_say`'s four bare strings do not say which program could not start. Composing the name in costs the toast's 24-character budget, so the reasons would need shortening.
- **A disabled glyph cut by a horizontal display seam restarts its checkerboard phase at the seam**, so an odd cut flips the stipple for that cell's far half (two displays, VID_L_BELOW, 1bpp).
- **The full tier's declared seconds (700) exceed its 600 s ceiling**, though the wall clock fits today (381 s here, with MartyPC staged); the serial lane alone declares ~500 s. Whether to raise the ceiling or move a row down a tier is the decision the runner's own message describes.
- **`tests/msegnomem.py` fails on this machine, and it fails identically at your pushed head `0d610c66`** (built with only the NASM fix applied): the toast after MSEGBIG's refusal reads `Not enough memory`, the parts loader's own string, where the row asserts the kernel's `Load failed` from SPEC 21 step 10 has replaced it. Either the loader's step-10 toast is not landing over a live toast, or the row asserts the race its own comment says it avoids. I did not change either; it is worth a look, because the body says the soak rows for the parts standard pass.
- **`tests/msegnomem.py` builds its `DISKCNT=1` kernel over `build/` and restores with a plain `make` only on a clean exit** - the same shape I moved `curdisk` off; a row that dies mid-way (mine did once, on a double-click the emulator missed) leaves every later row booting a counted kernel. The `BUILD=build/<row>` arm `curdisk` uses now is the pattern.
- **`tests/curdisk.py`'s `LIT_MIN` is applied to the launch scenario**, which the plan itself says it cannot separate; it can only ever produce a false failure there.
- Comment-level things I did not touch: `STK0_SIZE`'s justification sentence, `SCH_PARTITION`'s 192-class note naming the Task Manager, `tests/ovlrefs.txt`'s rows that argue `.ovlw` bodies against `spl_finish` instead of the mount, and the `.claude/` LESSONS edits (two factual lines, kept).

## What I checked and cleared

`make`, `make small`, `kernsize --bless` both variants, `os88disk --verify` on all eleven images, `make test-full` (53 passed, 2 skipped for the optional C toolchain, 381 s of a 600 s budget) with MartyPC staged for this branch (`make marty`; a symlink to another checkout's staging fails five rows, which is what the first run showed), the soak rows for every subsystem I touched (multiseg, mseg360, mseglazy, msegxms, pkgbig, pkgfence, fcpcopy, fdlgdrop, curdisk, paint1bpp, paint1blit, paint1load - all pass; msegnomem is the one above), the merged desktop and System menu on VGA and CGA, Hibernate greyed "(No HD)" in the right slot, kern_small mounting the 1.44MB `smallapps.img` in B: (four folders, 1140K free), the stack watermarks with a sound driver, and the changed packages assembled in both arms with the shipping arms byte-identical to what you built.

https://claude.ai/code/session_01UsfsKifrD4PGB3Bw23Jrrn

jggonz:
I read the commit you pushed after the review (`91db9e79`, SPEC.md 24.5 and `os88sym.py`), rebuilt the head in a clean worktree, and pushed one commit on top of it (`a2fdaf4e`). Nothing of yours was changed; the PR is still mergeable.

**What your commit got right.** The eight omitted packages sum to exactly 99,749 bytes on this build, the 360KB apps disks come out at 344 and 234 of 354 clusters, and a symbol in `.modp` or `.modd` now gets the same "on-demand module" answer `.modc` and `.modh` already gave. All three figures and the reader's new behaviour check out.

**What the one commit I added does.** All host-side tooling; no kernel byte or shipped disk moves.

- *The symbol reader was still one section short, and it is this branch's own section.* Your comment says `SECTION_SEG` is now the union of both builds, but `.ovlw` — the first-mount half of the boot overlay that SPEC.md 2.5.2 introduced on this branch — had no entry. `main` has no such section, so its reader never needed one. On the PR head `python3 tools/os88sym.py --all` died on the first `.ovlw` row of either kernel, and `linear()` on any of the 246 boot-overlay labels (`mouse_init`, `desk_init`, `wm_init`...) raised the same "add it rather than assuming KERNEL_SEG". SPEC.md 2.5.2 pins `.ovlw` to `FAT_SEG`, so that is the entry now. The listing runs end to end on both kernels, and `mouse_init` resolves to `1760:0232` on kern_big, which is `FAT_SEG*16 + off`.
- *The Boot Ladder's self-check refused on the merged tree.* `python3 tools/os88ladder.py --selfcheck` said "kmain calls nothing on the page knows about: hb_probe_x". `main`'s Hibernate (#143) put the resume probe in `kmain` between `drv_boot` and `xm_boot`, and my merge carried the call without the stage, note and title the page requires of every phase. The tool has no suite row, so nothing caught it. The probe is in the mount stage now, where `kmain` puts it (SPEC.md 87.5). A real measurement under MartyPC then regenerated the page with the probe as its own phase, under a millisecond on a floppy-only machine, which is what §87.5 says it costs.
- *The Makefile comment above `$(SMALLOMIT)` still quoted 90,510 bytes and seven programs.* Same figures as your SPEC change, in the file the list lives in.

Verified in the worktree: `make` (43 fast rows, checkdocs clean), the symbol listing on both kernels, `os88ladder.py --selfcheck` and a full ladder run. Two instruments worth adding to whatever you run before a push, since neither has a suite row: `os88sym.py --all` on both kernels whenever a new `section` appears, and the ladder's `--selfcheck` whenever `kmain` gains a call.

https://claude.ai/code/session_01UsfsKifrD4PGB3Bw23Jrrn

PR #148, opened 2026-09-04
Feature/vmmouse absolute

stuaxo:
Via Claude.

Hooks into the vmmouse driver + absolute mode which Qemu and V86 support.
Checks that the correct sequence of bytes is sent before enabling since it does use 386 instructions.

Tested in qemu and v86.

This does make it a lot less of a pain to use - capture mouse isn't needed, having said that I did ask it to make sure that toggling capture on and off works too.

For V86 I used the test harness it built locally.


jggonz:
Thanks for this — it's a genuinely nice thing to have, and the browser build is much less awkward with it. I've merged `main` in and pushed two commits, so the PR is up to date and mergeable again.

The short version: **the feature stays, the 386 code moved out of the kernel into `VMMOUSE.DRV`.** Everything you worked out about the backdoor — the protocol, the packet framing, the READ_ID/ABSOLUTE dance, the contest, the v86 8042 wake — is intact and is now in `drivers/vmmouse/vmmouse.asm`. What changed is where it lives, plus five bugs found on the way.

---

## Why it moved

Two reasons, and the second is the one that actually decided it.

**1. It cost every XT 512 bytes of RAM for code it can never run.** `tools/kernsize.py` measured the resident version at 548 bytes of `.text` and printed its own alarm — *"the image rung CROSSED: 512 bytes of every machine's RAM, gone"*. `main` had just merged #147, which spent a lot of work shrinking that kernel by 12KB. Worse: post-#147 `main` has only 32 bytes of headroom left in that rung, so 548 bytes spills into a *second* one. The rewrite is +259 `.text` and +124 `.cold`, which fits in one.

**2. A `[cpu_tier]` gate can't survive hibernation, and this is the part your instinct about the danger was right about.** Your gating was genuinely sound as far as it went — I checked every path and nothing in normal operation can reach the island on an 8086. But §87's hibernate writes the machine's RAM *wholesale* and validates five things on resume: magic, build number, module stamp, `mem_top` and `vid_kind`. **Not the CPU.**

So: hibernate on a 386. Move the disk to an 8088 with the same kernel build, the same 640KB and the same CGA card — an entirely ordinary pair of machines. Resume. `cpu_detect` runs and correctly writes `CPU_8086`… and then the image is restored *over it*, putting `cpu_tier = 2` and `vmm_on = 1` back. The first UI pass hits `VMM_POLL`, the gate says yes, and the 8088 executes `push eax` — where `0x66` isn't a prefix at all. CLAUDE.md names `xt-mfm` as *"the machine to install and hibernate on"*, so this isn't contrived.

`XMEM.DRV` is immune to this because hibernate detaches every non-disk driver before it writes. Your island would have been the first 386 code in the *resident kernel*, which is exactly what hibernate restores verbatim. Making it a file closes that by construction.

There's a nice precedent for all of this: SPEC.md §41.12 makes almost word-for-word the same argument for `XMEM.DRV`, down to "reserved forever on a machine that can never reach it".

## The shape it took

`DRVC_OVL` **with** a `drv_tab` row — the only such combination in the tree, and each half earns its place:

- **No class.** A class is a *publication slot* (§51.2.1) — how a package finds "the driver serving sound". The kernel is this one's only caller, so there's nothing to publish. `DRVC_MAX` stays 5.
- **A row anyway.** §41.12 declined one for XMEM because "a machine either has memory up there or it does not". Half of that holds here — but **the probe is itself 386 code**, so unlike `xm_sniff` there's no resident test that could ask the question. The `SYSTEM.CFG` bit is both the request and the record of it, and Control Panel → Drivers is where you turn it off.

It's **not wanted by default**, like every other row, so a stock disk reads no sector of it. `make vmmousetest` builds one with bit 5 set, which is what the gate boots.

## Bugs fixed on the way

Five, each of which was real. No blame in any of this — the 8042 one in particular is a trap the kernel's own code only avoids because of a comment that explains why.

1. **`vmm_p2wake` could feed the 8042's acks to `int 09h` as keystrokes.** It wrote `0xD4/0xF4/0xD4/0xF5` with IRQ1 unmasked and the keyboard interface live, then read the acks testing OBF alone. §9.9.1 step 1 says `mou_p2_init`'s `0xAD` is *"not tidiness"* for exactly this reason — the controller's own replies raise IRQ1 too. And `kbm_isr`'s aux filter is gated on `[mou_p2]`, which your path leaves at 0. In your world: a stray `Z` appearing in a document at boot, or a keystroke you typed just vanishing. It's now bracketed one command at a time, with reads filtered on status bit 5, restored on every exit including the timeouts.

2. **`vmm_bd` ran with interrupts on.** The comment cited `xmem.asm` as precedent, but `xmem.asm` does the opposite — its islands sit inside `pushf`/`cli`…`popf` and its header calls that window *"the whole correctness argument"*. Since `task_yield` is a caller, a tick between `mov eax, VMM_MAGIC` and `in eax, dx` switches tasks, and `sch_switch` only saves 16-bit registers — so XMEM's own 32-bit movers can clobber EAX's top half. Symptom: a mouse that stops working occasionally, for no reason you could reproduce.

3. **`vmm_flush` was an unbounded loop, running at attach inside the boot sequence.** "The count strictly fell, so this terminates" is true of a host that behaves; a host that doesn't is a machine that never finishes booting. Bounded now.

4. **The apply ran on the calling task's stack.** This is the one #147 changed under you. §9.10 had just moved the `mou_apply → cur_move` chain onto a private stack *specifically* so it stops being a cost every task slice must be sized for — and then halved `STK0_SIZE` from 1,024 to 512 citing that removal. Polling from `task_yield` put the same chain back on every task's stack by a different door: `task_yield`'s deepest path measured **28 → 66 bytes**, which puts the 128-byte idle slice within one word of `sch_stkdie`. It now borrows `mou_pstack`, which is provably idle once the backdoor has won (`mou_lockon` has retired both mouse ISRs), with the swap inside the `cli` — `SP` is what `sch_switch` banks, so a tick during the swap would save the mouse's stack as that task's.

5. **The `vmport=off` mitigation lived inside `$(QEMU)`.** Six recipes and four documents tell the reader to replace that variable wholesale — the band benchmarks' `QEMU="qemu-system-i386 -icount shift=3,sleep=off"` is one — and any of them would have silently dropped the flag, putting the default `pc` machine back with `$(MOUSE)`'s msmouse still attached. That's both pointing devices at once, which is the `[mouse_btn]`-sticks-pressed freeze your own commit fixed. It's now `$(QEMUMACH)`, appended at each of the eight use sites.

Plus one the merge found: after #147 moved `mouse_init` into the `.ovlw` boot overlay, **four of your calls crossed a segment boundary as near calls**. `tools/os88ovlchk.py` refuses the build over it; without that gate they'd have assembled cleanly and executed FAT-window bytes at boot. The rewrite removes all four.

## Flagged, not changed

- **Dual display.** `[vid_w]`/`[vid_h]` are the *union bounding box* on a two-monitor desktop, so absolute coordinates would spread across both at ~2×. Your §9.11.4 already names this. I left it as written rather than adding a refusal: v86 presents one framebuffer and the extended desktop is 86Box's two-card XT, which has no backdoor, so the two can't meet today.
- **`.cold` rather than `.ovl`.** `.ovl` is handed back to the heap at `spl_finish` and would be free after boot, but the blob is full — putting `vmm_boot_x` there trips `OVL_AT + OVL_SIZE > BOOT2_PAD`, and `BOOT2_SECS` is a decision about how many sectors stage 1 reads rather than a build fix. So the attach costs ~124 resident bytes.
- **The boot order.** `vmm_boot_x` runs after `mouse_init`, so a browser still pays `MOU_IDWIN` for a serial identify it won't use. That's the safer order — a machine whose backdoor turns out not to answer still has whatever it really has — but your plan's note about an early return is still worth doing as its own change, with its own before/after measurement.

## Testing

`make vmmousetest` builds the gate disk. `tests/vmmouse.py` now also drags through a menu, which is what actually exercises the `task_yield` service point — the thing that would freeze if the pump were wrong. `make test-fast` and `make test-full` are both green, `kern_small` is byte-identical, and I checked it on the glass: the menu tracks under the drag, the selection takes, Bounce launches, and the pointer lands on the pixel asked for.

Commits: `c3bcd28d` (the `main` merge, with the §9.10/§9.11 section collision resolved) and `cf795e4a` (the rewrite).

stuaxo:
Cool cheers - when I got claude to start this I'd only heard about most of these things.

It makes it actually possible to use the paint program and other things.

It makes me start to think about other guest integrations, but I'm not quite sure how they should work, but maybe the things they would use don't work yet.

So - clipboard would be good, but does os8088 have one (yet).

If it had one it would be nice if it worked a bit like a virtual filesystem.

jggonz:
@stuaxo I'm working on merging this in as a driver.

jggonz:
Reopened this — it looked like the close was accidental, since it landed a couple of minutes after your comment. If you did mean to close it, say so and I'll put it back; no harm either way.

**Pushed one more commit** (`d096bee9`), all small follow-ups from the review:

- **The `task_yield` gate got cheaper on the target machine.** The old hook was `push ds / push ax / mov ax, KERNEL_SEG / mov ds, ax / cmp / jz / pop / pop` — and because the pushes happen whether or not the branch is taken, an XT executed all 16 bytes of it on *every task switch*, forever, to answer a question whose answer is always no. The compare needs no register at all: every entry to `task_yield` has CS = KERNEL_SEG, so `cmp byte [cs:vmm_on], 0` works and the segment load moves inside the branch. **8.2 µs instead of 19.1** — 1.2% of a 693 µs switch rather than 2.8% — for one byte of image.
- **The test can now see an off-by-one at the screen edges**, which it couldn't before. It hardcoded 640×480 and took a 4 px tolerance; it now reads `[vid_w]`/`[vid_h]` out of the guest (so the row means the same under `VIDEO=cga` and `VIDEO=ega` — a hardcoded target is wrong *silently* on a smaller screen, because it still lands inside it), injects `(0,0)` and `(vid_w-1, vid_h-1)`, asserts every point lands strictly on-screen, and runs at 2 px. The extremes are the interesting ones: SPEC §9.11.3 claims `(v * vid_w) >> 16` maps 0 → 0 and 0xFFFF → `vid_w - 1` for every width, and that's exactly what the interior points can't test.
- SPEC §9.11 now records the one hazard I left open rather than implying there isn't one: if a task is preempted inside `vmm_poll` and an exclusive bracket (§53.2) is then armed, the guard's holder can be ineligible for the whole bracket and the pointer is dead until it ends. Tens of microseconds wide, and the right fix isn't obviously the guard's, so it wants §53.2's rules confirmed before anyone calls it a bug.
- `docs/TESTING.md` gains the **seventh entry in QEMU's closed list**, which that file explicitly asks for when one turns up: MartyPC decodes nothing at port `0x5658` — which is the reason the protocol chose that port — so there's no backdoor to host anywhere else.

One thing I got wrong and have corrected: one of my reviewers reported that the four cross-segment calls were in your original and would have hung every boot. I checked before repeating it, and it's not so — at your branch's actual base (`682988bc`) `mouse.inc` had no `.ovlw` section at all; the overlay split arrived with #147. The reviewer had used the wrong merge-base. What I said in my first comment stands: it's a merge regression, not your bug.

---

## On the clipboard — yes, there is one

`kernel/clip.inc`, SPEC §55: **one text buffer, owned by the kernel, shared by every application on the machine.** Cut/Copy/Paste is §22.3 (`kernel/filecp.inc`, and on `kern_small` it's an on-demand module). So the thing you'd build a host integration *onto* already exists.

It is **text only**, and the file is explicit that this is a decision rather than a first step — a typed clipboard needs a format registry and a negotiation between the putter and the getter, which is a lot of machinery for a machine this size.

Worth knowing for the shape you were sketching: a package reaches anything outside the machine through a **driver**, not through the kernel (§72 is the worked example — `ETHER.DRV` publishes the socket surface as `OSAPI_DRV_CALL` verbs, so the kernel learns nothing about TCP and a machine with no card refuses at one fence). A host-clipboard bridge would want that same shape, and so would a host-filesystem passthrough — `DRVC_FILE` already exists for exactly "a volume served by somebody answering questions about *files* rather than about sectors", and §51.2's own comment names "a host-filesystem passthrough under an emulator" as one of the four things that class was drawn for. So the seams are there. Whether any of it is worth building is the maintainer's call, not mine.

stuaxo:
My bad, I assumed it was already in.

PR #160, opened 2026-09-07
Fix the build: OSAPI_XCELL/OSAPI_NCELL must jump STRICT near

Koriban:
**This is latent on nasm 2.x, which is what CONTRIBUTING targets and what you build with. It bites on nasm 3.x.** Nothing is on fire; the argument for taking it is that the code currently depends on an optimiser choice rather than on the assembler being told what to emit.

## What happens

`OSAPI_XCELL` and `OSAPI_NCELL` end in `jmp near api_x` / `jmp near api_n`. nasm 3.x shortens those to `EB rel8` when the target is close enough, so the cell emits **7 bytes instead of 8**. Five cells do it, the table comes out 5 bytes short, and `kernel.asm`'s own assertion fires:

```
kernel/kernel.asm:3805: error: os8088 API jump table must be exactly 159 8-byte slots
```

nasm 2.x does not shorten them, so the same tree builds and the table is exactly 159 slots.

**`OSAPI_JSLOT` already documents this and guards against it**, three macros up:

> Without it NASM shortens the jump to `EB rel8` whenever the target is within 127 bytes … every slot after the shortened one slides down a byte and answers at the wrong address. That is the whole reason the length assertion below exists, and it is what caught this.

It uses `jmp strict near`. The two newer macros do not. `api_x`/`api_n` and `OSAPI_NCELL` came in with #147; `OSAPI_XCELL` with #151.

The assertion is doing its job either way — without it, a build that shortened a cell would ship a kernel where every slot past `0x0498` answers at the wrong address, silently. What `strict` buys is that the emitted size stops depending on which nasm is installed.

## The second half is version-independent

`tools/stkbalance.py`'s `JMP` pattern accepts `short|near|word` but not `strict`, and `MACJMP` only `short`. So `jmp strict near api_x` parses **"strict"** as the target and the edge is lost — `api_x` then walks from zero and reports `ret at depth -1`.

That hole predates this PR: `OSAPI_JSLOT`'s existing `jmp strict near %1` has been mis-parsed the same way since #123, on every nasm version, so those five tail jumps were never followed.

## Verified against `783a322` (#166)

| | |
|---|---|
| tip, nasm 2.15.05 | builds; table is 113 `SLOT` + 41 near cells + 5 `JSLOT` = 159 slots, 1,272 bytes, none shortened |
| tip, nasm 3.01 | 5 cells emit `EB rel8`; assertion fires |
| tip + this PR, nasm 3.01 | `make` exit 0, **48 passed, 0 failed** |

Cherry-picks onto the tip with no conflict. Under nasm 3.01, `682988b` (#146) builds and `b9bb040` (#147) does not.

*(Earlier revisions of this description said "main has not built since #147" and that #147 moved existing cells into range. Both were wrong: it builds fine on 2.x, and #147 introduced the macros rather than moving anything. Corrected here so the description stands on its own.)*

### DIFF
--- a/kernel/kernel.asm
+++ b/kernel/kernel.asm
@@ -2677,14 +2677,14 @@ dbg_reg_at:                     ; 0060:000E - THE DEBUG REGISTRY (SPEC.md 57)
 %macro OSAPI_XCELL 1                ; 8 bytes exactly
     push bp                         ; 55        the CALLER's, under the frame
     mov bp, %1                      ; BD lo hi
-    jmp near api_x                  ; E9 lo hi
+    jmp strict near api_x           ; E9 lo hi
     db 0
 %endmacro
 
 %macro OSAPI_NCELL 1                ; 8 bytes exactly
     push bp
     mov bp, %1
-    jmp near api_n
+    jmp strict near api_n
     db 0
 %endmacro
 
--- a/tools/stkbalance.py
+++ b/tools/stkbalance.py
@@ -98,7 +98,7 @@ POP = re.compile(r"^(pop|popa|popf)\b", re.I)
 RET = re.compile(r"^(ret|retn|retf|iret)\b", re.I)
 RETF = re.compile(r"^retf\b", re.I)
 PUSHF = re.compile(r"^pushf\b", re.I)
-JMP = re.compile(r"^jmp\s+(?:short\s+|near\s+|word\s+)?(\S+)", re.I)
+JMP = re.compile(r"^jmp\s+(?:strict\s+)?(?:short\s+|near\s+|word\s+)?(\S+)", re.I)
 JCC = re.compile(r"^(j[a-z]{1,3}|loop|loope|loopne|loopz|loopnz)\s+"
                  r"(?:short\s+)?(\S+)$", re.I)
 SPADD = re.compile(r"^(add|sub)\s+sp\s*,\s*(\S+)$", re.I)
@@ -118,7 +118,7 @@ MACDEF = re.compile(r"^%macro\s+([A-Za-z_][A-Za-z0-9_]*)\s+(\d+)", re.I)
 MACEND = re.compile(r"^%endmacro\b", re.I)
 # `jmp %1` / `call %1` inside a macro body: the macro's ARGUMENT is a branch
 # target, so an invocation is control flow and not an address being taken.
-MACJMP = re.compile(r"^(?:jmp|j[a-z]{1,3})\s+(?:short\s+)?%1\s*$", re.I)
+MACJMP = re.compile(r"^(?:jmp|j[a-z]{1,3})\s+(?:strict\s+)?(?:short\s+|near\s+)?%1\s*$", re.I)
 MACCALL = re.compile(r"^call\s+%1\s*$", re.I)
 # `jmp short $+2` and friends: a jump to the very next instruction, used all
 # over this tree to let an I/O port settle.  It is not a tail call.


Elendilon:
I'm wondering if you are building with a different version of NASM Kori? I've never hit any of these.

Koriban:
**Correction to the description above: this is nasm-version-dependent, and I stated it too broadly.**

I said "`main` has not built since #147". That is wrong as written. `main` builds **fine** on nasm 2.x. What I should have said is:

> `main` does not build under **nasm 3.x**.

Same tree at `a20e2cd`, same flags, only the assembler differs:

| nasm | result |
|---|---|
| **2.15.05** | `make` exit 0, `test-fast` 45 passed 0 failed. Table: 113 `OSAPI_SLOT` + 41 near cells + 5 `OSAPI_JSLOT` = **159 slots, 1,272 bytes, zero shortened** |
| **3.01** | 5 cells assemble as `EB rel8`, table is 5 bytes short, and the assertion fires |

So the two things both had to happen: **#147** brought `api_x`/`api_n` inside `rel8` range of five cells (on 3.01, `682988b` built and `b9bb040` did not), and **nasm 3.x's optimiser then shortens the jump where 2.x leaves it as `E9 rel16`**. Neither alone is enough, which is why this shipped — if you build with 2.x you would never have seen it.

That makes it more worth fixing rather than less. `strict` is the documented way to pin the encoding, `OSAPI_JSLOT` already uses it for exactly this reason, and the alternative is a fixed-size table whose size depends on which assembler you happen to have. CONTRIBUTING notes Homebrew ships nasm 3.x, so this is what a macOS contributor hits on a clean clone.

The `tools/stkbalance.py` half of the patch is version-independent — its `JMP` pattern accepts `short|near|word` but not `strict`, so `OSAPI_JSLOT`'s existing `jmp strict near %1` has been mis-parsed since #123 on every nasm.

Apologies for the overstatement; the diff is unchanged.

Koriban:
> I'm wondering if you are building with a different version of NASM Kori? I've never hit any of these.


3.01, as per my other comment. 

Seems 2.x compiles it fine and 3.x wants this fix.

Elendilon:
I'll see if I can get a 3.x build in the soak, thanks.

Koriban:
**One correction to my previous comment, and a re-verification against today's `main`.**

I wrote that #147 "brought `api_x`/`api_n` inside `rel8` range of five cells". That is wrong, and it makes the bug sound more incidental than it is. **#147 introduced them.** `api_x:`, `api_n:` and `OSAPI_NCELL` all first appear in `b9bb040`, along with the `jmp near` that has no `strict` — the macro was written that way, rather than drifting into range later. `OSAPI_XCELL`, the second macro with the same pattern, arrived with #151 (`7ae7627`).

The measurement behind it is unchanged and I re-ran it: under nasm 3.01, `682988b` (#146) builds and `b9bb040` (#147) does not.

**Re-verified against `783a322` (#166), the tip as of now:**

| | |
|---|---|
| tip, nasm 2.15.05 | `build/kernel-full.bin` builds, 119,961 bytes |
| tip, nasm 3.01 | `kernel.asm:3805: … must be exactly 159 8-byte slots` |
| tip + this PR, nasm 3.01 | `make` exit 0, **48 passed, 0 failed** |

The branch still cherry-picks onto the tip with no conflict, so there is nothing to rebase.

The `tools/stkbalance.py` half is unchanged on the tip too — `JMP` still accepts `short|near|word` and not `strict`, and `MACJMP` only `short` — so `OSAPI_JSLOT`'s existing `jmp strict near %1` is still being mis-parsed there on every nasm version.

jggonz:
Picking this up for review. **Short version: I am not planning to change anything — the patch is right as submitted, and I can now show the trap is armed on my machine too, on a different nasm than either of yours.**

Here is what I checked and what it said.

## It merges clean, and the diff stays four lines

Merged today's `main` (`783a322`, #166) into the branch: no conflicts, and the merged tree differs from `main` by exactly the two hunks in this PR. Nothing was lost from `main` and nothing needs rebasing.

## The condition is present on the tip, and I measured it

I am on **nasm 3.02** (Homebrew, macOS). On the tip with your `strict` removed, `make` succeeds for me — so at first glance this looked like it did not reproduce.

It does. I decoded the assembled table and measured the distance from every `OSAPI_XCELL`/`OSAPI_NCELL` to its target:

| target | cells pointing at it | nearest cell |
|---|---|---|
| `api_x` | 34 | **23 bytes** |
| `api_n` | 7 | 139 bytes |

Counting the cells nasm would be *entitled* to shorten — target within `rel8` reach of a two-byte `EB` encoding:

```
SHORTENABLE cell 0x14b8 -> api_x  d= 96
SHORTENABLE cell 0x14c8 -> api_x  d= 80
SHORTENABLE cell 0x14d0 -> api_x  d= 72
SHORTENABLE cell 0x14d8 -> api_x  d= 64
SHORTENABLE cell 0x1500 -> api_x  d= 24
total shortenable: 5 of 41
```

**Five** — exactly the five you saw shorten under 3.01. So this is not "a 3.01 quirk". The cells really are inside range; 3.01 takes the shortening and 3.02 declines to. Which of those two an assembler does is its own business, and that is precisely the thing `strict` exists to take away from it. The table's fixed-size assertion is currently resting on an optimiser's discretion.

Same measurement on **`kern_small`**: also 41 cells, also 5 shortenable. That build is ~12KB tighter, so if anything it is the one more likely to bite next.

You were right to widen the framing in your follow-ups, and the `#123` observation lands — `OSAPI_JSLOT` sits eighteen lines above these two macros with a nine-line comment explaining this exact failure, and the two new macros were written without the word.

## The `stkbalance.py` half is not a cleanup — it is required

This is the part I want to flag back to you, because it makes the case for shipping the two hunks together stronger than the PR body claims.

With `strict` added to `kernel.asm` and the **old** `JMP` regex still in place, the fast tier's `stkbalance` row **fails**:

```
kernel/kernel.asm: api_n: ret at depth -1
kernel/kernel.asm: api_x: ret at depth -1
stkbalance: 2 unbalanced path(s)   [exit 1]
```

The regex takes `strict` as the branch target, the edge into `api_x` is lost, and the tool walks the routine from zero. With your fix: `0 unbalanced path(s)`, exit 0. So the second hunk is not optional tidying that happened to ride along — without it this PR would trade a build failure for a test failure.

On the pre-existing gap you describe: I A/B'd it against `main`'s own `kernel.asm` too. `OSAPI_JSLOT`'s `jmp strict near %1` is indeed mis-parsed there, and the walk covers slightly different ground (4,151 entries vs 4,146) — but it has never produced a wrong verdict, both arms report 0 unbalanced. So it is a real coverage hole that has not yet cost anything, which is about the best time to close one.

I also checked the widened `MACJMP` cannot over-match. Across `kernel/`, `apps/`, `drivers/` and `tests/` the only macro-body jumps to `%1` are seven bare `jmp %1` in `apps/weave/wvm.inc` and `OSAPI_JSLOT`'s `jmp strict near %1`. Adding `strict` and `near` to the pattern picks up the latter and nothing else.

## Verification here

- `make` — 48 passed, 0 failed
- `make KERN_SMALL=1` — builds
- `python3 tools/kernsize.py --build build` — both kernels under their guards
- `python3 tools/os88disk.py --verify` — clean on the shipped images
- **`build/kernel.bin` is byte-for-byte identical with and without this patch on nasm 3.02.** On an assembler that was not shortening these anyway, the change costs the shipped image nothing at all. It only ever removes a choice.

## What I am not changing

Nothing in the patch. Your closing question — whether the rule belongs somewhere a new slot macro cannot miss it — is a good one and I think it is worth doing, but it is a separate change and a maintainer's call about how to enforce it, not something to bolt onto a four-line build fix. I would rather this went in as it stands.

Thanks for the bisect, the table decode and the two corrections on your own description — the second one in particular (that #147 *introduced* these macros rather than drifting them into range) is the detail that makes the pattern clear.

PR #161, opened 2026-09-07
Picture decoders: .PIX/.BMP/.PCX into one 4bpp form (apps/os88img.inc)

Koriban:
Three file formats into one in-memory form: **packed 4bpp, two pixels per byte, high nibble leftmost**, which is exactly what `OSAPI_GFX_BLIT4` takes and exactly what a `.PIX` block already holds. However a picture arrived, there is one drawing path for it.

`BLIT4` and not `OSAPI_GFX_BLITP`, which would suit a planar PCX better — the reasoning is in the section.

## What is here

- `apps/os88img.inc` — the decoders. Owns no state; the caller passes a block in `SI`. 8-bit files are **refused by name, not approximated**.
- `apps/imgtest/` — its self-test package, and `tools/os88imgcase.py`, which generates the corpus.
- **SPEC.md §91.**

## The test is the point

`imgtest` is not on any disk and is built by `all` so it cannot rot — `fptest`'s arrangement, for `fptest`'s reason. Its expectations are **computed on the host from the format documents**, not by running the decoder, so a reader and a writer that share a misunderstanding cannot agree with each other into a pass.

The corpus is seventeen generated cases, or twenty-two if you have five third-party files from the Dr. Dobb's File Formats disc that this repository does not ship (the generator says which it built). A third-party file is the only kind that cannot share a misreading with the decoder, so run it with them if you have them.

## Verification

```
make                     # exit 0
make test-fast           # 45 passed, 0 failed

python3 tools/os88imgcase.py
python3 tools/os88disk.py -o build/imgtest.img --size 1440 \
    APPS:build/imgtest.o88 $(addprefix APPS:,$(wildcard build/imgcases/*))
make test TESTAPPS=build/imgtest.img
```
and read the window: one row a case, ALL PASS or FAILURES.

## Note on ordering

Built on `a20e2cd`. **`make` does not currently complete on `main`** — see #160, which is four lines and independent of this. With #160 applied on top, the numbers above are what I measured.


jggonz:
Reviewed, merged `main` in, and pushed two commits — the merge and one commit of fixes. **The PR is mergeable now**; the only thing left on it is the approval, which is mine to give.

`make` and `make KERN_SMALL=1` are clean, `python3 tools/os88test.py full` is **39 passed, 0 failed, 0 skipped**, all eight shipped images `--verify` clean, and I ran the decoder's own gate on the glass: **ALL PASS over 27 cases**.

This is a good piece of work and the section is unusually well argued. Everything below is a place where the code and that section disagreed — which is only findable *because* the section says what it says.

---

## The merge

`main` moved by one commit while this was open, but that commit was the Elendilon merge (#172), which added CLEAR SKIES as §88 and renumbered **THE WIRE from §88 to §92**. Your section claimed §92 as well, so after the merge SPEC.md would have had two of them — and `tools/checkdocs.py` cannot see a duplicate heading, because a citation to §92 still *resolves*, to the first one it finds.

THE WIRE keeps §92 (it is published and referenced from `CLAUDE.md`, `docs/WIRE-PLAN.md` and `tests/`), and **the picture decoders are §93**. Three files carried the number and all three moved.

One related thing worth knowing: `tools/os88index.py` said **§88** for `os88img.inc` while SPEC.md said §92, so `docs/INDEX.md` was pointing a reader at THE WIRE. Both say §93 now. `docs/INDEX.md` is regenerated, so it also picks up #172's new `ships` column — IMGTEST lands beside FPTEST as `no`, which is right.

---

## What I fixed

### 1. A BMP whose pixels start past 64KB decoded the file header as a picture

**In your world:** you open a large BMP and instead of "this file is too big", you get a window full of noise, and the decoder says it succeeded.

**The mechanism:** the two tests at the top of `img_bmp` name `bfOffBits` and `biSize` in their comments and read `[es:16]` and `[es:20]` — which are `biSize`'s high word and `biWidth`'s. `biWidth`'s is then checked *again* at `.geom`, which is the tell that the pair had slipped one field along. So `bfOffBits`' high word was never read at all: a file with `bfOffBits` = 0x00010000 kept only the low word, `IMG_P` became 0, and the row loop read its pixels out of `'B','M'`, `bfSize` and `biSize`. Every read stayed inside `IMG_SRCLEN` so nothing was corrupted — but §93.1's own reasoning ("a picture past the 64KB horizon has its pixels past it too") says refuse, and `img_pix` **does** refuse exactly this at its `[es:di+10]`.

**Changed:** `[es:12]` and `[es:16]`, and the comment now says which dword each is the high word of. `177ff74`

### 2. A `biSize` of 65532 wrapped the palette pointer back into the file header

**In your world:** a corrupt or hostile BMP comes back as a picture in sixteen colours taken from its own header bytes, rather than being refused.

**The mechanism:** `biSize` is floored at 40 and never ceilinged. `mov di, 14 / add di, [si+IMG_P]` can carry out of 16 bits — and the `jc` was on the *next* add, not that one — so `di` wrapped to 10, the bound test then passed on the wrapped value, and `img_palmap` read offsets 10..73 as sixteen B,G,R,reserved quads.

**Changed:** one `jc .trunc`. `177ff74`

### 3. A refusal left a complete, plausible picture size in the block

**In your world:** a package that reads `IMG_W != 0` as "there is a picture" draws half a truncated file and half of a stale claim.

**The mechanism:** `img_setgeom` stores `IMG_W`/`IMG_H`/`IMG_STRIDE` *before* any decoder has proved its pixel data is inside the file. `CUT.PCX` reaches it with an intact header, decodes about half the rows, and then refuses — returning `CF=1` with 37x11 stride 19. `imgtest.asm`'s own comment already believed otherwise: *"a refusal case ends here: there is no picture to compare and IMG_W is 0"*.

**Changed:** `img_load`'s `.fail` clears all three, and the harness now compares the geometry **on a refusal too**, against zero, so the claim has a case rather than a comment. With those three lines taken back out, four cases fail (`CUT.PCX`, `CUT.BMP`, `BADOFF.PIX`, `BIG.BMP`) — I ran that A/B. `177ff74`

### 4. `.PIX` was the one format that did not mask its odd-width padding nibble

§93.1 spends a paragraph on that nibble being a property of the *output* format, "decided in one place and not two". The one place was `img_rowconv`'s tail — and `.PIX` converts nothing, so it never reaches `img_rowconv` at all. The format this tree writes was the one format whose padding nibble was whatever the file said. It is `img_rowmask` now, called by both. `177ff74`

### 5. A PCX at version 3 decoded to a picture entirely in black

**In your world:** you open a PC Paintbrush 2.8 file and get a black rectangle, reported as a success.

**The mechanism:** ZSoft's version byte at offset 1 is **3** for "2.8 *without* palette information" — the sixteen triples at offset 16 are then not a palette. A writer that leaves them zero sends every index through `img_palmap` to the nearest colour to black, which is `CBLACK`. What version 3 means is the fixed EGA palette, and that **is** os8088's own sixteen in order, so that case takes `img_palident`. `NOPAL.PCX` in the corpus is `P4.PCX` with that one byte changed and its palette zeroed, and it decodes to the identical checksum. `177ff74`

### 6. The committed `imgcases.inc` was the Dr. Dobb's disc build

**In your world:** anybody who follows the recipe in the Makefile — which is everybody but you — reads **FAILURES** off a gate that is working perfectly. I have the screenshot: five `FAIL` rows for `MAIN.PCX`, `INSTALL.BMP`, `START.BMP`, `SAMPLPIC.BMP` and `HELP8.PCX`.

**The mechanism:** the `.inc` names 23 cases; the generator writes 18 in a checkout without the five third-party files, and `OSAPI_FILE_READ` answering `CF=1` for a file that is not there is `.fail` in `it_one`. It also meant `--check` could never become a `make` row: it would have to fail on your machine or on every other one.

**Changed:** the disc cases are opt-in (`python3 tools/os88imgcase.py --with-disc`) rather than picked up when present, so the committed table is the one this repository can reproduce; the generator prints a note if it sees the disc files and was not asked for them. `177ff74`

### 7. The checksum could not see a row swap — which is the one thing it exists for

Three separate places say it: §93.3 ("A plain sum would not notice two rows swapped; making position part of the answer is the whole point"), the generator, and `it_cksum`.

**The mechanism:** a 16-bit rotate-by-one is a linear map of **order sixteen**, so any two rows whose byte distance is a multiple of 16 land on the same rotation and exchanging them leaves the value unchanged. On the three 1bpp cases (stride 16) that is *every* pair of rows in the picture; on `MAIN.PCX` (stride 576) likewise. A decoder that reversed or rotated the row order passed — and a bottom-up BMP decoding upside down is one of the two defects §93.3 credits this corpus with catching. I verified it numerically: stride 16, rows 0 and 1 exchanged, both `0x1C81`.

**Changed:** multiply by *x* modulo x^16 + x^12 + x^5 + 1 instead, which has order **32767** — no picture this decoder accepts can move `(j-i) * stride` that far. Zero collisions across every corpus stride. It is also four instructions where the rotate was seventeen, and one line on the host. `177ff74`

### 8. `IMG_DSTMAX` was a constant 0, so the capacity check was untestable

`imgtest.asm` set `IMG_DSTMAX` to 0 ("the whole 64KB") for every case, which takes `img_setgeom`'s own `jz .fits` and leaves the compare below it unreachable — **a decoder that ignored the caller's capacity entirely would have passed the whole corpus**. That is not an uncovered path, it is one that adding a case cannot reach.

**Changed:** the record carries a capacity per case (`IMGC_REC` 14 → 16), and `BIG.BMP` is a 30x9 picture given 134 of the 135 bytes it needs. `177ff74`

### 9. Five of the nine refusal codes had no case at all

`IMG_E_SHORT`, `IMG_E_DIM`, `IMG_E_BIG`, `IMG_E_COMP` and `IMG_E_VER` were only ever described. `IMG_E_SHORT` is the loud one: §93.1 spends a paragraph on the per-format length floor and on how a single floor of 128 was *wrong* once 1bpp arrived, and that reasoning was checked by a described mutation rather than by a file.

Added `TINY.BMP` (ten bytes — under `img_load`'s own sixteen), `SHORT.BMP` (forty — past that and under `img_bmp`'s fifty-four, so the two floors have a file each), `CUT.BMP` (the format that refuses on arithmetic rather than inside the RLE, so `CUT.PCX` does not stand for it), `BADOFF.PIX` (a directory entry pointing past the end — the `.PIX` bounds test had nothing that made it fire), `RLE4.BMP`, `W0.BMP`, `V2.PIX`. Twenty-seven cases. `177ff74`

### 10. Building it is not running it

`all` names `imgtest.o88`, which keeps it **assembling** — the FPTEST arrangement, and right. But the four steps that actually exercise the decoder were a comment. Added **`make imgtestdisk`** (generator + floppy in one step) and an **`imgcases` row in `make test-fast`** running `--check`, so the generated expectations cannot go stale under you. Possible because fix 6 made the output deterministic. `177ff74`

---

## Checked and clear

- **Memory safety: I could not construct an out-of-bounds read or write anywhere in the file**, and neither could two independent reviewers. Every loop bound is derived from a value already clamped, and several of those guards are load-bearing in ways that are not obvious from the call site: `IMG_TOT` capped at `OS88IMG_RAW` after `or dx,dx` on the `mul`; `BytesPerLine` refused if zero, odd, or under the width's own minimum *at that depth*; `IMG_SBPR` bounded before the copy loop uses it; `img_setgeom` proving `stride*H` fits a word **and** the caller's capacity, which is what lets `img_rowout`'s `mul` ignore DX; a PCX run count of 0 refused rather than spun on; `img_row_planes`' highest address provably `< NPL*BPL`; and both `xlat` feeds masked to a nibble. `OS88IMG_RAW` + the packed half is 640 + 640 with nothing spare, so `OS88IMG_WMAX` is load-bearing on the `.PIX` path too, which is easy to miss because `.PIX` decodes nothing.
- **`cpu 8086`**: swept the whole file — no `pusha`, `movzx`, `push imm`, 32-bit register or `shl reg,imm>1`.
- **Register discipline**: `img_row_planes` keeps BP as a *value* and never as a base, which matters because SS ≠ DS; `img_row_nib`'s `mov [di], al` rather than `stosb` is right, and its comment says why. `img_palmap`'s `push dx` around the `mul` is right too.
- **The bottom-up BMP fix is real** in the current code: `biHeight`'s high word decides, `0x8000` is refused, and `0xFFFF0000` negates to 0 and is caught by `img_setgeom`.
- **The independence claim in §93.3 holds** where it matters. I checked it structurally: `pack4()` — the model of the *output* format — is called by no encoder, each encoder writes its own nibble packing inline, and `ref_pcx` carries its own RLE decoder rather than `rle()` inverted. So the checksum really is a model of the output applied to a model of the input.

## Flagged, not fixed

- **`biClrUsed` is ignored.** `IMG_NPAL` is inferred from the depth alone, but a 4bpp BMP may legally carry fewer than sixteen entries — a 4x4 4bpp BMP with `biClrUsed = 4` is 86 bytes and is refused as truncated, which is the identical failure §93.1 records for `M1.BMP` before the floor was made per format. The safe shape is to *shrink only, never grow*, so the sixteen-entry read stays bounded whatever the field says. I left it because it widens what the decoder accepts, and that is your call rather than a repair.
- **`docs/IMGCONV-PLAN.md` is not in the tree**, and two places leaned on it (§93.1's refusal argument and the `biCompression` comment's "3a gives it its own extension"). I made both sentences stand on their own rather than invent the document — if you have it, it belongs in `docs/plans/` and the citations can come back.
- **`img_pcxbyte` clobbers AH** against its "preserves everything else" header. Harmless today — its only caller does not keep anything in AX across it — and AH is 0 on every `CF=0` return, so the honest fix is to document AX as the output. I did not touch it.
- **`CUT.PCX` writes partly-decoded rows into the destination before refusing**, and the harness cannot tell that from a refusal that wrote nothing. Pre-filling the destination with a sentinel would close it. Not a safety issue — every one of those writes is inside the capacity `img_setgeom` proved.
- `tools/kernsize.py` reports +39 against the blessed baseline on this branch. It is **not yours** — this PR touches no kernel, boot or driver file, and `main` reports the same.

## What I left alone

Everything about the design: `BLIT4` over `BLITP` and the reasoning for it, the one-output-format decision, the caller-owned block, `IMG_ERR` as a number, `img_pal8088` through CS, Manhattan over Euclidean, the two-row-routines dispatcher, and the choice to refuse 8-bit by name. Those are the section's arguments and they are good ones.

---

**To pull this down:** `git fetch origin pull/161/head && git checkout -b pr/imgdec FETCH_HEAD`, then `make imgtestdisk && make test TESTAPPS=build/imgtest.img`.

PR #162, opened 2026-09-07
SCRIBE — a fork of WORD as its own package (apps/scribe/)

Koriban:
**Please read the first section before the diff — this one is a judgement call, not just code.**

## What it is, and why it is a second package

`apps/scribe/` is a copy of `apps/word/` that keeps **every symbol name, every include filename and every line number**, so `diff -r apps/word apps/scribe` is exactly what the fork changed and an upstream fix to WORD reads straight across.

The divergence is 2,363 lines and coherently themed — **Insert ▸ Picture** and **the overlay split**. Nothing else differs; `wdutil.inc` is byte-identical.

It is **not on any shipped disk**. WORD is the one that ships; `make scribedisk` builds SCRIBE's floppy on demand, which is `cword`'s arrangement and for `cword`'s reason. Putting both on the apps floppy would spend 49KB to show two word processors.

It **is** in `all`, so a source edit reaches the binary. While it was a phony target and nothing else, plain `make` never assembled it — an edit could look applied while the emulator ran the previous build. That happened to me; it is why the rule is in the Makefile comment.

**If you would rather these changes landed in WORD itself, say so and I will rework it that way.** The fork exists because Insert ▸ Picture is a scope decision about WORD that is yours, not mine — SPEC.md §92 states the case, and §68.15 records where the two diverge inside WORD's own chapter.

## Depends on #161

`wdrtf.inc` and `wddoc.inc` decode pictures through `apps/os88img.inc`, which is #161. This PR does not build without it.

## Sizes

`scribe.o88` 53,778 bytes → 43,909 resident + `SCRIBE.OVL` 9,869 of the 12,288 `WD_OVKB` reserves.

## Verification

```
make            # exit 0
make test-fast  # 45 passed, 0 failed
make scribedisk
```

`tests/textsites.txt` gains SCRIBE's line: two transparent-text sites, inherited from WORD at the same line numbers (`wd_drawrun`'s BOLD strike and the ruler's inch digits), which is §6.6.2 cases 6 and 1.

## Screenshots

`screenshots/scribe-ovl-doc-load.png` — WELCOME.DOC, a native Word `.DOC` rendered by the fork
`screenshots/scribe-ovl-rtf-picture.png` — a `.RTF` with an embedded picture
`screenshots/scribe-ovl-rtf-resaved.png` — the same file after a re-save, picture intact
`screenshots/xq3-scribe-insert-menu.png` — the Insert menu
`screenshots/xq3-scribe-report.png` — a report with a chart pasted in from SHEET

## Note on ordering

Built on `a20e2cd`. **`make` does not currently complete on `main`** — see #160. The numbers above are with #160 and #161 applied.


Koriban:
**Updated: SCRIBE now carries its own `sc_` prefix.**

The branch previously kept WORD's `wd_` symbols and `wd*.inc` filenames, and the section argued for that: it made `diff -r apps/word apps/scribe` the whole of what the fork changed, and let an upstream fix to WORD read straight across.

Two things outweighed it:

- The app is called SCRIBE, and every symbol in it named the program it was forked from.
- The two shared a symbol **namespace**, not just a shape. `tools/stkbalance.py` walking both files in one pass saw **645 names defined twice** and merged the two apps' call paths — so SCRIBE had to be walked alone to be walked at all. A gate that cannot see two files together has a hole in it.

1,395 identifiers renamed across four files; `wddoc.inc`/`wdrtf.inc`/`wdutil.inc` are now `scdoc.inc`/`scrtf.inc`/`scutil.inc`, and `WD_OVKB` is `SC_OVKB` (read by name in the Makefile and `os88ovlchk`).

**It is a pure rename and that is proved, not assumed**: `build/scribe.bin` has the identical MD5 before and after — 53,778 bytes, 43,909 resident + 9,869 in the overlay.

The header records both sides rather than deleting the old argument, since the cost is real and paid knowingly: an upstream fix to WORD is no longer a patch that applies here.

One inherited gap closed on the way — `sc_rnval` in `scutil.inc` was missing the `; STKBALANCE-LOOP` marker its WORD twin carries. `stkapps` only found it once the namespace collision stopped masking it. All four file pairs now agree on their marker counts.

`make` exit 0, `test-fast` 45 passed 0 failed (with #160 applied, since `main` still carries the build break).

jggonz:
Merged `main` in and pushed two commits (`6e42510` the merge, `ebd0a48` the fixes). **PR #162 is mergeable now** — the only thing left on it is the review approval.

Thank you for the header in `scribe.asm` that records *both* sides of the `wd_` → `sc_` argument rather than deleting the old one. That is the note that made the rest of this review readable, and the `stkbalance` finding behind it — 645 names defined twice — is a real hole you closed.

---

## The merge

`main` gained #161 (the picture decoders) and #172 while this branch was open. Two things collided that git could report and three that it could not:

- **`apps/os88img.inc` and its friends: I took `main`'s.** Your branch carries the pre-review version of #161, and `main`'s went through its own review and differs in *substance*: the case table grew a per-case `IMG_DSTMAX`, a refusal now zeroes `IMG_W`/`IMG_H`/`IMG_STRIDE`, and the corpus checksum became an LFSR because the rotate it replaced had order 16 and could not see two rows of a 16-byte-stride picture swapped. I checked every `img_*` call site in `apps/scribe/` against the shipped contract afterwards — they are all correct, including `IMG_DSTMAX = SC_PICKB*1024 - 1`, which is one byte conservative in the right direction.
- **SPEC section collision.** You numbered the decoders 92 and SCRIBE 93; on `main`, 92 is THE WIRE and 93 is the decoders. SCRIBE is **§94**, and the renumber followed it into §68.15, into every `SPEC.md 93.x` citation in `apps/scribe/`, and into `tests/textsites.txt`. Your own stale `SPEC.md 85` citations — the decoders' number before #161 moved them — are 93 now.
- **`$(BUILD)/scribe.bin` listed neither `apps/os88img.inc` nor `apps/os88type.inc` as a prerequisite**, so an edit to either would not have rebuilt SCRIBE — the exact failure your PR description says bit you with the phony target. `main`'s new `pkgdeps` row catches it, and that row did not exist at your base, so your green `test-fast` was honest.

---

## What I fixed

### Two wrong DS restores, and they are the ones that matter

`push cs / pop ds` is the one idiom that does not survive the move into `.modc`. Resident it is correct — CS *is* the package. In the module CS is the 12KB `SCRIBE.OVL` claim, so the `SCVAR` the next instruction names sits roughly **33KB past the end of it**, in whatever else holds that heap.

It survived the sweep in two places, and both are on live paths:

- **`sc_dpapin`, on every `.DOC` load** (`scdoc.inc`). One byte per paragraph mark written into an arbitrary segment — so opening any `.DOC` with a `¶` in it, which is any `.DOC`.
- **`sc_rpap`, on every `.RTF` save** (`scrtf.inc`). This one is why the verification passed: the write and every read use the *same* wrong address, so the four PAP bytes round-trip perfectly and the file comes out right. What actually happens is that four bytes of somebody else's heap are overwritten once per paragraph.

The reach is larger than those two instructions, because the restore stays wrong for the whole near-call cone below it: `sc_dpapat`, `sc_dprl`, `sc_dfkfind`, `sc_dfkrun`, `sc_dfcof` and `sc_dpget` all read package `SCVAR`s bare — and `sc_dpapat`'s `call far [sc_v_papfind]` reads *its own vector* through DS, because the vector table is in `.text`. Both are `mov ds, [cs:sc_pkgseg]` now. The third `push cs / pop ds` (`scdoc.inc:1173`) is **correct and untouched** — `sc_d_normal` really is module data, which is exactly the distinction §94.8.2 is about. `ebd0a48`

### Four ways a picture claim went missing

Each of these is up to 64KB of heap gone for the life of the instance, and on a 640KB machine a handful of them is the difference between Insert ▸ Picture working and refusing with nothing on screen to explain why.

- **A `\pict` the file ends inside.** `sc_rpfin` is reached only from the `}` handler, so a truncated `.RTF` — a copy that stopped short, or simply a file whose `\pict` is the last thing in it — loaded successfully with the claim orphaned in no `pictab` slot, where `sc_pictfree` cannot see it. `.deep` had the same exit. Both free it now.
- **A second `\pict` inside an open one.** `sc_rword3` re-armed the collector and zeroed `[sc_rpseg]` without freeing — one leak per repetition, and a few-KB file can carry hundreds.
- **A picture registered into a document that is already full.** `sc_rputp` returned silently when `[sc_len] >= [sc_cap]`, so the slot and the claim were held for a picture no character refers to, and the load reported plain success. It answers CF now and `sc_rpfin` unwinds, with `[sc_dtrunc]` set so the "…truncated" toast actually appears.
- **`\picw`, `\pich` and `\wbmwidthbytes` arriving after the claim.** This is the serious one. `sc_rpclaim` sizes the claim on the first data byte, but those three control words go on being accepted for the rest of the group and `sc_rpfin` copied whatever they held into the record. A crafted file could claim 1KB and describe `1280×200`, and `sc_picdraw` would then hand `OSAPI_GFX_BLIT4` a 128,000-byte read from it — **127KB of somebody else's heap painted into the document window**. `sc_rpfin` re-makes every check `sc_rpclaim` made, on the final values.

### Three more from the same read

- **The `\pict` depth test was missing `sc_rskip`'s `dec`.** So the picture finalised on the closing brace of any group *nested inside* it — and `{\*\picprop}` is what Word 6 and WordPad actually write. The collector tore down, and the ~17,000 hex characters that followed fell through to `sc_rput` and went into the document as literal `f`s. Your own writer emits a flat `{\pict …}`, which is why the round trip looks perfect and only foreign RTF is affected — the same shape as the bug §94.6.2 already records.
- **Skip now wins over collect.** A `{\*\…}` destination inside an open `\pict` had its text paired into pixels. With an *odd* number of hex digits in it, the leftover nibble shifted every byte of the picture after it.
- **`sc_penadv` was measuring the wrong picture.** `sc_picw` indexes by `[sc_i]`, and two of its four callers — `sc_wordfit` and `sc_rowmeasure` — walk `SI` ahead with `[sc_i]` frozen at the word's or row's first character. So it read *that* character's CHP byte as a picture index: `SCAT_BOLD` is 1 and `SCAT_UL` is 4, so a bold character measured picture 1 and an underlined one picture 4. Fixed by banking `[sc_i]` across the call, which changes no routine's contract.

### The `.DOC` path

- **`sc_dgrpc` took a picture index straight off the disk.** `sc_dispic` tests only `cmp al, SC_PICCH`, so any `0x01` in a hostile file's text is "a picture", and `sc_dpicr` *ends its walk* at `SC_PICMAX` — leaving later ones with the CHP byte the file built, 0..0x7F, indexing a 16-byte `sc_dpicfc`. In-segment, so a disclosure rather than a crash, but it steers what goes into a saved document. Out of range is an ordinary attribute set now, which is what the byte would have meant on any other character.
- **The reader ignored `lcb`** and stepped by its own recomputation, so a file whose stated length disagreed had the *next* record read out of the middle of this one's pixel data — and bytes that happen to pass all seven field checks are decoded and drawn as a picture. It reads the DWORD, requires the high word zero and `lcb >= 46 + bw*h`, requires the whole record inside the file, and steps by it. Your own files are unaffected: `bw` is always even, so your `lcb` is already even.
- **`dyaGoal` wrapped past `h = 4369`.** Nothing anywhere bounds a picture's *height* — `OS88IMG_WMAX` bounds only the width — so `sc_dpicw` carries the ceiling.
- **`fHasPic` is set.** It is the one FIB field that describes the feature this PR adds, and leaving it zero means the header contradicts the body.

### The RTF writer

- **The emitter kept running after the staging claim filled.** `sc_re` stops storing and sets `[sc_rfull]`; nothing in `sc_rpict` read it, so it walked every remaining picture byte at five near call/ret pairs each. That is **~2.2 seconds per 30KB of picture on a 4.77 MHz 8088**, on the UI task, with no yield, to produce nothing — on a save that was already going to be refused. §94.5 says the ceiling is about 30KB across a document, so a second picture is past the fill point by design.
- **A tall picture came out with a negative height.** `sc_rnum` emits *signed*, and a 2-pixel-wide `.PIX` of 40,959 rows is a 40,991-byte file that passes every check and gets claimed. It saved as `\pich-24577`, which your own reader then refuses on the way back in — a save that reported success with the picture gone. `sc_rpict` refuses the save past `h = 2184` (2184 × 15 = 32,760). `sc_dpicw`'s ceiling is 4369 for the same arithmetic and twice the room, because a PICF's `dyaGoal` is an unsigned `stosw`; the comments say why the two differ.
- A zero stride can no longer make `loop` run 65,536 times off the end of the picture (not reachable today — a defence note), and the four headers that said "preserves every register" say "all but DI and the flags", which is what `sc_res` and `sc_rnum` already said of themselves.

### The gate that could not see any of it

`tools/os88ovlchk.py`'s package walk named `apps/word/word.asm` **and nothing else**, so SCRIBE's nine call-outs and six verbs — the entire subject of the split — were walked by nothing, and it printed `1 package(s) keep every overlay call far`, which is exactly the number a reader expects. §94.8.4 says the checker proved the move; that was true of WORD.

It could not simply be added, either: `scribe.asm` `%include`s `os88img.inc` **twice**, once behind `%define OS88IMG_CONSTS_ONLY` for the constants and once inside `.modc` for the code, and `expand()` inlined both copies verbatim — 21 `.text -> .modc` calls that do not exist in the binary. `expand()` follows `%ifdef`/`%ifndef` now, but only as far as a `%define` *in the stream* proves them and only in the direction that drops rows, so a build-flag guard like `%ifdef SBDRAGOFF` is judged exactly as it was before. It reads `2 package(s)` and SCRIBE is clean.

### The build and the disk

- **`make scribedisk` builds all four geometries.** CLAUDE.md makes that the rule for an on-demand application floppy, and 360KB is 62 of 354 clusters — a sixth of the disk. Without it the XT and the AT-class 5.25″ machines, the ones SCRIBE would actually be looked at on, had no disk to boot it from.
- **`apps/scribe/welcome.wtx` was referenced by nothing.** Editing it changed nothing and the disk still carried Word's text, straight out of `apps/word/welcome.wtx`. It builds `SCWELCOM.RTF` now and that goes on the floppy — and it closes a second gap, because RTF is Scribe's own default format (§94.4) and the disk carried none of it. It is `SCWELCOM` rather than `WELCOME` because cword's rule already owns that name.

### Documentation

- **§94.1 argued at length for keeping `wd_`** — heading and first two paragraphs — over code you had already renamed, and contradicted itself in its own last paragraph. It records both sides now, following your source header, including the 645-names-defined-twice argument and the cost you paid knowingly.
- The `SPEC.md 89` in the Makefile block (that is PACMAN), "neither is in `all`" (the package is, for `wire.o88`'s reason), the `stkbalance` row in §94.7.4 and §94.9.2 (the row that actually covers `apps/scribe/` is **`stkapps`**, which globs `apps/`), the six sizes in §94.3 and §94.8, and §94.9.4's "a picture in a document that scrolls is always inside [an armed clip region]" (SCRIBE arms none — BLIT4 is still the right choice, for §93's reason) are all corrected. README.md gains its two lines.

---

## Flagged, not fixed

**Two things here are yours to decide, not mine to change.**

### 1. The fork point — `apps/scribe/` came from a `word.asm` older than the branch it arrived on (new §94.10)

A `git merge` cannot repair this: a file a branch *adds* takes nothing from the other side, so everything `main` landed in WORD after your copy was taken is simply absent, and git reports no conflict for it. `git log -S<symbol> -- apps/word/word.asm` puts all of it in **#172**:

| what | WORD | SCRIBE |
|---|---|---|
| menu/dialog **save-unders** (`wd_suab`/`sudlg`/`surest`) | banks the pixels and puts them back | repaints the document under every menu and dialog |
| the shared menu bar (`os88ui_mnbar`) | one call | ~179 lines of hand-rolled drawing, superseded |
| the movers (`wd_mvup`/`wd_mvdn`) | `rep movsw`, 36.0 cycles a byte | inline `rep movsb`, ~36% slower on the tail move, twice a keystroke |
| `OSAPI_WM_ONWAKE`, `os88ui_drop`, `OS88_REGION_MOVABLE`, `OS88_WORKER_RESTARTABLE`, `wd_eoutck`, `wd_nlpush`, `wd_upheight`, `wd_bandx` | present | absent |

None of it reaches a shipped disk. It is a maintenance fact: the two files are no longer the same program with a different prefix, so someone diffing them to port a fix across will read hundreds of lines of divergence as intentional. **Re-taking the fork from `main`'s `word.asm` and re-applying §94.5–§94.9 on top is cheapest now**, while nothing downstream depends on SCRIBE — but that is your call, and §94.10 stands either way.

### 2. The picture layout (new §94.11) — a picture has two widths

`sc_penadv` answers the picture's real width and is what the wrap test and the two measure walks use. `sc_advof` — which moves the pen — knows nothing about pictures and answers **8**. Everything downstream of the pen therefore believes a picture is eight pixels wide, and six defects fall out of that one sentence. The two you will notice first:

- **Text after a picture on the same row is drawn nowhere.** `sc_rowhc` marks a row a picture row when its *first* character is `SC_PICCH` and nothing ends the row *after* the picture, so `sc_rflush` takes the one-blit branch and discards the glyphs the row buffer collected. Insert a small picture, type `hello`, and the text is invisible until you press Enter — the caret moves, selection works, Save writes it, it is just not on the glass. §94.9's "the existing wrap rule ends the row after it with no special case at all" is true only when the picture nearly fills the row.
- **Two adjacent pictures: only the first is drawn**, because the pen advanced 8 and the second is not at `[sc_rowx0]`.

Plus: a centred or right-aligned picture blits past the right margin over the scrollbar; the caret and the click hit test place a picture at 8px; a picture row never updates the row cache; a deleted picture's claim and slot come back only at the next `sc_pictfree`; the `.DOC` writer emits records in **table** order while the reader takes them in **document** order (insert A, then insert B *above* it, and the `.DOC` pairs the wrong picture with the wrong character — RTF is fine, it carries the picture inline); a picture too big for the staging claim is refused as *"Document too complex to save"*; and the reduction and expansion are per-pixel loops with a variable shift, which is seconds on the target for a full-width picture.

**The cheapest place to start is the pen** — making `sc_advof` answer what `sc_penadv` answers for `SC_PICCH` collapses the first five into one change. I did not make it, because it moves every caret, hit-test and alignment decision in the package at once and that is a change to *look at* rather than to infer. §94.11 is the standing list.

One more, smaller: **`\wbmwidthbytes` is emitted odd for an odd-width picture**. A Windows DDB scan line is word-aligned, so a reader that builds a `BITMAP` from what the group says is handed an illegal stride — your own §94.5.2 specimen, `INSTALL.BMP` at 177 wide, writes `\wbmwidthbytes89`. The round trip inside os8088 closes either way, because `sc_rpclaim` requires only `stride >= (w+1)/2`. `.DOC` pads (§94.7.1); RTF is the path that skipped it. Padding is a format change rather than a repair, so it is recorded in §94.11 rather than made.

---

## What I checked and cleared

So it is not re-derived:

- **Every `img_*` call site in `apps/scribe/`** against `main`'s decoder contract, including whether a refusal's zeroed geometry can reach a division, a zero-size claim or a 65535-iteration loop. It cannot: `sc_pictload` reads `IMG_W`/`IMG_H` only on the `jnc` arm.
- **`apps/scribe/scutil.inc` against `apps/word/wdutil.inc`** — genuinely byte-identical apart from the rename. All 40 differing lines are the uppercase half of the prefix change.
- **Every `SCVAR` read inside `.modc`** while DS is not the package — a tracker over all 420 names found exactly the nine sites inside the two windows above and **no third defect**. Your `PKG_*` sweep is otherwise complete.
- **Every `call far [sc_v_*]`** — nine sites, all reached with DS = the package once the two restores are fixed.
- `sc_pictkeep`'s stack and segment discipline (eight pushes, eight pops on every exit including `.no`), `sc_rpoff` being reset per `\pict`, `sc_dpicr`'s `SC_PICMAX` bound, `sc_rpbyte`'s `[sc_rpcap]` clamp, the `PKG_LD*` macros being register- and flag-clean, and no picture claim being marked `OSAPI_MEM_MOVABLE` anywhere — so §66's relocation-proc trap does not apply to any of this.

## Verification

```
make                                    exit 0
python3 tools/os88test.py full          39 passed, 0 failed, 0 skipped (89.7s of a 180s budget)
python3 tools/os88disk.py --verify      all four scribe*.img
tools/os88ovlchk.py                     2 package(s) keep every overlay call far
```

And on the glass under QEMU, end to end: Insert ▸ Picture on a 200×105 4bpp BMP → Save (`DOCUMENT.RTF`, 21,541 bytes) → File ▸ New → File ▸ Open — **the picture comes back pixel-identical**. The same sequence with the name changed to `.DOC` writes 4,729 bytes and comes back **reduced to one bit per pixel**, which is what §94.7 says it should be. Both were re-run after the last fix, not just after the first.

A total of 11 of the review's findings did not survive verification and are not in this list — including one that was priced at 13.7 seconds on a keystroke and turned out to be bounded.

https://claude.ai/code/session_01THkvjaUhidZUNuv8LEj7XY

jggonz:
One correction to the comment above, from the review's own verification pass — pushed as `1d75f69`, and `build/scribe.bin` is **byte-identical** either side of it.

I said the `.RTF` DS bug had a second effect: that a full staging claim would set `[sc_rfull]` in the overlay instead of the package, so the *"it did not fit"* refusal would never fire and the file would truncate silently. **That is wrong.** `sc_rtfimg`'s `.tail` emits `sc_r_par` and `sc_r_tail` with DS back on the package, so `sc_re` sets the package's `sc_rfull` there and the refusal fires regardless. Your refusal path was never broken. The bug's whole cost was the four stray bytes per paragraph — which is plenty, and the fix is unchanged.

While correcting it I added one thing to §94.8.5. Its three verification bullets were taken on a build where `sc_dpapin` still restored DS with `push cs / pop ds`, so the `.DOC` bullet was far-calling `[sc_v_papfind]` through a dword read out of the overlay claim, and the round-trip bullet closed *because* the wrong address was wrong consistently. They are honest evidence about a build that no longer exists, so the section now says that and carries the re-taken run on the fixed tree instead. (Its `rotate-xor`/`BCE9` figure is also pre-merge — `main`'s checksum is the LFSR now, §93.3.)

https://claude.ai/code/session_01THkvjaUhidZUNuv8LEj7XY

PR #172, opened 2026-09-08
Elendilon -> Main (CLEAR SKIES, Compression, Nothing Permanent Pinned on the Heap, App Menu Save-Unders, Parallel Soak Multi-Lane, Bugfixing)

Elendilon:
# DESCRIPTION

The integration branch, squashed onto main. 338 commits, 583 files,
+105,239 / −26,969. Cut from #150 and merged with main's own #151–#166
underneath — **the merge base IS main's tip at time of creation**, #160's `strict near` fix
included, so nothing published upstream is behind this.

## WHAT IS IN IT

- **CLEAR SKIES** (SPEC.md 88, `apps/skies/`). A filled-polygon flight
  simulator: take off, fly, land — or crash, and be put back on the runway.
  **Nine locations in eight worlds** (Cairo-SPX, London-LCY, Miami-MIA,
  Nepal-VNLK, NYC-JFK, Paris-Issy, Paris-LBG, Rio-SDU, SanFan-SFO) and **five
  aeroplanes** (Cessna 172, Pitts Special, Fouga Magister, Wassmer Bijave,
  Icon A5), each a **record in a table** rather than a constant in the code,
  so a tenth of either is a row and not a rewrite. It is TANK ATTACK's shape
  (85) with the other half of the 1983 vocabulary attached — Tank draws
  nothing but lines and this fills the space between them — fsx-exclusive for
  Tank's reason (53.7), and built against one number: twelve frames a second
  on a 4.77 MHz 8088 with a Hercules card. A polygon is filled by its EDGES,
  never by its pixels; a row that is all sky and was all sky last frame is not
  touched; the view is sized to the BUDGET and not to the screen — 320×112 on
  CGA, 320×144 in Mode X, 400×112 in the middle of a Hercules's 640 — and
  **88.15 is a CGA in SIXTEEN colours**, the 160×100 text hack. Measured on
  MartyPC's Hercules over eleven builds: the runway frame 339 → 153 ms, the
  tower — the worst frame in the world, the Eiffel's 32-edge model at its near
  size — 313 → 172. With it: a title page, a Settings page whose five knobs
  are also hotkeys, a five-rung Detail Level, an occlusion pass only Nepal
  asks for, impostors, and `CSDIAG=1`'s watchdog, which answers from ONE
  photograph and says where. Six sources, no worker task, **and no kernel
  change of any kind.**

- **Everything on a shipped floppy is compressed, and so is the kernel**
  (SPEC.md 20.13, 20.14, 2.9.13). `PKGZ ?= lz4`, so a plain `make` packs every
  package, every driver, the on-demand kernel modules, the ten faces,
  README.TXT and the licence; `make PKGZ=` and `NOKZIP=1` are the A/Bs. Two
  codecs (LZ4 ships, LZB assembles and is gated), `OSAPI_DECOMP` for a package
  that wants one, both expanding **in place** against a measured margin, and a
  build that refuses a file whose margin the reader could not honour. A
  compressed file is expanded transparently on read — with 20.14.3's **two
  sizes**, because transparency is not what a raw copier wants: the hard-disk
  installer copies now instead of expanding, and FTPD's LIST, SIZE and RETR
  are one number again (`OSAPI_FILE_FIND_RAW`). File > Compress lands in the
  file manager (22.22), its encoder a 712-byte on-demand module. Measured on a
  cycle-accurate 4.77 MHz 8088 booting the 360KB disk: `KERNEL.SYS` **204 →
  164 sectors** and the whole boot **13,877.9 → 13,278.8 ms**, 47 fewer
  sectors being 1,278 ms of BIOS against 799 ms of decode. `SYS360OMIT` is
  empty and measured — the 360KB system disk carries `PAINT.O88` again at 277
  of 354 clusters — and the 360KB apps disk carries every package at 346 of
  354. `BEVERLY.MOD` is 116,085 bytes packed to 42,177 and **still rides its
  own floppy** (24.4): the arithmetic that was going to collapse that split
  was spent by what else landed on the disk.

- **Nothing on the heap is pinned but an on-demand module** (SPEC.md 66). A
  package's **region** moves and the package keeps working — a worker-owning
  region declares a restart point (`OSAPI_TASK_RESTARTABLE`), and 165,765
  bytes of shipped region plus every C package declare it; a **driver image**
  moves, and the kernel asks the 8086 rather than asking the driver; an
  **overlay** is a CS and takes the top-down door; and `SOUND.DRV`'s image and
  its 8KB DMA ring move, which was the last pinned population. The heap packs
  from the top as well now, in the same walk. **The finding worth more than
  the feature**: the first IVT patch swept all 256 vectors, on the argument
  that asking the 8086 is right about a driver nobody has written yet. The IVT
  is not a table of pointers — its unused slots are scratch, and an XT-IDE
  option ROM keeps two words at int C1h and C3h. One of them read 0x8000, a
  claim moved off 0x8000, the sweep rewrote the ROM's word, and the machine
  came out of a compaction with **no C: drive, silently**. `mem_iv_patch` is
  sixteen slots now — int 08h–0Fh and 70h–77h, what a driver can legitimately
  own — which is a fact rather than a guess; `tests/hdmove.py` caught it after
  `make test-full` and thirteen of fifteen heap rows had gone green. A size
  pass then bought the same feature for a third fewer bytes, +1,207 → +799.

- **A package that draws its own menu bar gets a save-under**
  (SPEC.md 68.2.1, 46.5.1). Word and ArtfulType draw their own bars, so they
  own the dismissal too, and each had only one way to spell it: repaint every
  covered row FULL WIDTH at ~900 µs a glyph cell. `OSAPI_GFX_SAVE` /
  `OSAPI_GFX_REST` are **40 bytes of kernel `.text`**, and on a 4.77 MHz 8088:
  Word's close **521.4 → 19.7 ms (26.5×)**, its round trip 621 → 139.6, and a
  File → Help drag ~4.97 → ~1.12 s; ArtfulType's close **104.9 → 14.5 ms
  (7.2×)** for 250 bytes of the package. Every refusal is the refusal that was
  already there — no claim, a straddling rect, a zero-byte rect — and the
  fallback is the old repaint.

- **The in-window menu becomes a shared ELEMENT** (`apps/os88ui.inc`,
  docs/plans/UI-MENU-ELEMENT.md), in three waves: the record and geometry, the
  drawing and the hit test, then the gesture, the bank and two hooks. Word
  drew its own nine-title bar because `MENU_APPMAX` is five, and Sheet drew a
  second one whose own comment says it follows `wd_mtrack`'s pattern — a body
  copied by hand, which is what `os88ui.inc` exists to stop. Word's fields are
  **aliases into the element's record**, not copies, so the control half moved
  out a routine at a time with all hundred-odd references still working. A
  drop-down joins it; Clear Skies' Settings page is its first user, and Word's
  ruler Style and the ribbon's Font and Pts combos (68.2.3) are the second.
  A Word scroll no longer repaints the chrome (68.2.4).

- **The soak runs in parallel, and the tiers are recut to what each is FOR**
  (docs/plans/SOAK-PARALLEL.md, `tools/os88soak.py`). Rows declare their
  artefacts and read a **private tree**, so `build/` stays the operator's
  while a soak runs and the serial lane falls from 41 rows to 3. The theory
  that contention would break rows was measured and overturned: twelve
  emulator rows at width 3 **with two extra CPU hogs** pass 12/12, the sum of
  row times up 1.06×, the wall clock **892 → 486 s**. `fast` is cut to what a
  writer who did not touch the subject can break — 55 rows → 25, 17.3 → 8.8 s
  wall, before main's own rows bring it back to 33 — and `full` to "did you
  obviously break the OS", 14 rows → 5 under a 3-minute budget. Nothing is
  deleted; every retired row is one `-k` away.
  docs/WRITING-TESTS.md is the rules, and it is made of the eighteen bugs that
  produced them: 28 rows fixed behind the harness layer, then four harness
  causes behind 60 of a 76-failure soak, and a width-4 run at 294 of 296 with
  neither of the two this tree's.

- **Redraw, elsewhere.** Paint's tool palette is composed a ROW at a time and
  blitted once — 24 drawing calls become 4, 111 → 33 ms, and a fresh window's
  paint 337.7 → 258.5 (42.26.1) — and its one-bit canvas reaches `gfx_blit1`
  at any width (5.4.2.5). Tank Attack goes **5.2 → 13.9 fps on Hercules**
  over two measured rounds (85.3.4). ArtfulType's keystroke is five separate
  wins: a straight-line emitter for the unstyled cell, a raw parse for a line
  with no markup, the LAYOUT proving a line plain because it is already
  reading the bytes, the erase being the text's tail rather than the region,
  and an append repainting one line — 1.87×, 2.29× on six. Fractal answers
  the Mandelbrot's interior **without iterating** (40.5–40.7): the two closed
  forms in the core's own Q4.12, at most four `imul` against the 144 a capped
  pixel costs — default view 120.0 → 49.4 s, zoom 3 **361.3 → 8.2**.

- **Six that came off real machines.**
  - **Every key on a 286 typed a different character** — `f` typed `\`. Bit 5
    of the 8042 command byte is the auxiliary clock on a PS/2 controller and
    **PC MODE** on an AT one, where it stops set-2 → set-1 translation; the
    probe's `.fail` path restored a banked byte with bit 5 forced set. One
    bit, latched for the session (FIELD-NOTES 39, SPEC.md 9.9.7).
  - **Every menu title on the bar dropped the System menu on any machine
    booted from an installed hard disk.** `hb_ok` reaches `OSAPI_XM_CAPS`,
    which answers in BL as well as AX and DX:CX, and saved BX alone — so CX
    came back 0, every press hit-tested at x = 0, and no floppy-booted machine
    could show it.
  - **Note Pad froze on a scroll UP.** 27.7.6's "does this need the count?"
    test was unsigned and every arm feeding it is relative, so the up arrow at
    top 0 asked for row −4 and was read as 65,532: 1,053 ms of background
    count, and it scrolled nothing. **1,053 → 20.2 ms.** A track click no
    longer takes the scroll bar off the screen either.
  - **A restart parks the floppy heads** (18.100) — 31 bytes of `.cold` and 5
    of `.text`, so 18.97's probe does not inherit a head above cylinder 77 and
    read the equipment-check signature it retires a drive on.
  - Clear Skies took **eleven reports off the glass** in two field rounds.
  - `kern_emu` is a **third build** (9.11.7): the VMware absolute pointer's
    resident half leaves `kern_big`, because the probe itself is 386 code and
    the XT this project is calibrated against cannot even ask the question.
    385 bytes of image and **one crossed 512-byte rung** off every XT.

- **docs/ is a taxonomy rather than a directory.** Ninety-five documents sat
  flat, so every one read as current and the filename was the only signal.
  80 files moved and 265 references rewritten into `docs/` (how it works
  today), `docs/plans/` (open), `docs/plans/completed/` (the record behind
  something that shipped) and `docs/history/` (superseded); docs/README.md is
  the rule, including the instruction to move a plan to `completed/` in the
  commit that lands its work. Then an audit, document by document, against the
  tree — which closed the FIELD-NOTES this tree has fixed, kept the seven that
  are open, and found that C64-SPEC's wave 4 never landed.

- **And the round before this one, which never went up:** ArtfulType styled on
  a 1bpp adapter (46.2.1–46.2.3), `OS88NET.COM` as a self-extracting archive
  (62.12, 19,333 → 11,653 bytes), the Task Manager's small build on both small
  floppies (28.12, 6,693 against 11,138), `BROWSER.HTM` replacing `DEMO.HTM`
  (71.12), and the typefaces moved into `SYSTEM/FONTS` (19.8.1).

## TWO THINGS A REVIEWER SHOULD LOOK AT

**1. The API table collided at 0x04F8 and 0x0500, and merged CLEANLY.** Both
sides appended to the same tail in the same round, so `apps/os88api.inc` came
out with **no conflict marker and two names at one address** —
`OSAPI_DECOMP`/`OSAPI_PKG_RUN` and `OSAPI_FILE_FIND_RAW`/`OSAPI_DESK_SVC`.
This is the exact failure docs/UPSTREAM.md names, and the check it carries is
what found it. The branch keeps 0x04F8–0x0518 and **main's two slots move to
0x0520 and 0x0528**; UPSTREAM.md's collision check comes back clean on the
result. Twelve bytes of `.text` is what the move cost. Anything built against
main's old numbers — The Wire's `.o88`, any `tests/` binary on a scratch image — must
be rebuilt, which is UPSTREAM.md's renumbering warning arriving on schedule.

**2. SPEC.md 88 collided: CLEAR SKIES on the branch, THE WIRE on main.** Per
the ModPlug precedent (#58) the incoming section renumbers: **THE WIRE is 92**,
and 272 citations move with it — including seven outside the conflict region
that would otherwise have resolved silently onto Clear Skies' subsections.
PERFORMANCE.md Set 116 collided the same way and PACCMAN's is Set 118. Two
citations survived the first sweep (README.md's Wire section and 24.3.1.1) and
`checkdocs` could see neither, because both numbers name a real heading; they
are fixed in `d3ffb10`.

## MAIN'S OWN WORK IS PRESERVED

Three kernel collisions, and **two of the three merge cleanly and wrongly**,
which is why this was not a `--theirs`:

- the API table, above;
- **`gfx_blit1`'s stack frame.** The branch added 5.4.2.5's masked tail byte
  (any width) and main added 5.4.2.2.1's Map Mask split, both on the same
  frame and both claiming BL. The merged frame is nine words, BH carries the
  tail mask and BL pass 2's Map Mask, and the pen's fourth refusal goes with
  main's split;
- **`loader.inc` was restructured underneath.** Main split `ld_alloc`/
  `ld_start` out for `OSAPI_PKG_RUN`; the branch's compression is ported onto
  main's `.read` rather than left in the pre-split copy git kept.

`SMALLSYSAPPS` filters `SMALLOMIT`, so The Wire stays off a kernel with no
NIC; `FACESARG360` takes `SYSTEM/FONTS:`; `os88index` gains a "Packages and
the desktop" group for main's two slots; and `apps/cc/os88.h`'s three counts
are re-counted rather than incremented, to 101/160/129.

**CLAUDE.md is main's copy**, carrying only this branch's knobs and nine
paragraphs chosen one by one — the fork's own "Working in this fork" section
stays on the integration branch and is not in this PR.

## WHAT IT COST THE KERNEL

`docs/reports/KERNEL-BYTES-SINCE-SQUASH-2026-09-07.md` measures all four
points — the merge base, this branch's tip, main's tip and the merge — with
`kernsize --json`, which re-assembles rather than reading a build.

| `kern_big` | branch's arm | main's arm | merged |
|---|---:|---:|---:|
| `.text` | **−44** | +90 | +58 |
| `.bss` | **−64** | +74 | +10 |
| `.cold` | +1,234 | +424 | +1,657 |
| **`KERN_SIZE`** | **+1,536** (3 rungs) | +512 | +2,048 |

Everything above — compression, the heap work, the save-under slots, the
floppy park — fits in `.cold`, and `.text` and `.bss` **shrank**. On
`kern_small`, the 128KB floor machine, the whole +1,536 is this branch's:
main's nine features cost that build 256 bytes of section and **not one byte
of footprint**. `KERN_BUDGET` is 129,536 and `KERN_SMALL_BUDGET` 107,520 at
every point — nothing here is a budget move.

## GATES

At the tip: `make` exit 0, `checkdocs` 0 problems, and the fast tier **33
passed, 0 failed**. The last full-tier run on this lineage is the Clear Skies
merge — `make test-full` **31 passed, 0 failed, 1 skipped in 49.1 s** — and
the merge with main before it ran 34 passed, 0 failed, with `ctoolchain` and
`ps2mouse` skipping for want of SmallerC and QEMU on that box. The heap work,
Clear Skies, the compression waves and the parallel soak each carry their own
rows in `tests/suite.py`; `python3 tools/os88soak.py start` is the whole tier.

No binary is added: `git ls-files build | wc -l` is 0.

https://claude.ai/code/session_019hecoDF8DZ1eLwXJYE7CwY


jggonz:
Starting the maintainer review now — fetching the branch, building both kernels, and running the review over the units this PR splits into (Clear Skies, the compression path, the heap unpinning, the save-under slots, the shared menu element, and the field fixes).

Given the size (583 files, 338 commits) I'll be selective: the review weight goes on memory safety, anything lost from `main` in the merge, the redraw budget, and disk/format integrity. I'll post findings and any fixes back here when it's done.

https://claude.ai/code/session_01Cf6sn5PaQJL4Hvmfyv7tuV

jggonz:
## Maintainer review — merged nothing, pushed five commits, the PR is mergeable

`main` had not moved since you cut this, so there was no merge to do: the merge base **is** `main`'s tip `ed97745e`, and GitHub agreed (`MERGEABLE` before I started). I checked `main`'s five most recent commits are all intact in your tree anyway — #159's imager tools, #160's `jmp strict near`, #163's `mov cx, [mouse_x]` in `menu_track`, #165 PaccMan, #166's Telnet BBS — because a long-lived branch losing something `main` added is the failure git reports nothing about. It lost nothing.

The API table cross-check is clean: all **155** slots `main` published keep both their address **and** their name, five new ones appended, no duplicate addresses. Your 0x04F8/0x0500 collision write-up is exactly right, and the resolution is right.

Eight reviewers went over this by subsystem, then a second agent tried to **refute** every finding they raised. 32 findings judged, 29 survived, 3 refuted. I read the two blockers myself in the source before calling them blockers. Below is every fix, then what I left alone.

---

## Fixed

### 1. A drop-down's save-under buffer was sized from two registers that had already been destroyed
`apps/os88ui.inc` · `f3273967`

Open a list with more than two or three items and the OS would write **hundreds of bytes of screen pixels past the end of a heap claim**, straight through whatever block came next — a corruption with no message and no crash at the time, surfacing later as some other program misbehaving.

`os88ui_drrect` hands back x2 **in CX**, and the next instruction is `mov cl, 3` — which overwrites x2's low byte before `shr cx, cl` reads it. Separately, the plane byte-count was parked in CX and then `OSAPI_WM_DISPLAY` was called, which returns **the dock's first row** in CX; the code then multiplied the dock row by the plane count. The claim came out ~1–2 KB whatever the list was, and `OSAPI_GFX_SAVE` wrote the real `planes × rows × bytecols` into it.

It survived review because of a coincidence worth knowing about: SPEC.md 13.14.1's measurement is CLEAR SKIES' **two-item** Location list, and two items × 12 rows + 2 lands on the same 1 KB the broken arithmetic produces. Every longer list overran — including Word's ruler Style and the ribbon's Font and Pts combos.

The fix banks x2 in DI before the shift count lands, and brackets `OSAPI_WM_DISPLAY` with `push ax`/`pop ax` — which is exactly the shape your own `os88ui_mnbank` uses two screens down, and `atui.inc`'s `at_surban`. Those two got it right; this one is the odd one out.

### 2. The whole socket surface far-called into a driver image with no `wm_nest` bracket
`kernel/driver.inc` · `f3273967`

While a package was inside `ETHER.DRV` — which is **every socket call** the Browser, Telnet, FTPD and The Wire make — nothing pinned the driver's image, so a heap compaction could relocate the code the package was executing in, and the return would land in whatever moved into its place.

You gave the bracket to every other kernel→driver site in this PR: `drv_call`, `drv_svc_call_x`, `drv_blk_call_x`, `drv_fs_call`, `drv_cp_call_x`, plus `xmem.inc`, `vmmouse.inc` and `wm.inc`. `drv_pkg_call_x` — the `OSAPI_DRV_CALL` cell — is not on that list. And it is the one that most needs it: `mem_can_move`'s `.drvimg` arm asks only `mem_in_nest` (the stack this site never pushed onto) and `[drv_wcnt]`, which is permanently **0** for ETHER because that driver has no worker at all. The cell takes no lock by design, so the task is fully pre-emptible in there.

One wrinkle worth flagging back: this is the single site where **ES is not `KERNEL_SEG`** at the push — it is the calling package's, and the driver reads its buffer through it. ES is borrowed for the ES-relative push and handed straight back, and SPEC.md 66.6.3's "ES is KERNEL_SEG at all N sites" prose was updated to say so, along with the same counts in `wm_nest_pushd`'s and `mem_in_nest`'s headers. **16 bytes of `.text`, no rung crossed.**

### 3. The nest stack itself was never initialised
`kernel/wm.inc` · `9ed3b92b`

`wm_pkgd` and `wm_pkgs` were added at the **head** of `wm.inc`'s `.bss`, immediately before `wm_wins` — but `wm_init`'s `rep stosb` starts **at** `wm_wins`, so the 18 new bytes are outside every run it makes. The routine's own comment four lines down says why that matters: *".bss is never zeroed at boot, so this is not redundant."* And `int 19h` keeps the previous kernel's `.bss`, so a restart carries the old value straight in.

The consequence runs both ways. A leftover depth above `WM_PKGD_MAX` makes the compactor pin every package region and driver image for the whole session — the heap never compacts again, and a launch that would have fitted refuses for want of memory, which is your headline feature silently doing nothing. A leftover depth below it has `mem_in_nest` compare **garbage segments**, which can let a driver image move out from under a live return address.

The fix moves the run's start back to `wm_pkgd` and grows the count by 18. Same two instructions, same encodings — the kernel is byte-for-byte the same size. The section header comment that said "these three must stay contiguous" now says five; that is the comment that went stale when the pair was prepended.

### 4. A corrupt package header could make the loader read 65,024 bytes past its own region
`kernel/loader.inc` · `e9390901`

`ld_check_hdr` sets `[ld_cfmt]` from the compression bits **unconditionally**, on every arm of the size ladder — but the bits are only *validated* on the `.over` arm, where `image > file`. Three combinations reach step 6 with `[ld_cfmt]` non-zero and `image <= file`, and step 6 computes `R = roundup512(image - file)` with a plain 16-bit `sub` and no borrow test. It borrows, `ld_roff` becomes up to 0xFE00, and `dsk_read_chain_x` writes the file well past the claim.

A single flipped byte on a floppy, or a `.o88` from someone else's disk, reaches this — and the failure looks nothing like a disk error. It answers **"Bad package"** now. The three arms route through one shared `.notover` label testing **both** bits (0x18, not just bit 3 — bit 4 alone already makes `[ld_cfmt]` non-zero and takes the same path), which is why it cost **+7 bytes of `.cold`** rather than +21. 450 bytes still left in that rung.

### 5. Fractal's mirror was not firing at all, and one row per render was painted from the wrong data
`apps/fractal/fractal.asm` · `7a70d020`

Two symptoms of one design collision, and the second is visible.

`fr_take` had a single `.none` exit shared by three very different cases, and two of them — no cache claimed, and the ordinary "the cursor is at the frontier, compute it" — parked `0FFFFh` in `[fr_lrow]` even though they had never touched `fr_line`. Since `fr_worker` calls `fr_take` and then `fr_twin` in that order, the common case wiped the sentinel in on the line immediately before `fr_twin` compared against it. **SPEC.md 40.6's mirror never fired**, so the default view took the ~55 s the table says it takes *without* the mirror, not the ~30 it advertises — 25 seconds of extra wait per frame on a 4.77 MHz 8088, and the documented speed-up did not exist.

And `0FFFFh` is not a safe sentinel here, because `2*rc - row` **is** −1 whenever `row == 2*rc+1` — an ordinary row of the walk. So on any view whose mirror axis sits above the middle of the canvas (a click below centre reaches one), one row got painted *and cached* from an unrelated row's colours: a wrong stripe that survived every repaint.

The fix splits `.none` from a new `.quiet` that leaves `fr_line` alone, and range-tests the twin against `[fr_ch]` before comparing. I verified the render on the glass afterwards — the Mandelbrot comes out **99.71% mirror-symmetric about y=173 with no row below 0.90**, i.e. the mirror is firing and nothing is stripey.

### 6. Five section numbers that every gate passes over
`922b4b40`

These all *resolve*, which is why `checkdocs` is happy and why they're worth a commit: each names a **real heading**, just not the right one.

- **The Wire is §92 now** and five citations still said §88 — which is CLEAR SKIES. `apps/thewire/thewire.asm`'s own banner, three Makefile comments, the release skill. You caught two of these in `d3ffb10`; these are the rest.
- **The save-under pair is §5.3** — *"The save-under pair is published, slots 0x0508 and 0x0510"*. Three places cited **§5.4.3**, which is `gfx_blitp`: `apps/os88api.inc`'s banner for the pair, `kernel.asm`'s `api_gfx_save`/`api_gfx_rest` header, and SPEC.md's own 68.2.1 paragraph. Word's `wd_subank` already cites §5.3 correctly — so the one pointing at the wrong routine was the SDK header a package author reads first.
- **CLAUDE.md's `SPLSTARS` row** said the blob problem *"is not any more (§15.3.8.5.1)"* while `make SPLSTARS=1` now refuses without `NOKZIP=1`. The `NOKZIP` row three lines below already said so; the two rows contradicted each other.
- **`docs/plans/HEAP-UNPIN-PLAN.md`** still opened *"Status: RESEARCH. Nothing here is built"* — this branch's headline feature is that plan built, with four soak rows gating it. Status line fixed; the file stays in `docs/plans/` because §12's open questions genuinely are open.
- **`docs/UPSTREAM.md`** gains #172 beside #58, #92 and #144: the §88→§92 renumber and the 0x04F8/0x0500 API collision. That list is where the next session looks, and your own collision check is what found it — worth recording that it worked.

---

## Found, verified, and deliberately **not** fixed

Both of these are real. Both need your design intent, so a fix would have been me guessing.

### `mem_in_nest` rests on one global stack that pre-emption can leave out of order

`wm_nest_pushd` stores at `[wm_pkgd]` and increments; `wm_nest_pop` is a bare `dec` and is passed **nothing**, so it cannot know which level is its own. Nothing serialises the nine bracket sites — `drv_svc_call_x`, `drv_blk_call_x` and `drv_pkg_call_x` take no lock and run with IF=1. Two tasks inside two different images, **outer returning first**, leaves the inner one's entry above `[wm_pkgd]` and invisible to `mem_in_nest` while its frame is still standing in that image.

I tried the conservative repair (clear the popped slot, shrink only while the top is empty) and it **inverts on exactly the hazardous case**: when the outer task returns first, the top slot belongs to the inner task still executing, so clearing it *un-pins a live image* — it turns a missed pin into a guaranteed one. Never shrinking leaks the other way, and the depth never returns to zero, which is the "silently does nothing" shape SPEC.md 66.6.1 names as the worst outcome available.

The real repairs are all design changes: pass the image segment to the pop and clear the matching slot (a signature change at nine sites, several with no spare register and relying on the pair writing no flag, plus a 66.6.1/66.6.3 contract change); move the stack into the task record; or take a lock, which `drv_pkg_call_x` explicitly may not (SPEC.md 20.6 rule 7). Worth noting `[sch_lock]` does **not** close this — it is raised across `mem_compact`, but the out-of-order return happens *before* the compaction, so the stale entry is already invisible when the walk starts.

The question underneath is whether the nest is meant to be machine-wide or per-task. That is yours.

### ETHER's socket pool is movable at every instant

`sk_try` calls `OSAPI_MEM_MOVABLE` on the pool. Its `MC_OWN` is the driver's own segment, so `mem_can_move` takes `.notslot`, the `MC_DMA` head is 0 for a top-down claim so the `[drv_wcnt]` gate is skipped, and `mem_busy_seg → inst_seg_parked → inst_svc_parked` answers "free" because ETHER has no service task. **Fix #2 above does not reach this** — the nest bracket pins the driver *image*; the pool is a separate claim, and `mem_busy_seg` must not consult the nest (SPEC.md 66.6.3 fact 2, in capitals in its own banner).

The exposure is also wider than `sk_reloc`'s header allows for. Its argument covers four sites holding `[sk_seg]` across a `rep movsb`, but the same window exists between any `call sk_ring` and the last access through the ES it returned — `sk_rxput_i` holds ES across two `rep movsb` runs plus a dozen record updates, and `sk_rxget_i` loads `[sk_seg]` into DS and may not name a driver variable again for the rest of the routine. That rules out the *same* task, not another one claiming while this one is pre-empted.

Three candidates, with the arithmetic:
1. **Un-declare the pool** — one line, but it reverts the stated purpose. The pool is 8,192 + 1,024 + 3 × (1,024 + 512) = **13,824 bytes**, and HEAP-UNPIN-PLAN §2.0's whole case is that it is the largest single block in the ceiling wall.
2. **Have `.notslot` ask `mem_in_nest` on `MC_OWN` for a driver-owned claim** — about a dozen bytes and it closes this exactly, but it rewrites the rule 66.6.3 fact 2 states, and it would pin a driver's other data claims during any compaction that driver itself triggers (`sbl_v_grant` claims from inside `SOUND.DRV`, so the depth is non-zero at precisely that moment).
3. **Have the driver pin itself** with `OSAPI_MEM_MOVABLE AX = 0` — but guarding only the copies is not enough, so the bracket has to sit at every socket verb's entry and exit with a nesting count: two far calls (2 × 46.7 µs) plus two `mem_find_own` scans of 32 records per verb, on the FTP hot path.

Each changes either a stated SPEC rule, this PR's declared intent, or a measured hot path.

---

## Also flagged, not touched

- **`heapcheck` has never run here.** It dies on `sys.path.insert(0, "/home/user/os8088/tools")` before importing anything. To be clear, **that is `main`'s, not yours** — the path predates this PR. But `heapcheck` is *the* thirteen-check gate for compaction, which is this PR's largest subject, and every sibling row (`hdmove`, `heaphi`, `regmove`) already derives its path from `__file__`. Two lines, and your compaction work gets its most direct gate back.
- **`kern_small` crosses a cold rung** (+512 bytes of every machine's RAM). I measured it at your pristine head to be sure it wasn't ours, and it reads identically there. Both budgets still hold — 27,648 spare on small, 17,408 on big, `KERN_BUDGET` untouched.
- **The `kernsize` baseline in `docs/KERNEL-MEMORY.md` is stale** — `+1 .text`/`+15 .ovlw` on big and `+694` on small. I deliberately did **not** run `--bless`: it would fold your unblessed growth into a baseline written by our commit, and that figure is yours to bless.
- **A latent overflow in both save-under sizings.** `os88ui_mnbank` and `atui.inc` both discard DX on the `mul bx` by the plane count. The one-plane guard comment is right, but it's the **×4** that overflows, at a quarter of the size the comment argues about — a rect needing more than 16,383 bytes per plane wraps. Not reachable today (your largest dialog, `wd_dlgpara` at 448×134, needs 30,780 of 65,536), so I left it; worth a `jc`/`or dx,dx` if a bigger panel ever banks.
- **One stale comment above the driver.inc edit**, left alone as out of scope: `drv_pkg_call_x`'s header still says *"which is why the five sites below name `drv_fptr`..`drv_fptr5` outright"*, but you collapsed those five into the single `drv_pkg_disp` this PR adds.

The 20 remaining confirmed findings — mediums and lows across the compression hint refusals, `gfx_blit1`'s masked tail byte on VGA/EGA, the published `MN` record template's offsets, and a handful of header contracts that no longer match their code — are yours to take or leave; say the word and I'll write them up in detail or fix them.

---

## What I ran, on your tree

- `make` **exit 0**, fast tier **33 passed / 0 failed**
- `make KERN_SMALL=1` and `KERN_EMU=1` — all three variants assemble, no rung crossed on `kern_big`, `KERN_BUDGET` and `KERN_CODE_MAX` untouched
- **`make test-full`: 38 passed, 0 failed, 0 skipped** (I staged MartyPC and SmallerC into the worktree so nothing skipped)
- `checkdocs` **0 problems**; `os88disk --verify` clean on all **nine** shipped floppies; no duplicate `§` headings; `git ls-files build` = 0
- The gates your features own: `reg*`, `drvmove`, `sndmove`, `hdmove`, `heap*`, `mseg*`, `lz*`, `kz*`, `cs*`, `skiesui`, `drvovl` — **all green**. `regapp` and `regmove` failed in the wide run and passed alone; that is the documented MartyPC double-click-under-load flake, and they failed the same way before any of these fixes.
- **On the glass**, before and after: boots to a desktop from the compressed kernel, `SKIES.O88` launches out of a compressed package, the Plane drop-down opens and its dismissal restores **pixel-identical** (only the clock and the cursor differ), Fly reaches the Paris-Issy runway, and Fractal renders 99.71% symmetric with no rogue row.

This is a big, careful piece of work and most of what I found is at the edges of it — the two blockers are both "one site out of nine didn't get the treatment the other eight did", which is the kind of thing only a second pair of eyes finds. Nice PR.

https://claude.ai/code/session_01Cf6sn5PaQJL4Hvmfyv7tuV

PR #177, opened 2026-09-10
Scribe: read SCRIBE.OVL at entry, so a disk swap cannot strand Open and Save

Koriban:
**A regression from #162 (mine): a freshly launched Scribe cannot open a document from a second floppy, or save a new one to a blank disk.** This loads `SCRIBE.OVL` at entry instead of at first use. 19 resident bytes.

## What happens

§94.8 moved Scribe's file formats into `SCRIBE.OVL`, and §68.10's design reads a module "the first time one of its features is asked for" — from the folder the package was **launched** from (`sc_ovdir`/`sc_ovdrv`). For WORD's module, a picture feature most sessions never touch, that costs nothing. For Open and Save it means the first use is the first file operation, and by then the launch drive typically holds the disk the user is opening from or saving to:

1. Launch Scribe from its disk in B:.
2. Swap a data disk into B:.
3. File ▸ Open a document on it (or type something and Save onto it).
4. `sc_ovneed` goes to the launch folder, finds no `SCRIBE.OVL`, and refuses.

That refusal is the ordinary path §68.10 describes — "a disk swapped for one without" the module — and for a picture it is a missing feature. For Open and Save it is an editor that cannot read from or write to a second floppy. WORD does not have the problem: its formats are resident.

## The change

- `sc_entry` calls **`sc_ovload`** after its three claims and before the window — §50.3's claims-at-entry.
- `sc_ovload` is `sc_ovneed` without the voice: CF, and the message in AX instead of posted. `sc_ovneed` becomes the wrapper that posts, so its one caller (`sc_ovcall`) and its contract are unchanged.
- The entry call is silent and reads no CF. Without heap or file the editor still runs, and the first file operation retries and says why — exactly as today.

## Verified

A/B under MartyPC on a 5150, cold launch, no file touched — `main`'s binary against this one, reading `mem_tab` through `tools/heapmap.py` and the floppy controller through `Marty.disk()`:

| | Scribe's claims | `SCRIBE.OVL` resident | launch floppy traffic |
|---|---|---|---|
| `main` | 3 of 8 | no | 6 reads, 87 sectors |
| this | 4 of 8 | **yes**, 12K pinned | 16 reads, 141 sectors |

Once resident, `sc_ovneed` returns at its first compare and touches no disk, which is what makes a swap harmless.

**The swap itself was not driven.** MartyPC's debug server has no mid-session mount, and `os88mouse` is MartyPC-only, so no harness in the tree can change a floppy under a running package. The claim above rests on the A/B plus the fact that the load path is never entered once `[sc_ovseg]` is set.

Fast tier green, including `stkbalance` and `ovlchk`; checkdocs clean. SPEC §94.8.6 records the change and the numbers.

## The cost, and something it exposed

Every launch now pays those **54 sectors**, including sessions that never open or save; before, they fell on the first file operation of the sessions that did.

54 for a 20-sector file is worth a look on its own. The load navigates to the launch folder and back and reads a cluster at a time — the §18.91 shape `Marty.disk()`'s docstring describes. That is the module load path, not this change, and every overlaid package pays it; I have not touched it here.

https://claude.ai/code/session_01Gn9xNSCBfSvQxxiUoKv328

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -107874,6 +107874,41 @@ from `.RTF` (21,541 bytes) and **reduced to one bit** from `.DOC` (4,729),
 which is what §94.7 says each should do. The checksum bullet's `rotate-xor` is
 also `main`'s **LFSR** now (§93.3); the `BCE9` figure is the pre-merge one.
 
+#### 94.8.6 Loaded at entry, not at first use
+
+§68.10 reads a module "the first time one of its features is asked for", and
+for WORD that cost nothing: its module held a picture feature most sessions
+never touch. **§94.8 put Open and Save in this one**, so "first use" became the
+first file operation — and `sc_ovneed` reads the module from the folder
+Scribe was **launched** from (`sc_ovdir`/`sc_ovdrv`). By the time a user
+opens a document off a data disk, or saves a new one onto a blank one, the
+launch drive holds that disk and not `SCRIBE.OVL`. The refusal was the
+ordinary path §68.10 describes — "a disk swapped for one without" the module —
+but for a picture that is a missing feature, and for Open and Save it was a
+Scribe that could neither read a document from a second floppy nor write one
+to it. WORD never had the problem: its formats are resident.
+
+`sc_entry` now calls **`sc_ovload`** after its three claims and before the
+window, which is what §50.3's claims-at-entry asks for. `sc_ovload` is `sc_ovneed` without the
+voice: it returns CF and, on failure, the message in AX rather than posting
+it, and `sc_ovneed` is now the wrapper that posts. The entry call is silent
+and does not read CF — a machine without the heap or a disk without the file
+still runs the editor, and the first file operation retries and says why.
+
+Measured under MartyPC on a 5150, cold launch, no file touched:
+
+| | Scribe's claims | `SCRIBE.OVL` resident | launch floppy traffic |
+|---|---|---|---|
+| before | 3 of 8 | no | 6 reads, 87 sectors |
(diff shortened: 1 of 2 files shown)


jggonz:
Reviewed and verified. No changes needed, so I pushed nothing to the branch. It's current with `main`, it merges cleanly, and `make` (fast tier 34/34) is green.

**The code.** Splitting `sc_ovneed` into the silent `sc_ovload` plus a wrapper that speaks is correct. Every exit from `sc_ovload` balances its pushes. The worker path still leaves AX alone for `sc_ovneed`'s callers. If `sc_entry` bails out after the new call (window table full), the loader's `ld_unreserve` frees the overlay claim along with the other three, the same as before.

**The swap you couldn't drive.** QEMU can change a floppy mid-session (`change floppy1 ` over QMP), so I drove the exact scenario from your description. I launched Scribe from its disk in B:, swapped in a disk holding only `WELCOME.DOC`, and ran File ▸ Open on it:

| | result |
|---|---|
| `main`'s Scribe | an **empty** document titled `WELCOME.DOC` |
| this PR | `WELCOME.DOC` opened and rendered in full |

So the regression is real and this fixes it.

**Flagged, not fixed (it's on `main`, outside this PR).** The `main` row above is worse than a refusal. When Open can't load the module, the window still takes on the name of the file it failed to read, with nothing in it. A Save at that point would presumably write that empty document over the user's `WELCOME.DOC` (I didn't try the Save). After this PR, that can only happen when the load at entry also failed, meaning no heap for the 12K or no `SCRIBE.OVL` beside the package. But it's worth a follow-up so the failed Open leaves the title and document as they were.

PR #178, opened 2026-09-11
Sheet: a saved file said what the sheet did not — stale values and BIFF error codes

Koriban:
**Two ways a file Sheet saved disagreed with the sheet it came from.** Both were invisible from inside, because Sheet read its own files back the same wrong way. Two commits, one SPEC section (§81.22), 139 bytes.

## Measured on this tree's build

A 5150 under MartyPC, checked with a host-side reader that shares no code with Sheet:

| | before | after |
|---|---|---|
| `=A1*1` … `=A1*10`, opened and saved as SYLK | `1 2 3 4 0 0 0 0 0 0` | `1 2 3 4 5 6 7 8 9 10` |
| an error constant, saved as Normal | error byte **02H** (not a BIFF code) | **07H** |
| `=1/0`, saved as Normal — its cached result | **02H** | **07H** |
| Excel's `#DIV/0!` (07H) as a constant, opened | **`#N/A`** | `#DIV/0!` |
| Excel's `#DIV/0!` as a formula's cached result, opened | **`#N/A`** | `#DIV/0!` |

## 1. A Save recalculates first (§81.22.1)

Evaluation is lazy: `sh_eval_cell` runs when a cell is read, and a repaint only reads what's on the glass. The SYLK and BIFF writers read the **stored** value (`sh_cellval_to_acc_si`), while DIF goes through `sh_getcell2`, which evaluates. So a formula that has been off-screen since it was loaded, or since a cell it names changed, is saved with whatever it last held — after a load, the zero `sh_setformula` leaves. That's the first row of the table: the four formulas on the glass saved correctly, and the six below them as 0.

`sh_dowrite` now calls `sh_recalc_all` before choosing a writer. It decides nothing itself: `sh_eval_cell`'s pass stamp already knows what's stale, in both calculation modes (automatic advances `sh_pass` on every repaint; manual doesn't, so only never-computed cells run). Each record is evaluated as its own sheet, using `sh_rowcol_op`'s impersonation, because `sh_findcell` packs `[sh_cursheet]` into every reference. 64 bytes.

## 2. Error codes in the file's own numbering (§81.22.2)

This is the gap `docs/BIFF-NOTES.md` already records, fixed at exactly the four places it names. `SH_C_AUX` numbers errors as ERROR.TYPE does (1–7); BIFF's error byte uses `00 07 0F 17 1D 24 2A`. `sh_biff_e2b` converts on the way out (`.aserr`, `.errresult`) and `sh_biff_b2e` on the way in (`.isboolerr`, `.isformula`). An out-of-range code in either direction becomes `#VALUE!`. The `SH_ERR_*` comment and BIFF-NOTES now describe what's done rather than what's missing. 75 bytes.

## Checked

Fast tier green, checkdocs clean. Each row of the table was measured before and after on this branch's own build, and the first commit alone fixes row 1 and leaves the rest as they were, so the evidence tells the two fixes apart.

These come from the fork, where Sheet has moved further ahead: its §81.38 and §81.48 are the same two fixes, and it gates them with a SYLK/BIFF round-trip test. That harness isn't in this PR; the table above is its output on this tree.

https://claude.ai/code/session_01Gn9xNSCBfSvQxxiUoKv328

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -96661,6 +96661,36 @@ overrun corrupts a value; this one corrupts an *address*, and the next large
 it is worth knowing that bss adjacency can arrange it without anyone loading a
 segment register wrongly at all.
 
+### 81.22 A saved file said what the sheet did not
+
+Two ways a file SHEET wrote disagreed with the sheet it came from, each
+invisible from inside because SHEET read its own files back the same way.
+
+#### 81.22.1 A Save recalculates first
+
+**Evaluation is lazy.** `sh_eval_cell` runs when a cell is *read*, memoized
+against `sh_pass`, and a repaint reads only what is on the glass. That is the
+right shape for a display and the wrong one for a file, because two of the three
+writers never ask: the SYLK and BIFF writers read a cell's value with
+`sh_cellval_to_acc_si`, the **stored** double, while DIF reads through
+`sh_getcell2`, which evaluates. So a formula off-screen since it was loaded, or
+since a cell it names changed, was saved with whatever it last held — after a
+load, the zero `sh_setformula` leaves. Ten formulas `=A1*1` … `=A1*10` on a
+5150, opened and saved as SYLK: the four on the glass came back right and the
+six below them came back **0**.
+
+`sh_dowrite` now calls `sh_recalc_all`, which walks every record and calls
+`sh_eval_cell` on each formula. It decides nothing: the pass stamp already knows
+what is stale, in both modes — automatic advances `sh_pass` on every repaint, so
+anything not recomputed since recomputes; manual does not, so only a cell never
+computed (stamped `0xFFFF`) runs. **Every sheet is evaluated as itself**, the
+`sh_rowcol_op` impersonation, because `sh_findcell` packs `[sh_cursheet]` into
+every reference. 64 bytes.
+
+What it costs is time at Save for a sheet whose formulas were not computed at
+paint, and nothing on the glass says so — which is also why a harness that waits
+for the *screen* to settle reads the floppy before the file exists.
+
 ## 82. CHART — charting, and the buffer both halves draw into (`apps/chart/chart.asm`, `apps/os88chart.inc`)
 
 Two consumers, one rasterizer. **CHART.O88** is a standalone viewer that reads
--- a/SPEC.md
+++ b/SPEC.md
@@ -96691,6 +96691,29 @@ What it costs is time at Save for a sheet whose formulas were not computed at
 paint, and nothing on the glass says so — which is also why a harness that waits
 for the *screen* to settle reads the floppy before the file exists.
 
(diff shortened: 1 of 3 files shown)


jggonz:
Thanks for this. I checked both fixes on the machine and they do what the PR says. I pushed one small commit, comments only, and the PR has no conflicts with `main`.

## What I checked

I built the branch and the fast tier passed (34 rows). Then I booted it in QEMU with a floppy holding two files I made on the host, opened each file in Sheet from the Disk window, saved it, and read the saved file back with a separate host-side FAT reader.

**A Save recalculates first (§81.22.1).** The SYLK file had `A1=1` and ten formulas `B = A1*1 … A1*10`, placed 50 rows apart so only the first one was on screen. I changed A1 to 3 and chose File ▸ Save. All ten came back current, including the nine that were never on screen:

```
C;X2;Y1;ERC[-1]*1;K3
C;X2;Y51;ER[-50]C[-1]*2;K6
…
C;X2;Y451;ER[-450]C[-1]*10;K30
```

**Error codes in the file's own numbering (§81.22.2).** The BIFF file had a `BOOLERR` holding 07H and a `FORMULA` whose cached result was error 07H.
- **Reading:** both opened as `#DIV/0!`. Before this fix they showed as `#N/A`.
- **Writing:** I typed `=1/0` into a cell and saved. The file now has `BOOLERR … 07 01` for both constants and `FORMULA … 02 00 07 00 00 00 FF FF` for the formula. Before this fix those bytes were 02H.

The saved floppy also passed `os88disk.py --verify`.

## Reading the code

- `sh_recalc_all` keeps the register contract `sh_eval_cell` needs. It saves CX and DX, since the parser underneath uses them, and it reloads `ES` from `sh_cellseg` on every record, which is right because the cell store can move.
- Taking the sheet from the top two bits of the row word matches `sh_unpackrow`.
- DIF's `sh_getcell2` doesn't evaluate twice: `sh_pass` doesn't advance during a save, so the second read is a cache hit.
- In manual mode, only formulas stamped `0xFFFF` get evaluated, as the PR says.
- `sh_biff_e2b` also catches code 0, because `dec` wraps it to 255 and the unsigned `ja` sends that to `#VALUE!`.

## The commit I pushed: 8404d17b

The new `sh_biff_errtab` / `sh_biff_e2b` / `sh_biff_b2e` block had been inserted between `sh_biff_cells`' header comment and the `sh_biff_cells:` label. Anyone reading the source would take that header as a description of the error helpers. I moved it back above its label. Only comments changed: I assembled `SHEET.O88` before and after and the two builds are identical.

## Noted, not changed

- A Save now takes longer on a sheet with many formulas that haven't been computed since the last repaint. In automatic mode that is every formula off screen. Nothing on screen shows that the save is busy. §81.22.1 already says so, and correct values are worth the wait, so I left it alone.

PR #179, opened 2026-09-11
Elendilon -> Main (CLEAR SKIES performance, DOT DELIRIUM, GFX Lines Library per app, Office/Games/Network 360k disks)

Elendilon:
**46 commits, 323 files, +131,248 / −9,184** against `2ce1e37`, the last squash.
Cut fresh from `main` at `52d482b`, so #161, #162, #171 and #176 are already in
and this merges cleanly.

**The kernel gets SMALLER on both builds** Every figure below is read off a clean build of the named commit
in its own worktree — `docs/reports/PR-CYCLE-ACCOUNTING-2026-09-11.md` is the
measurement and its method.

| | kern_big `.text` | kern_big `KERN_SIZE` | kern_small `KERN_SIZE` | shipped floppies |
|---|---|---|---|---|
| this PR | **−1,544** | **−1,536** | **−4,096** | 9 → 12 disks |

## CLEAR SKIES — the frame, measured scene for scene

The largest single subject. `tests/skiesprof.py` brackets each stage at its
CALL SITE and subtracts the emulator's own cycle counter, so no arm is compared
with another to get a number:

| scene | base | now | |
|---|---|---|---|
| mean of eight | 189.77 ms, 5.27 fps | **166.26 ms, 6.01 fps** | **−12.4%** |
| `sparse` | 103.65 | **68.42** | **−34.0%** |
| `rollsweep` | 294.44 | **240.54** | **−18.3%** |
| `turnhold` | 294.17 | **256.74** | **−12.7%** |
| `slightbank` | 232.91 | **209.69** | **−10.0%** |

## Also in

- **DOT DELIRIUM** (§93) — a new game, and back on the small floppies (§24.5.5)
  once the kernel change its omission rested on withdrew that ground. **An
  omission's ground is a claim about the KERNEL**, so a kernel change that
  touches it is a change to the disk list.
- **ArtfulType** — a dock box and the system clipboard (§46.5.2, §46.6.1).
- **`mem_regrow` sheds a purgeable claim before refusing** (§50.6.2.1, §27.6.1)
  — reported from the field as *"kern_small says Too big opening README.TXT"*,
  where the arithmetic says it should work. TWO defects each hid the other: the
  16KB grow was refused with **31,744 bytes sitting in three purgeable caches**
  — memory the kernel holds on the explicit understanding that it can give it
  away — and then `np_load` ignored that CF, so a **memory** refusal was
  reported as `FERR_BIG`, a sentence about the **file**, which sent the field
  looking for a size limit that was not the cause.
- **Three 360KB category disks** (§24.6) — office, network, games. 354 clusters
  is the geometry that runs out first and this project keeps making
  applications, so the answer is a disk per SUBJECT. `apps360.img` went from
  **8 spare clusters to 41**
- **The 128KB machine is single-disk** (§24.5.6) — its system disk carries the
  whole apps payload, 12 files to 25.
- **The `gfx_line` family is out of the kernel** and into `apps/os88gfx.inc`
  (§5.12.7): eight routines and three walk slots are `stc`/`ret` cells, and the
  image rung uncrosses four steps of 512. The glyph convergence then took
  ~110 bytes out of **eleven packages and five drivers** by changing ONE
  routine's body inside `%ifndef OS88UI_NOBTN`, with no per-package work.
- **An hourglass** (§7.5) for a machine inside a gfx-lock hold — the one shape
  no window owns, put on and taken off inside the hold so nothing can leave it
  on the screen.

## Testing

`make test-fast` 37/37 and `make test-full` green, plus the soak rows each
subject can reach — `soak -k 'skies*' -k 'cs*'` is 42 rows.

**Ten soak failures were investigated and closed this cycle and NOT ONE was a
defect in shipped software**: eight were rows that never declared the artefacts
they read or carried expectations that had gone stale, and two were host-clock
waits that hand the guest up to 37% less work under contention. Each is written
up as an incident in `docs/WRITING-TESTS.md` §13, which is the list the rules
are made of.

The sharpest of them is worth a line, because it is the failure mode this suite
keeps hitting: `wdcombo` slept 1.4 s for a press whose own work is a
`SYSTEM/FONTS` disk scan, and under load printed **five reds in a row** — a
broken record, a broken bank, a missing drag edge and 3,135 wrong pixels — for
**one late event**. The tell was in its own output all along: the same word read
0 at the press and 1 four steps later. Rewritten on the guest's clock it is 4/4
and its wall time no longer moves with load at all.


jggonz:
Merged `main` into this branch (only #177 had landed since your fork point) and pushed **8 commits** on top of yours. The PR is mergeable again, both kernels build inside their guards, and every gate I ran is green — details at the bottom.

The review ran as ten subsystem reviewers, each followed by an adversary trying to refute its findings; 41 findings went in, 14 were refuted, and I read every one of the 27 that survived in the code before touching anything. All 27 are fixed below, grouped by commit. Nothing here changes a design decision of yours — each is a repair in the shape the surrounding code already uses.

### The merge — `2777a98a`
`main`'s #177 added Scribe subsection 94.8.6 (SCRIBE.OVL read at entry). Your branch had already inserted §93 DOT DELIRIUM and moved Picture decoders to §94 and Scribe to §95, so the incoming subsection became **§95.8.6**, with the one `scribe.asm` comment that cited it following. Recorded beside the other renumberings in `docs/UPSTREAM.md`. `scribe.asm`'s code merged cleanly.

### `make` was red on a Mac — `8867a362`
**What you'd have seen:** on macOS, a bare `make` builds every floppy and then exits 1 on the new `deps` fast-tier row, with nothing wrong with the OS. **Why:** `tools/setup-linux.sh --check` refuses Darwin with exit 2 before it parses `--check`, and `t_deps.py` asserted `(0, 1)`. **Change:** on a Mac the test asserts exactly that refusal (2, naming `setup-macos.sh`); every other host keeps `(0, 1)`. Not `(0, 1, 2)` — a Linux box exiting 2 for another reason must still fail.

### DOT DELIRIUM — `d35834ba`
**What you'd have seen:** on a 128KB machine where the board picture's claim is refused, the window opens saying *Window too small.* — and then Enter, a click, Game ▸ New or Esc **hangs or reboots the machine**. **Why:** `dd_fit_claim`'s refusal leaves `[dd_bdseg]` = 0 with the geometry already published, and `dd_bd_wipe` is a `rep stosw` through that segment: 1,736 words at 0000:0000 on a windowed CGA — the interrupt vectors and the BIOS data area. None of the four routes to `dd_board_render` tested `[dd_ok]`. **Change:** the writer and the walk return on a zero segment, and `dd_new_game`/`dd_attract_begin` return on `[dd_ok]` = 0.

Three smaller ones in the same commit:
- A refused **regrow** (pressing F on that machine) was permanent: `.redo` banks the box *before* the claim, so `dd_relayout_ck`'s compares matched forever with the old claim under the new geometry, and the worker re-filled the whole content 18.2 times a second. `.fail` now poisons the bank with your own `[dd_lvkind] = 0FFh` idiom, so the next paint re-asks and it recovers when memory does.
- The arena's refusal said *Window too small.*, which is advice that makes it refuse harder (a bigger window asks for more). It says *Not enough memory.* now.
- Windowed, nothing drained the BIOS key buffer while an arrow was held (`OSAPI_KEY_DOWN` is state, not a dequeue), so §9.8 dropped the newest arrival — the P or Esc just typed. `dd_kbdrain` is `cy_kbdrain`'s shape, at the top of `dd_step`.
- `dd_tile_put`'s corner repair was the one pen path not returning on a one-plane adapter: two blits of bit-identical pixels a tile on Hercules/CGA, gone.

SPEC.md 93.3.4.5 and 93.7.2.1 record all of it.

### Kernel — `ff8052db`
- **The hourglass could strand itself on the glass.** `cur_busy_on`'s DOWN arm (`[cur_level]` = −1) called `cursor_show` without §7.4.3.1's lock test. Under a *free* lock that −1 is somebody else's hide — `fpg_arm`'s own bar-row hide, or a saver session's — and the show settled that debt twice: `[cur_level]` ends at +1 for good, `cursor_hide` never erases again, the ISR keeps drawing. FTPD writing a received file during a saver session reached it with no race. The arm now requires the lock held (your measurement found it 0 of 3 lock-free, so nothing measured is lost), and a saver session refuses the call the way `wm_clip_set` does. SPEC.md 7.5.3.
- **`OSAPI_PKG_REHOME` trusted the BX the entry returned** to mean "no window", where `.abort` — freeing the same region — sweeps by `wm_create`'s W_SEG stamp for exactly the reason its comment gives. A loader that created a window and returned BX = 0 left a record naming freed memory for the next `wm_paint_all` to far-call through. The re-home now runs `.abort`'s sweep before checking the part's header. SPEC.md 20.12.10.6. (No shipped loader creates a window, so this is a guard, not a repro.)
- **Control Panel whole-pane paint wrote every unselected row's ground twice**: `cp_list` lays the pane white, then `cp_listrow` filled each unselected bar rect white again — 5–8 `gfx_fill` calls a paint for nothing on the glass, PERFORMANCE.md rule 2. `[cp_lfull]` around the row loop, the way your `[cp_tfull]` does it one pane along. SPEC.md 31.1.4.
- The kernel's own API table now says RETIRED over the four line cells (the SDK already did), and `sched.inc`'s kern_small banner describes the four-slot table it emits rather than six.

kern_big `.text` +15, `.cold` +21; kern_small `.text` +8; no rung crossed on either; kernsize blessed for both.

### Paint, Tank, Missile, ArtfulType — `982d61dc`
- **Paint:** open any colour BMP/PIX/GIF on a Hercules or CGA machine and every 1px pencil stroke comes out as a smear of noise about four times too wide (healed by the next repaint, back with the next stroke). `pt_lnblit` reads the canvas as one bit a pixel but was gated on `[pt_mono]` (the display), where `pt_blit`'s identical fast path gates on `[pt_1bpp]` (the document). A colour file opened on a mono screen is a packed canvas under `[pt_mono]` = 1. The gate used to guard `pt_lndraw`, which didn't care about the format. It now falls to `.slow` on a packed canvas. `tests/paintstroke.py` can't see this — its canvas is born on a mono screen.
- **Tank (APP_SMALL):** the 16KB rung leaves 16,384 − 16,000 = **384** bytes of pool, not the 0 the ladder's table says, so `or ax, ax` never fired and the bottom rung started with `[tk_tmpl]` = 1 over a pool the panel overflows on its first encode (one wasted whole-viewport blit, a frame or two of cleared panel). `TKT_POOLMIN` = 1280.
- **Missile:** `mc_line`'s `.walk` didn't bank BP, which `gfxe_wstep` keeps the error term in; `mc_tr_step` and `mc_dsc_run` in the same conversion did.
- **ArtfulType:** after a dock restore, a *refused* `at_fs_enter` left a blank white card — `at_paint`'s `.win` arm deliberately draws nothing, so "the splash stays" was true for `at_argload` and not here. `at_onwake` draws the splash itself on refusal; and `at_paint` takes `OSAPI_WM_WAKE`'s CF=1 as a restore nothing will finish.

### SDK — `2fc08d43`
- **Radio group:** the ring sat at the row's top while the label was centred — 2 px apart at pitch 16, 4 px at pitch 20, the two pitches §13.17 names. `os88ui_radboxy` centres the ring (and the dot) `(PITCH − 12) >> 1` down; `tests/radio.py` reads it there and now asserts the centring. SPEC.md 13.17.1. No shipping caller yet, so nothing on a disk moved.
- **`gfxe_bclear`** cleared in words off a stride the contract only makes a multiple of 8: an odd stride left the band's tail unpapered, and a stride of 1 assembled the shift count as −1 (an 8086 doesn't mask it: nothing papered, the previous frame blitted back). Odd strides take a byte-counted arm, decided at assembly time like `GFXE_ROWOFF`. WIRE's band is 128 wide, so every shipped package is byte-identical.
- `OS88UI_RAD` gets the `%error` for `OS88UI_NOBTN` that `OS88UI_CHK` already has; three "fills its own ground" sentences in the SDK now say the caller owns the ground; `apps/os88gfx.inc`'s seven `GLS_*` offsets join `t_mirror`'s list (the kernel's copy has had no reader since 5.12.7); `apps/cc/os88.h`'s count recipe reads 164 in all four places (it read 162 in three and 160 in the fourth).

### Build and docs — `9d8e675c`
- **`make skiesprobe` didn't assemble** — the counting arm outgrew the 0xC200 overlay by 87 bytes, so every A/B price in SKIES-FRAME-PLAN.md was unrepeatable, and no tier could see it. 0xC300; all three diag arms assemble at 53,056 bytes, shipped build unchanged. (One trap I hit: a stale `build/skiesprobe/cswidx.inc` keeps the old address, since it doesn't depend on the Makefile — delete it after changing `CSVOCABAT`.)
- **The three category disks reached no release.** `SHIPIMGS` went 9 → 12 and SHEET/CHART moved off `apps360` onto `office360`, but the release skill's allowlist still packed nine — so the next 360KB release would have lost its spreadsheet and chart viewer. The manifest carries all twelve now. (@jggonz: this edits your release skill; if the category disks were *not* meant as release content, the SHEET/CHART removal from `apps360` is the half to revisit instead — both can't stand.)
- `checkdocs.py` wrote `(59, 20)` twice in `SITES_HIST` and Python kept the second, so the entry this PR added never existed; one key names both moments. `os88soak.py`'s preflight, `os88test.py` and CLAUDE.md no longer name the deleted `wirefps` row; `docs/TESTING.md` reads 37 fast / 371 soak; `tests/wireflick.py`'s docstring describes the absolute floor it implements rather than the comparison 78.5.1 made impossible.

### Tests — `5e1f4c0b`
`tests/atclip.py`, `skiesink.py` and `skiesspan.py` carried `/home/user/os8088` in `sys.path`/`ROOT`, so the gate for this PR's own ArtfulType dock round trip died on `No module named os88fixture` on any checkout that isn't that container. They derive `ROOT` from their own location now, the idiom the rest of `tests/` uses; both rows pass here. (Thirty-seven older test files carry the same path and predate this PR — left alone, but worth a sweep some day.)

### Flagged, not fixed
- `tests/curbusy.py` double-clicks the B: drive *under* its `Watch` sampler, and on this 8 GB Mac the sampler's reads on MartyPC's single debug channel push the two presses to 10 ticks apart against a 9-tick window — three runs, three misses at that first double-click, before any hourglass code runs. Opening the drive before the watch starts (a two-line variant) passes the whole row: hourglass drawn 71 samples across a launch, arrow again once idle, `OSAPI_CUR_BUSY` reached under a held lock. Your box is faster, so you may never see it; moving `ui.open_drive("B")` above the `with Watch` would make the row robust.
- SPEC.md 5945 and 97974 still name `tests/wirefps.py` as half of 5.6.4.1's A/B. Either the sentence stops naming it or the row comes back — your call, since the deletion isn't mentioned in the commit that made it.
- `make small -j8` races: `small360.img` and `small.img` each recurse into a sub-make for `build/smallk/`, and `os88kz` then reports "the two passes packed DIFFERENT bytes". Serial `make small` is fine. This predates the PR (same recipe on `main`).
- Reviewers also noted, and I left alone: `mem_regrow`'s shed loop can purge the directory read-ahead window on a grow that refuses anyway, and kern_small's `MEM_P_VIEW` listing cache isn't released on a Disk window close. Both are design questions rather than repairs.

### What I checked and cleared
API slot table by address and name (four slots appended at the tail, nothing moved, the four retired cells still at their addresses answering CF=1); the three deleted tests (all yours, no dangling registry rows); no duplicate § headings; nothing in `build/`, no new binaries; all 14 shipped images fsck clean; `make` (fast tier 37/37 on a Mac now), `make small`, `kernsize` both variants, `test-full` 41/41 with the C toolchain, `t_mirror` 383 checks, `make skiesprobe`, and the soak rows the fixes reach, every one green here: dotdel (all three arms), dotdelpen/win/md, ddsmall, radio, curdisk, rehome/rehomemove/rehomeabort, regrowshed, paintstroke/paintsu/paintnogrow, tanksmall, atclip, mcseg, mcperf, wireflick, cplistrow, skiesspan, and the eleven `tm*` rows. Several MartyPC rows needed a second run alone on this host (the *two FIRST clicks* and *did not happen in 30 GUEST seconds* shapes, always at a disk-window open before the row's subject was reached) — the same flake docs/WRITING-TESTS.md 13 already describes, not a defect in the code.

jggonz:
One more: `main` gained #178 (the Sheet save fix) while this review ran, so I merged it in as well — `91f772a6`. Its §81.22 and your §81.30 were appended at the same place and are different numbers, so both stand, 81.22 first; `sheet.asm` and `docs/BIFF-NOTES.md` merged cleanly, both kernels are unchanged in size and the fast tier is green. 52 commits on the branch now, and GitHub reports it mergeable.

PR #197, opened 2026-09-18
Elendilon -> Main (DOS: .COM and .EXE In-OS or Whole-Machine, Shared Button Control, Machine-Wide Icon Store, No Global Directory Listing, Region Self-Compaction, Both Kernels Smaller than at the last Squash)

Elendilon:
780+ commits since #179's squash, across 554 files. DOS, and the rest of the cycle is either the kernel surface that section needed or work that stands on its own.

**The headline is §96: this machine runs real DOS programs now.** The other
half of the headline is that **both shipped kernels come out of the cycle
smaller than they went in** — `kern_big` by one 512-byte rung and `kern_small`
by three — after adding DOS, an icon store and a document-glyph column, and
after taking `main`'s Dock and USB mouse on top.

## DOS (§96 — `apps/dos/`, `kerndos/`)

**Not an emulator.** This is an 8086 in real mode and a DOS `.COM` is machine
code for the processor already running, so nothing is interpreted: it
builds the memory a DOS program expects, points `INT 21h` at code of its own,
and far-jumps in. What it provides is the *operating system* the program
calls, not the *processor* it runs on.

It ships as two arms, and the Memory page makes the trade the user's to make
(§96.25):

**In-OS** — `SYSTEM/DOS.O88`, a `SYSAPPS` package, on all four system-disk
geometries. A console band, a setup area and a path box in one window;
`.COM` and `.EXE` by signature rather than extension; a real MCB allocator;
file handles, the DTA, find-first/find-next, the five standard handles, `CON`
as a device; `INT 33h` as a translation rather than a driver; XMS over four
slots the kernel already published; `AH=4Bh` running a child and coming back
to its parent; a shell with built-in commands rather than a `COMMAND.COM`;
`DIR` with its switches, sorted the way a real DOS 3.30 sorts; arguments, an
environment, and a shortcut that carries how to run it. About **447 KB** of
arena — it was 446 until §22.6.3 below gave the heap back the 1,536 bytes the
global listing was holding.

**Whole-machine (`kern_dos`)** — the machine goes away, runs DOS, and comes
back. About **585 KB** to the program, against IBM DOS 3.30's 590. The return
is a live resume, so the session comes back with no boot — §87.5's hibernate
machinery with a different payload, staged through the text framebuffer, the
one RAM a conventional-memory image does not cover.

**It loads faster than IBM DOS 3.30.** Test Drive III's load, one machine and
one disk, both arms watching `INT 13h` from the launch to the same
four-second silence:

| | `kern_dos` | IBM DOS 3.30 |
|---|---:|---:|
| `INT 13h` calls | **37** | 57 |
| sectors | 425 | 348 |
| sectors a call | **11.5** | 6.1 |
| wall | **16.1 s** | 20.7 s |

That is §18.95's read-ahead surviving the handover — `kern_dos` keeps one rung
of it, 9 KB of the program's own memory, and buys 4.7x over keeping none.

Programs that run: Prince of Persia, Sopwith, Test Drive III, Commander Keen,
Tom & Jerry, Microsoft Works — `INSTALL.EXE` to completion, and it saves a
document — and mTCP's own `PING` over a packet driver of ours.
**`docs/reports/DOS-GAMES-2026-09-16.md`** is fifteen programs measured with
every refusal explained.

Design record: `docs/plans/DOS-EXEC-PLAN.md` and `docs/plans/KERN-DOS-PLAN.md`.
**`docs/DOS-DEBUGGING.md` is the toolkit**, and its premise is that the box is
not debuggable from one side — our own trace says what we were asked and what
we answered and both look right, through four separate defects in one session.
`tools/os88dosdbg.py` runs the same program under our box and under a real IBM
DOS and aligns the two traces on (function, call site), which turns *"they
diverge somewhere"* into *"they diverge at this instruction"*.

## What DOS asked the kernel for

Every one of these is a published door useful on its own, not a private hook:

- **`OSAPI_FILE_COPY` / `OSAPI_FILE_MOVE`** (§22.24–22.25) — the copy engine is
  a door rather than a menu item; one door with a verb byte, −215 bytes against
  two.
- **`OSAPI_FILE_WRITE_AT` / `_APPEND`** (§18.4.7) — one body behind two doors;
  `kern_small` keeps only the append door.
- **`OSAPI_VOL_STAT`** (§18.4.6) — the volume's size in `AH=36h`'s own four
  registers, over one slot that answers a family.
- **`OSAPI_FILE_PATH`** (§19.2.4) — a package can name where it stands; three
  walkers the file system already had, 393 → 137 resident bytes.
- **`OSAPI_PKG_START`** (§21.5) — one door into the loader, and a document may
  ride with it.
- **`OSAPI_DRV_SUSPEND`** (§51.11) — the hibernate's own sweep, moved into
  `HIBER.DRV`; the DOS handoff is its verb 2. It is what gets the sound
  driver's 14 KB back for the program (§96.35).
- **Raw frames on `ETHER.DRV`** (§72.22) — whole frames for a consumer that IS
  a stack, which is what the packet driver needs.

And §18.95's read-ahead window grew up: it claims no 64 KB page head (that was
placing it), it MOVES, it asks what the machine HAS, and a program about to
take the arena may COMMAND its width. **530 resident bytes** after a size pass.

## There is no global directory listing any more (§22.6.3)

**This is the change that put both kernels under the squash, and its code
column points the wrong way.** `disk_dir` was the current directory as a
KERNEL-WIDE snapshot — `disk_mount` built it and every window's claim was a
COPY of it. `docs/plans/LISTING-HOME-PLAN.md` started from a size question,
*what is `DSK_NENT` buying us?*, and found that the listing is not the file
manager's cache at all: **no package anywhere reads it** (`OSAPI_FILE_FIND`
re-walks the directory through the sector cache, §19.7.1), and three of the
census's frightening-looking consumers — `ui.inc`, `assoc.inc`, `snd.inc` —
name the `dsk_get_dir` idiom while describing staging loops of their own.

So a listing is written into the store its CALLER supplied, and **a mount with
no destination is QUIET** — the scan, the sort and the icon harvest are all
skipped. The file dialog claims its own store, a Disk window supplies its own,
a launch reads the poster's own cache, and hibernate finds its image by name.

`.lowbss` falls **7,966 → 6,366** on `kern_big` — 1,600 bytes exactly, being
`DSK_NENT` 64 × `DSK_DE_STRIDE` 24 plus 64 bytes of `dsk_icoix` — and
`LOW_PARA` **8,704 → 7,168** with it. `.lowbss` sits below `HEAP_SEG`, so that
is **three 512-byte rungs off the image and 1,536 bytes onto every machine's
heap and every DOS program's arena**. The lane's own `code` figure is
**positive** (+335: `.text` +162, `.cold` +173, `.bss` +7) — nothing got
smaller, a buffer stopped existing.

**The hard disk's 6 KB-per-partition listing claim went the same way, for a
sharper reason** (§22.6). `HDD.DRV` donated it so a driver-backed volume could
list 64 entries where a floppy got 32, and **two unrelated changes had already
withdrawn that reason without anyone looking**: the icon store moved four
fifths of the claim's contents out, and §22.6.2 then raised `DSK_NENT` to 64
for DOS, so the floor became the cap the claim funded. It held 24 KB on
a four-partition machine with at most one of the four ever in use.
`tests/hdmove.py` — 394 lines proving the claim could move — goes with the
claim it measured, and `tests/hdnoclaim.py` asserts its ABSENCE instead. The
Control Panel page moved to `HDDTOOL.DRV` and took `HDD.DRV` **8,152 → 5,121
bytes**, where one byte was worth a kilobyte because `drv_load` claims whole KB.

## The button is a control, not a painter (§20.5.1.2–20.5.1.4)

`os88ui.inc` composed every shared control except the one every dialog uses.
The button was a painter each caller drove, and three defect shapes kept
recurring because of it. It is a control now:

- **`OS88UI_BTNREC`** — the record is DECLARED, not assembled at each site.
- **`OSAPI_WM_ONCLICK` (0x05A0)** — the PRESS belongs to the library, so no
  click path can skip the buttons.
- **The erase moved INTO the control** (§13.14.6), where it stops being
  copied wrong.
- **`os88ui_btnraw` is deleted** — 56 call sites, one control, no scaffold.

Dialogs across Word, Scribe, Artful, Audio and Note Pad now fire on the
RELEASE, which is what a button is. `tests/btnall.py` drives every button in
the tree and `tests/btnsites.txt` is the ratchet that keeps a new hand-rolled
one out.

Beside it, the **attribution sweep** (§20.5.1.2): seven packages whose About
card credited nobody.

## One icon body per machine (§25.9)

The listing kept an icon body per entry, per window. It is a machine-wide
store now, keyed by what the directory already says, purgeable at the cheapest
rank there is, and `ASSOC.DAT` absorbs into it rather than holding a claim of
its own. **The Disk window lists SIXTY-FOUR entries and is lighter than it was
at 32** (§22.6.2), and `.lowbss` comes down **1,216 bytes** — the first of the
two cuts to that section this cycle, §22.6.3 above being the second.

A package may also ship its own 8×8 document glyph (§54.3.2).

## A region that compacts itself (§66.4.3)

Two defects that each made the other measure as worthless, which is how both
stayed invisible. `mem_frameless` asks `mem_in_nest`, and a package reaches
`mem_claim` only from inside its own callback — so the compactor pinned the
asker's own region **by the act of asking**. And `.doit` ran ONE pass and
returned, so **a claim needing both passes got NEITHER**.

Measured: region pinned 494.5K, un-pinned with one pass **494.5K**, both passes
502.5K. **Each fix alone buys zero**; the pair is **35 bytes** and buys the
ceiling hole to the byte. `mem_avail` then had to learn to plan both passes,
and the what-if is a measurement rather than a promise (§66.4.3.2) — a failed
claim is destructive, so a refusal has to mean *"I could not have had this even
trying"*. Tracker is the first consumer and the exact-requirement one.

Alongside: a claimant may name a **purge floor** (§50.6.6) — *"compact the disk
cache, do not destroy it"* — and `mem_regrow` sheds a purgeable claim before
refusing (§50.6.2.1).

## Smaller things worth naming

- **`gfx_points` runs VIRTUAL** (§5.6.9): no bounding box, and a second display
  stops being a reason to abandon the inline loop. Extending the desktop was
  **4.59x** slower than a single display and is **1.02x** now, measured with
  the window placed so the second card never saw a pixel of it. `kern_big`
  `.text` +166 against the 370 the box would have cost; `kern_small` comes out
  **5 bytes smaller than before the feature existed**.
- **Scroll bars follow the thumb** (§13.10.5.4.1–2): the rate is a PAIR and the
  machine picks on `[cpu_tier]`, plus a SECOND trigger — commit when the hand
  STOPS, which is the only one an 8088 gets. Settled at 12 MHz on a 286.
- **The busy pointer is a CLOCK** (§7.5), and it belongs to a lock hold rather
  than to a window.
- **Alt+Enter** into and out of full screen, on every app that has one (§9.7.1,
  §11.2.1.1) — the combination no XT BIOS delivers at all.
- **A wheel mouse's fourth byte** no longer breaks the packet run (§9.5.4);
  `MOU_IDMAX` 8 → 128 and `MOU_DRAINT` 9 → 18 (docs/FIELD-NOTES.md 44).
- **PACMAN is retired** (§89.12), and `apps/RETIRED.txt` makes *"does not
  ship"* a thing the tree can check rather than a thing to remember (§20.16).
- **The 128 KB machine is single-disk**: the small system disk carries the
  whole apps payload (§24.5.6), and DOT DELIRIUM goes back on the small
  floppies, its ground having been withdrawn by §5.4.2.5.1 (§24.5.5).
- **ArtfulType** gets a minimize box and the system clipboard (§46.5.2,
  §46.6.1).
- **The console is `apps/os88con.inc`** now, and Telnet is one carrier
  (§70.8.12) — the full-screen renderer went with it.
- **Clear Skies** gains a per-aeroplane engine sound at 0.1–0.2% of a frame
  (§88.8.2), and scenery that had been wrong since it landed.
- **CYCLONE's claw sweeps an ARC** rather than sampling a point, the pickup
  window is an AREA rather than an instant, and a death no longer restarts the
  wave.

## Fixes to code `main` already has

These are not this cycle's own churn — each was checked against #179's tree
and the defect is in code that predates it:

- **Dual-screen Hercules/CGA was REVERSED** (§39.18.1.1, §39.19.4.1): the CGA's
  mode byte is ours and not the ROM's, and `vid_text` never got §39.19.4's
  `vid_cga_equip`.
- **An animated screen-saver session didn't darken the second monitor** (§79.1.1).
- **Chrome that is WHOLLY obstructed was drawn anyway** (§11.97.3–4) — an empty
  clip region meant draw everything rather than draw nothing. The 1bpp CGA
  title strip reads **1,434 → 0**.
- **`menu_hover` read the pointer twice**, so the second read could disagree
  with the first (§12.4.1).
- **The square wave restarted on every tone change**, so a glide wobbled
  (§34.1.1).
- **A floppy swapped under a running PACKAGE was invisible** (§18.9.1.1) — the
  package's quiet stand had no staleness predicate at all.
- **Every os8088 disk of a geometry signed the SAME**, and the reason that was
  safe was never true (§18.8.2).
- **A resize that changed NOTHING vacated a region anyway** (§11.91.5).
- **`font_char` dropped a cell the clip region cut VERTICALLY** (§11.3.4) — the
  horizontal cut had been fixed and the vertical one never had.
- **A driver's own heap claims were on no row of the heap page at all**
  (§28.4.6).
- **Twenty-one toast messages did not fit the bar**, and the claim that they
  did had never been checked (§59.10) — it is a gate now, and `TOAST_MAX` is
  asked of the machine.
- **Clear Skies**: the scene read the heading as the facing (§88.7.8.2), the
  bank and the panel read the Euler triple (§88.7.8.3, §88.9.2.6), San
  Francisco could not fly at all — the last stream can never fill a
  cluster-rounded read (§88.10.5.4.1) — the Golden Gate's roads had nothing
  under them (§88.6.4.1).
- **DOT DELIRIUM**: a refused claim could wipe `0000:0000` (§93.3.4.5).

## What it costs

`docs/reports/KERNEL-BYTES-SINCE-SQUASH-2026-09-17.md` is the accounting — per
point, with `kernsize` RE-ASSEMBLING each tree rather than reading a build, and
re-measured at this tip as point **G**. The previous point reproduces to the
byte on the box that took this one, which is what makes the columns comparable.

| | `kern_big` | `kern_small` |
|---|---:|---:|
| `KERN_SIZE` vs #179 | **−512** (one rung) | **−1,536** (three rungs) |
| section total | −421 (−537 resident, +116 overlay) | −1,615 (−1,636 resident, +21 overlay) |
| spare of budget | 19,456 (38 steps) | 33,280 (65 steps) |
| `.text`+`.bss` of `KERN_CODE_MAX` | 56,257 — **9,279 left** | 41,519 — 24,017 left |

**Both shipped kernels are smaller than they were at the squash**, having taken
`main`'s Dock and USB mouse on the way. That is not a general shrink and should
not be read as one: `.text`, `.bss` and `.cold` all went UP. It is `.lowbss`,
which sits below `HEAP_SEG` and so moves the whole image — the icon store took
1,216 bytes of it and §22.6.3 another 1,600.

**`KERN_CODE_MAX` is the line to watch rather than the rungs**, and this cycle
is the sharpest illustration of why the two differ: the window that took three
rungs off `KERN_SIZE` moved `KERN_CODE_MAX` **169 bytes the wrong way** in the
same commits, because `.lowbss` is not code and `gfx_points` is. It cannot be
raised at all — offsets are 16 bits — and **9,279** is what is left.

The report also carries the audit of **which concepts have had a size pass**,
over two squash cycles. The previous cycle has nothing outstanding; this one
has **one** — the mouse wire work, which has now stood through three audits.

## Testing

`tests/suite.py` registers **527 rows** against `main`'s 420 — 46 fast, 5 full,
476 soak, against 40 / 5 / 375. `make test-fast` is green on this tree
(**46/46**), and the cycle added rows for each of the families it touched — DOS
(`dos*`, `kd*`), the heap and the compactor (`heap*`, `reg*`, `trk*`), the
buttons (`btnall`, `btngesture`), the displays (`disp*`), the small-disk
requirements (`smallreq`, `ddsmall`) and the retired HDD claim (`hdnoclaim`,
which replaces `hdmove`).

**`docs/WRITING-TESTS.md` is new** and is the other half of that: how to
register a row and the four failures that keep coming back — a `secs` nobody
measured, a `builds=True` where a private tree was the answer, a hand-rolled
click at a remembered coordinate, and a `time.sleep` that hands the machine 37%
less work under load. Its §1 is the only question that decides whether a row is
worth having: break the thing on purpose and watch it go red.

## Merge notes

- **One new top-level section, §96, appended** — nothing renumbers on our side.
  §89 is marked **RETIRED** rather than removed, so its number stays taken.
- **The API table gained nine cells at the tail** — 174 against `main`'s 165.
  #172's cycle had both sides append there and merge *cleanly* into two names
  at one address; `docs/UPSTREAM.md` carries the collision check that found it,
  and it is worth running again.
- **`apps/dos/` is the only new directory under `apps/`.**
- **`CLAUDE.md` here is `main`'s file wholesale**, plus four knob rows for knobs
  this cycle added (`DPTROM`, `NOKDKBD`, `DOSRMARK`, `DOSNETCARD`). The fork's
  own branch rules and its extra MartyPC material are deliberately not in it.
- Nothing under `build/` is tracked.


jggonz:
Reviewed this one end to end — built both kernels, ran the full gate, and drove the DOS box on the glass. I've pushed **one commit** (`4fc7544`) with five fixes; the PR is **mergeable** now.

First, the headline: this works. `DOS.O88` launches, the shell runs, `DIR` reports the right free space through the new `OSAPI_VOL_STAT`, drive switching works, and a real `.COM` (`OS88NET`) loaded, went full screen, scanned the LPT ports and listened. `make test-fast` is 46/46 and `os88test full` is 50 passed / 0 failed, matching what you claimed. The lost-from-main sweep came back empty — Dock/`DOCK.DRV`, the USB mouse, the RunCPM zip, the imager fix, the Sheet and Scribe fixes are all intact.

---

## What I fixed

**1. A folder open could write into someone else's memory.** `4fc7544`, `kernel/memory.inc`

On a machine with a busy heap, opening a folder in a Disk window could write about 1.6 KB of directory entries and icon references into whatever claim now sits at the old address — a running package's region, another window's cache. It wouldn't crash; it would just be wrong, somewhere else, later.

The `MEM_RR_ONE dsk_dseg` row came out with the note *"a listing is never in a claim."* §22.6.3 inverts that — a listing is now **always** in the caller's store, and `fmv_load` aims `dsk_dest_x` at the window's own `FS_VSEG`, which `fm_kinit` declares MOVABLE. The mount then reaches `mem_claim` through `ico_need`'s lazy icon-store claim, so a compaction inside the harvest moves the cache while `fm_reloc` patches `FS_VSEG` and `[fm_vseg]` and not this word — and `dsk_sortdir` and every `ico_ref_set` below it write through `mov es, [dsk_dseg]`.

§25.9 says the ordering is *"the whole of the care"* and cites `tools/dsegaudit.py`. That tool models a segment held in a **register**; `[dsk_dseg]` is a variable the mount reloads, so it can't see this. I restored the row (4 bytes, survives any ordering) and corrected that paragraph beside it. Your ordering discipline is still the design — the row is just the backstop.

**2. The memory what-if could move a running program out from under itself.** `4fc7544`, `kernel/memory.inc`

A package asking `OSAPI_MEM_COMPACT`'s what-if could, if a tick landed in the wrong 10 instructions, have **another** task's compaction move its region while its own code was running there.

`[mem_cp_self]` is machine-wide, and `mem_can_move` reads it from `mem_cp_walk` — the real move, not just the plan. It was stored and cleared at IF=1 around a call that re-enables interrupts. Now bracketed in `pushf`/`cli` … `popf`. Worth noting you fixed exactly this class for the purge floor by stamping `[mem_pg_fl]` with `sch_cur`; this word just didn't get the same treatment.

**3. "Press any key to restart" on a machine that can't read a key.** `4fc7544`, `kernel/hbstub.inc`

If a resume failed, the user got a message naming the failure and then a dead machine — no key would do anything.

`.fail` sits past `.done`'s mask restore, so it printed that prompt with IRQ1 still masked by whichever host staged the stub (`hbm_handover`'s `or al, 0x1A`, or `kd_resume`'s). The ROM's `int 09h` never runs, nothing reaches the BIOS buffer, and the `int 16h` waits forever. Masks now go back first, so §96.49's promise that *every* refusal falls back to `int 19h` is true again. **This one reaches ordinary hibernate too**, not just the DOS resume — both stage this stub.

**4. The live DOS resume couldn't complete its own disk read on an AT.** `4fc7544`, `kerndos/kdresume.inc`

`kd_resume` masked the whole master PIC (`mov al, 0xFF`) where `hbm_handover` masks `or al, 0x1A`. The stub `sti`s and calls `int 13h` with your own comment right there — *"the ROM wants its interrupts on an AT."* Masking everything also takes IRQ0's tick, IRQ6, and IRQ2, the cascade an AT-class fixed disk answers on. So the read couldn't finish and the resume fell into the `.fail` arm above. Every `kdreturn*` row stays green because MartyPC's XT HDC and SeaBIOS both poll.

**5. The resume tracer corrupted the clock it was watching.** `4fc7544`, `kernel/hbmark.inc` (`DOSRMARK=1` builds only)

`HS_TICK0` is a **dd** at `0x0446`, so it owns `0x0446..0x0449`. `HM_CUR` was at `0x0448` — a two-byte overlap. `HBMARK` wrote 2 into the staged tick on every resume: an ordinary resume lost `hb_wake`'s `0xFFFFFFFF` sentinel, and a live DOS resume woke the clock about 22 hours out. Moved to `0x044A`, with a `%if` beside the declaration so it fails the build if they ever overlap again.

**What it cost:** `.text` +4, `.cold` +3. That crossed one 512-byte image rung — `KERN_SIZE` 110,080 → 110,592, spare 19,456 → 18,944 (37 steps) — because the tree was sitting at 464/512. `kern_small` is byte-identical. I deliberately did **not** re-bless `docs/KERNEL-MEMORY.md`: it still reads against `main`, which is what makes this cycle's deltas comparable. Re-bless at the squash if you'd rather the figures match.

---

## Flagged, not fixed

**`make NOKZIP=1` no longer builds.** It dies packing the 360KB system disk:

```
os88disk: error: packages need 356 clusters; disk holds 354
```

I A/B'd it against `main`, which builds it comfortably at 296/354 with `KERNEL.SYS` at 206 sectors. On this branch the kernel is 209 sectors and the 360KB system disk now also carries `DOS.O88`, which tips it over by 2 clusters. `SPLSTARS=1 NOKZIP=1` fails with it. This matters a little more than a knob usually would, because `NOKZIP=1` is the only thing that ever exercises the *unpacked* arm of either loader (`tests/kzboot.py --nokzip`). I left it alone because the fix is a decision about what rides that floppy, and that's yours. Every other knob I tried assembles: `DPTROM`, `BOOTSTOP=1/2`, `BOOTMARK`, `BOOTHALT`, `BOOTPROF`, `TRACKRUN`, `DLJUNK=0x61`, `NOKDKBD`, `DOSRMARK`, `bootdiag` and 13 drawing knobs — 24 configs, `boot.bin` exactly 512 bytes with 35–70 bytes spare each. `SPLSTARS=1` alone refuses with its documented sentence, correctly.

**The API table.** Three addresses `main` had published changed meaning: 0x0300 (`GFX_LINIT`→`DSK_CACHE`), 0x0520 (`PKG_RUN`→`PKG_START`, with the registers swapped), 0x0550 (`MOUSE_FEED`→`DRV_SUSPEND`). I checked every one and **the in-tree sweep is complete** — no live call site on a stale contract, the jump table matches, and no other slot's registers drifted while keeping its name. §20.8 rule 4 permits all three and your `1fca8b0b` argues the `PKG_START` merge well, so we're accepting them as-is. One thing worth writing down somewhere: The Wire serves prebuilt `.O88`s from os8088.com with no ABI stamp, and the package header carries only a *format* version. §20.8's licence rests on this tree hosting every package, and that's the one place it leaks. Not this PR's problem to solve — it's true of any table change, appends included — but three re-contracts make the failure mode quieter than a wild jump would be.

**The PR description says `SYSTEM/DOS.O88`, a `SYSAPPS` package.** The Makefile puts it in the system disk's `APPS/` and explains why at length (`assoc_locate` has no rung that looks in `SYSTEM/`). The code is right; the description is stale.

**`CD` with a multi-component path.** `cd system\dos` gives *Path not found* where `cd system` then `cd dos` works. I chased this before finding §96 documents it as *"a limit rather than a bug"* — so, not a defect. Flagging only because it constrains `AH=3Bh` for a DOS program too, not just the shell.

**Findings I did not verify deeply enough to act on.** These came from focused review passes and are worth your eye; I'm reporting them as leads, not conclusions, because I confirmed the five above by reading and stopped there:

- `fmv_load` aims the destination *before* `fmv_fit`, so a window whose cache was refused or shed may list nothing and have its debt already cleared — the blank-window case `fmv_demote`'s own comment says it exists to prevent.
- `fmv_copy_out` has no zero-segment guard where `fmv_copy_in` carries one and explains why (a copy from segment 0 reads the IVT).
- `dsk_dest_x` is a single global with no save/restore, and the Standard File dialog aims it for its whole lifetime.
- `ICO_NROW` 48 against `DSK_NENT` 64 — past 48 distinct keys the harvest still pays the sector read and then draws the generic icon.
- `[mem_pg_rank]` is handed to `mem_cpq_run_x` through a global across a window where `[sch_lock]` is 0.
- `mem_cp_both` plans with `mem_pg_cheap` where the real walk uses `mem_cp_drop` (which also refuses a dirty FAT window), so the what-if can over-report.
- `mem_free_rec_x` lifts the purge floor without checking `[mem_pg_ftask]`, where the SDK says it's lifted "for nobody else."
- `.both` runs two full compactions with no plan check first.
- `kd_resume` takes the directory entry as truth — no magic, stamp, or upper bound on `[kdr_secs]`, where `hbm_ask` checks all three on the reboot route.
- `KDL_VTAB + KDL_VMAX*KDL_VROW` lands exactly on `KDL_WAKE` with no `%error` holding it there.
- `dos_restore_machine`'s IVT copy uses `cli`/`sti` where the house rule is `pushf`/`cli`/`popf`, and it runs on a path `kd_leave` documents as entered with IF=0.

---

## What I checked and cleared

No duplicate `§` headings, nothing tracked under `build/`, `.gitignore` and `.claude/` untouched, `CLAUDE.md` is `main`'s file plus your four knob rows, and the only two deletions (`tests/hdmove.py`, the handoff doc) are both explained. All 14 shipped images pass `os88disk --verify`. Booted to a desktop on VGA and re-drove DOS after the fixes — identical behaviour.

Nice piece of work. The `.fail` and PIC-mask ones are the kind that only bite on hardware nobody here has, which is exactly why they were worth finding before this went out.

